PR #41, opened 2022-08-18
Feature/108343 asset allocation page

dotcom9:
Add hidden property to GovUkDateInput


sussexrick (changes requested):
This looks like this will render an HTML `hidden` attribute - is that correct? If so how does that help with data binding?

`hidden` is a boolean attribute in HTML, ie `hidden="false"` means an element is hidden, just like `hidden="honeymonster" does`, so this looks like hidden is always applied.

When updating the Umbraco config for the GOV.UK components the workflow is as follows:
- make the update in the example app
- uSync > Everything > Clean export
- copy any changed `GOVUK*` files from the example app to the `GovUk.Frontend.Umbraco` project. Usually this means files in `uSync\v9\ContentTypes` and `uSync\v9\DataTypes`
- ensure that the copied files have the same build properties as their new siblings, which they won't by default (editing the project file is the quickest way to fix this)

sussexrick:
@dotcom9 The code appears to be working without this change now?

PR #46, opened 2022-08-22
Moved exportable content to Pages and updated refs

bentpr:


sussexrick:
@bentpr In the Umbraco example app the partial views are no longer found:

image

sussexrick (approved):

PR #51, opened 2022-08-23
Feature/108343 asset allocation page

dotcom9:
Add extension methods to find a blocklistitem in a published request


sussexrick (dismissed):

sussexrick:
To be clear, and consistent with the existing implementation, this method should be called `FindBlockByBoundProperty`.

sussexrick (changes requested):
To be clear, and consistent with the existing implementation, this method should be called FindBlockByBoundProperty

sussexrick (approved):

PR #54, opened 2022-08-24
AB#108354 make legend optional

HarryAmmon:
Bonds Category breakdown page requires a fieldset to display the error for the total of the text inputs. However the prototype doesn't include a legend for the fieldset.


sussexrick (changes requested):
Yes it does have a legend. You need to go into the settings for the fieldset and turn on the "use legend as page heading" option. That might still create some "interesting" challenges for getting the HTML we need because I don't think there's currently scope to insert an error message, but it won't be because there isn't a legend.

**UPDATE** -- This [new PR](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/pull/55) has a new example page that shows you how to set it up

PR #59, opened 2022-09-12
Detect if field is numeric and add type=number

HarryAmmon:
AB#137989 

Currently if the `Range` attribute is not used for a numeric field (which is the approach taken for the SR Umbraco app for percentage fields) a user can enter letters in a number field. This change detects if the field in the view model is numeric and if so add type="number" to the input tag in the HTML which prevents the user entering letters.


sussexrick (changes requested):
As this check now applies to every field, it's safe to remove the old copy from the scope of the range validator.

Also there is a unit test which checks for `type='number'` on range fields, so it should be simple to apply the same test to another property to prove that it's no longer limited to range.

GovUk.Frontend.AspNetCore.Extensions.Tests/ClientSideValidationHtmlEnhancerTests.cs
```diff
@@ -521,6 +523,36 @@ public void Numeric_range_validator_adds_range_attributes()
             Assert.True(document.DocumentNode.SelectSingleNode($"//input[@type='number']") != null);
         }
 
+        [Test]
+        public void Numeric_fields_without_range_have_type_equals_number_added()
+        {
+            var viewContext = new ViewContext() { ClientValidationEnabled = true };
+            var propertyResolver = new Mock<IModelPropertyResolver>();
+            var property = typeof(ExampleClass).GetProperty(nameof(ExampleClass.NumberField));
+            //var min = (property.GetCustomAttributes(typeof(RangeAttribute), false)[0] as RangeAttribute).Minimum;
```
sussexrick: Commented lines can come out. Otherwise OK.

PR #60, opened 2022-09-12
Detect if field is numeric and add type="number"

HarryAmmon:


HarryAmmon:
Currently if the Range attribute is not used for a numeric field (which is the approach taken for the SR Umbraco app for percentage fields) a user can enter letters in a number field. This change detects if the field in the view model is numeric and if so add type="number" to the input tag in the HTML which prevents the user entering letters.

sussexrick (dismissed):

sussexrick (approved):

PR #66, opened 2022-10-03
Add duplicate of file TagHelperContentExtensions.cs

samuelhaa:
Because
- The class is declared internal in GovUk.FrontEnd.AspNetCore and we want to access it in our own assemblies

This commit adds the class to our project so that we can access it

This file should be deleted once the class is declared public in GovUk.FrontEnd.AspNetCore


HarryAmmon:
Have you considered a PR in that repo to make it public?

sussexrick (approved):

sussexrick:
@HarryAmmon We're asking if it can be made public, but even if it is it won't be available until the next NuGet release of that package, so hopefully this is a temporary copy.

PR #67, opened 2022-10-04
Make all numeric input fields of type text

HarryAmmon:
AB#133749 
Changes:
- Add the "text" attribute to input elements when the model is a numeric type
Because:
- We want to add commas to monetry fields which are not valid an input element of type "number". Also, the [GDS Standards](https://design-system.service.gov.uk/components/text-input/#numbers) say that type "text" should be used for entering numbers unless specific user research suggests we should do otherwise. There is a longer read on this on the [GDS blog site](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/).


sussexrick:
@HarryAmmon I'd expect the tests you've already found to be looking for `inputmode=numeric` on the range field, and for `pattern=[0-9]*` in both cases. Looks good otherwise.

HarryAmmon:
@sussexrick Added tests and custom range validator which removes any commas and copied the code from

TobiasBunyan-TPR (approved):

PR #68, opened 2022-10-04
Fixed issue with custom validators not respecting localizer

bentpr:


sussexrick (dismissed):
The checkboxes page in the Umbraco app looks like this:
image

Radios is similar.

bentpr:
> The checkboxes page in the Umbraco app looks like this: image
> 
> Radios is similar.

Resolved - duplicates were being added through the Umbraco app but not the standalone app.

dotcom9 (dismissed):

dotcom9 (approved):

PR #70, opened 2022-10-05
AB#133749 switch to replaceAll

HarryAmmon:


TobiasBunyan-TPR (approved):

sussexrick (commented):
@HarryAmmon This breaks [IE11 compatibility](https://caniuse.com/?search=replaceall).

GOV.UK service manual [says](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices) "You will still need to support Internet Explorer 11 if your analytics data shows at least 2% of your users arriving at the service’s start page are using it."

Our latest data still shows above 2% and the feeling is that traffic is so low outside of scheme return that we can't take a valid poll until DBSR23, so we have to support it. DBSR23 data will likely prove we didn't need to but that is the current data.

HarryAmmon:
@sussexrick resolved your comment in #72

PR #77, opened 2022-10-27
#126933 ignore trailing whitespace in class attribute

dotcom9:
Make XPath query ignore trailing whitespace


GovUk.Frontend.AspNetCore.Extensions/Validation/ClientSideValidationHtmlEnhancer.cs
```diff
@@ -60,7 +60,7 @@ public class ClientSideValidationHtmlEnhancer : IClientSideValidationHtmlEnhance
             if (inputs != null && inputs.Count > 0)
             {
                 // Get the output of the <govuk-input-error-message> grandchild tag helper, if present.
-                var errorMessage = document.DocumentNode.SelectSingleNode("//*[@class='govuk-error-message']");
+                var errorMessage = document.DocumentNode.SelectSingleNode("//*[@class[normalize-space()='govuk-error-message']]");
```
sussexrick: This doesn't work (you need `normalize-space(.)` with a dot to achieve what you wanted). However you have identified a genuine problem and I've come up with this:

`//*[contains(concat(' ', @class, ' '),' govuk-error-message ')]` 

which matches all the following scenarios:

```
class="govuk-error-message"
class="govuk-error-message "
class=" govuk-error-message"
class="other-class govuk-error-message"
class="other-class govuk-error-message other-class"
class="govuk-error-message other-class"
class="other-class      govuk-error-message"
```

dotcom9: Should really be fixed in GovUk.Frontend.AspNetCore. Tested and adopted your suggestion.

sussexrick (approved):

PR #83, opened 2022-11-01
The optional link around the TPR logo was required in Umbraco

sussexrick:
The optional link around the TPR logo was required in Umbraco. Reduce the minimum number of links in the property editor, and ensure fields are mandatory where a link actually is required. AB#142696

Also adds a missing datatype required for configuring the TPR header. AB#142694


bentpr:
Found one other issue with the context bar - if context-2 is single line, a partial border appears?

image

sussexrick:
> Found one other issue with the context bar - if context-2 is single line, a partial border appears?
> 
> image

Thanks @bentpr I'll take that as a separate issue. There will be a following story to tidy up a number of cosmetic issues.

HarryAmmon (approved):

PR #110, opened 2023-01-19
Feature/tpr notification banner

HarryAmmon:
Amends the govuk notification banner.

image

The govuk notification banner has a heading, content and a blocks property where a button, button group or link styled as button can be added and then targeted with Javascript.

EDIT: Updated image and changed text to govuk notification banner


GovUk.Frontend.AspNetCore.Extensions/GovUk.Frontend.AspNetCore.Extensions.csproj
```diff
@@ -235,4 +235,8 @@
 	  <UpToDateCheckInput Remove="Styles\_govuk-table.scss" />
 	</ItemGroup>
 
+	<ItemGroup>
+	  <UpToDateCheckInput Remove="Styles\_tpr-notification-banner.scss" />
```
sussexrick: The file name should be _govuk-notification-banner.scss. The pattern is that adaptations of a GOV.UK component start with _govuk, and components which are invented by TPR start with _tpr.

GovUk.Frontend.AspNetCore.Extensions/HtmlGeneration/ComponentGenerator.TprNotificationBanner.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Microsoft.AspNetCore.Html;
```
sussexrick: You don't need to create a new component for this, just restyle the existing one and add support for buttons.

<govuk-notification-banner> accepts a child of <govuk-button> so there is no block to doing this.

This follows the existing pattern set by all other components except TPR Header and TPR Footer, which are very different to the GOV.UK ones.

It has the following advantages over your approach:

- we diverge less from the base GOV.UK codebase
- the base GOV.UK notification banner gets the additional support for buttons
- converting from GOV.UK style to TPR style is as simple as adding an extra stylesheet, which is already documented in the README
- there is only one Notification Banner type, so documentation is simpler and editor UI is simpler
- you don't need to include a TPR-specific component in a datatype that is supposed to be for GOV.UK components, or alternatively create and maintain a separate TPR datatype
- it's less code because you don't need to repeat the tag helper, view etc

The following are issues with your current build but they would also be solved by customising the existing notification banner:

- "Tpr Notification Banner" document type should have TPR capitalised and be in the TPR folder, not in the GOV.UK folder, _but you can ignore this because you don't need the document type at all_
- In the Umbraco example app the 'Notification banner' page loads like this, _but you can ignore this because you wouldn't need to add a new supported component to the data type_:
Notification banner page in edit view shows Unsupported

GovUk.Frontend.AspNetCore.Extensions/Styles/_tpr-notification-banner.scss
```diff
@@ -0,0 +1,40 @@
...
+$tpr-notification-banner__heading-line-height: 1.5em;
+$tpr-notification-banner__content-line-height: 1.5em;
+$tpr-notification-banner-font-size: 1.4em;
+
+.tpr-notification-banner{
+    background-color: $tpr-colour-eastern-blue;
+    padding: govuk-spacing(6);
+    margin-bottom: govuk-spacing(8);
+}
+.tpr-notificaiton-banner__content {
```
sussexrick: typo on "notification"

GovUk.Frontend.AspNetCore.Extensions/Styles/_tpr-notification-banner.scss
```diff
@@ -0,0 +1,40 @@
+﻿@import '_variables.scss';
+@import 'govuk/base';
+
+$tpr-notification-banner__heading-line-height: 1.5em;
```
sussexrick: Defintely em not rem? There are times when you want em, but usually you want rem. The effect of em changes depending on the size of the surrounding text. rem is applied consistently based on the base font size set by the user in their browser (usually the default of 16px).

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/Test.generated.cs
```diff
@@ -0,0 +1,60 @@
...
+// <auto-generated>
+//   This code was generated by a tool.
+//
+//    Umbraco.ModelsBuilder.Embedded v10.0.0+e3f4b86
+//
+//   Changes to this file will be lost if the code is regenerated.
+// </auto-generated>
+//------------------------------------------------------------------------------
+
+using System;
```
sussexrick: This will need to go...

GovUk.Frontend.AspNetCore.Extensions/Styles/_tpr-notification-banner.scss
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+.tpr-notification-banner__heading {
+    font-family: $govuk-font-family;
+    font-weight: $tpr-font-weight-bold;
+    line-height: $tpr-notification-banner__heading-line-height;
+    font-size: $tpr-notification-banner-font-size;
+    margin: 0 0 govuk-spacing(2) 0;
+}
```
sussexrick: There are significant contrast issues that violate WCAG 1.4.11 https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html 
> "For active controls any visual information provided that is necessary for a user to identify that a control is present and how to operate it must have a minimum 3:1 contrast ratio with the adjacent colors"

I haven't run that contrast check so you would need to (or Jodi, because you need to work with her on this) but to my eye the following look like they would fail.

The default state of the default button and cancel link, and the outline of the secondary button:

Default state of buttons

The hover state of the default button:

Buttons showing default button in hover state

The hover state of the secondary button:

Buttons showing secondary button in hover state

The hover state of the cancel link:

Buttons showing cancel link in hover state

This also applies to links included in the text of the notification banner, which are fine on the GOV.UK implementation because they are on a white background.

GovUk.Frontend.ExampleApp/Views/NotificationBanner/Index.cshtml
```diff
@@ -16,6 +16,30 @@
...
+</tpr-notification-banner>
+
+<p class="govuk-body">Below is an example of how the TPR Notification Banner might be used with a button to dismiss the banner as part of a page.</p>
+
+<tpr-notification-banner>
+    <p class="tpr-notification-banner__heading">
+        You will be logged out in 30 seconds due to inactivity
+    </p>
+    <p class="govuk-body">Click the button below to dismiss the message and continue.</p>
+    <govuk-button class="govuk-button--secondary">Dismiss</govuk-button>
```
sussexrick: There is no guidance here about whether this would be a JS-only feature. If it is then you would want to render the Umbraco-configurable details as `data-*` attributes and then have JS that picked up on those and turned them into a button.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/TPRNotificationBannerBlockList.config
```diff
@@ -0,0 +1,55 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
sussexrick: In this data type you are not configuring the label property. When you edit the data type and click on 'Button' or 'Link, styled as a button' it should have a custom label. Compare to the 'GOV.UK Block list' data type:

Settings for including a 'Button' in 'GOV.UK Block list'

Following my other comments the name of this data type would become "GOV.UK Notification banner block list"

sussexrick (changes requested):

GovUk.Frontend.ExampleApp/Views/NotificationBanner/Index.cshtml
HarryAmmon: I've added guidance about this being a JS only feature, however I don't think I understand what you mean by having the Umbraco-configurable details as `data-*`. Can you explain more or is there an example I can follow?

GovUk.Frontend.AspNetCore.Extensions/HtmlGeneration/ComponentGenerator.TprNotificationBanner.cs
HarryAmmon: I have deleted the TPR Notification banner and made adaptions to the govuk notification banner. I've added support for buttons, button groups and links. I have retained the ability to have a success and default styled banner.

GovUk.Frontend.AspNetCore.Extensions/Styles/_tpr-notification-banner.scss
HarryAmmon: Jodi has said that only a reversed button and a link would be inside a notification banner. I have added the styles required for a reversed button and added the configuration in Umbraco so it can be made a reversed button with the radio buttons like warning and secondary buttons. I used this [prototype](https://xd.adobe.com/view/5ba7aba5-bf2b-431b-7084-513a0af3106a-2d96/screen/62ef7c21-8187-42c3-818b-2e09dada8feb/specs/?hints=off).

Jodi also said the hover state does not fail WCAG 1.4.11 as non-text contrast is not required to reach the 3:1 contrast ratio with the background.

sussexrick: There is still a problem with the unvisited link style in the success banner:
success banner showing unvisited link

Tab to the dismiss button on either banner. Press and hold space and you will see the focus + active state. which has insufficient text contrast:

focus plus active state of button

On the hover state, I disagree with Jodi as I think the [Understanding Non-text contrast guidance](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) is quite clear, under the 'Active User Interface Components' heading:

> For active controls any visual information provided that is necessary for a user to identify that a control is present and how to operate it must have a minimum 3:1 contrast ratio with the adjacent colors. Also, any visual information necessary to indicate state, such as whether a component is selected or focused must also ensure that the information used to identify the control in that state has a minimum 3:1 contrast ratio.

The hover state is now purple on both backgrounds, which does not happen in the prototype you refer to. It does not meet the 3:1 ratio against its adjacent colours, therefore there isn't sufficient contrast to show that the control is present in the hover state.

GovUk.Frontend.ExampleApp/Views/NotificationBanner/Index.cshtml
sussexrick: Now that you've made your example a bit more focussed I don't think you need to use any `data-*` attributes. But your example should degrade gracefully in a no-JS scenario as people will be following your example. For a dismiss button you can do that by assigning it a class of `govuk-!-display-none` so that it doesn't display for a no-JS scenario, then have JS remove that class so that it does display when JS is available. It would also be good to have the second Dismiss button similarly wired up.

GovUk.Frontend.AspNetCore.Extensions/Styles/_govuk-button.scss
```diff
@@ -81,4 +82,24 @@ input[type="file"]::file-selector-button {
 .govuk-button--warning:disabled {
     color: $tpr-colour-white;
     background-color: $tpr-colour-petite-orchid;
+}
+.govuk-notification-banner--success{
+    > * .govuk-button--reversed{
```
sussexrick: I'm not sure the * selector is doing anything useful here?

GovUk.Frontend.AspNetCore.Extensions/Styles/_govuk-button.scss
```diff
@@ -81,4 +82,24 @@ input[type="file"]::file-selector-button {
...
+        color: $tpr-colour-camarone;
+    }
+}
+
+.govuk-button--reversed {
```
sussexrick, line 87: You've got this button defined with different colours by default and when used inside a notification banner. I think we need to see an example on the 'Buttons' page.

GovUk.Frontend.AspNetCore.Extensions/Styles/_govuk-notification-banner.scss
```diff
@@ -0,0 +1,113 @@
...
+    background-color: $tpr-colour-eastern-blue;
+
+    &.govuk-notification-banner--success {
+        border-color: $tpr-colour-camarone;
+        background-color: $tpr-colour-camarone;
+
+        > .govuk-notification-banner__content {
+            background-color: $tpr-colour-camarone;
+
+            > .govuk-button--reversed {
```
sussexrick: I think you'd be safe to use a descendant selector rather than a child selector (ie remove the >) as my understanding is any button inside this banner should be reversed. Being so specific is forcing you to repeat the style below unnecessarily.

GovUk.Frontend.AspNetCore.Extensions/Styles/_govuk-notification-banner.scss
```diff
@@ -0,0 +1,113 @@
...
+                color: $tpr-colour-camarone;
+            }
+
+            > .govuk-grid-row {
+                display: flex;
```
sussexrick: display: flex / justify-content: center is duplicated below under .govuk-notification-banner__content. You don't need this copy.

GovUk.Frontend.AspNetCore.Extensions/Styles/_govuk-notification-banner.scss
```diff
@@ -0,0 +1,113 @@
...
+
+}
+
+.govuk-notification-banner__content {
+    background-color: $tpr-colour-eastern-blue;
+    color: $tpr-colour-white;
+    padding: govuk-spacing(6);
+
+    > .govuk-body {
+    margin-bottom: govuk-spacing(4);
```
sussexrick: Why redefine the margin on .govuk-body specifically in this context? I would expect we would want .govuk-body text spacing to be the same globally.

GovUk.Frontend.AspNetCore.Extensions/Styles/_govuk-notification-banner.scss
```diff
@@ -0,0 +1,113 @@
...
+    color: $tpr-colour-white;
+    }
+
+    > * {
+        max-width: none;
+    }
+
+    > .govuk-button, .govuk-button-group {
+        margin-top: 20px;
```
sussexrick: I do not see this having any effect on .govuk-button. On your example page there is a larger gap between the text and the buttons on the green banner than there is on the blue.

By the way, as a guideline we're setting vertical margins on the bottom only to avoid dealing with margin collapsing, but there will be cases where we need to set it and this may be one of them.

GovUk.Frontend.AspNetCore.Extensions/Styles/_govuk-notification-banner.scss
```diff
@@ -0,0 +1,113 @@
...
+        display: flex;
+        justify-content: center;
+
+        > .govuk-grid-column {
+
+            > .govuk-button {
+                margin-bottom: 0;
+            }
+
+            > .govuk-button-group {
```
sussexrick: Again you probably don't need child selectors here. You could simplify the CSS by using descendant selectors. 

You only need the specificity of child selectors if you're worried about nesting elements inside another one of the same element, but that's not going to happen with buttons and button groups.

GovUk.Frontend.AspNetCore.Extensions/Styles/_govuk-notification-banner.scss
```diff
@@ -0,0 +1,113 @@
...
+                > .govuk-button {
+                    margin-bottom: 0;
+                }
+            }
+        }
+    }
+}
+
+.govuk-notification-banner__link, 
+.govuk-notification-banner > * .govuk-link {
```
sussexrick: `> *` isn't doing anything useful and could be removed

GovUk.Frontend.ExampleApp/Views/Button/Index.cshtml
```diff
@@ -12,4 +12,15 @@
     <govuk-button class="govuk-button--secondary" type="button">Secondary button</govuk-button>
     <govuk-button class="govuk-button--warning" type="button">Warning button</govuk-button>
     <a href="/" class="govuk-link">Cancel</a>
-</div>
\ No newline at end of file
+</div>
+
+
+<govuk-notification-banner>
```
sussexrick, line 18: Looking at these examples in the browser you have contrast issues for various states:

Hover:
hover state
Focus:
focus state

Focus + active:
focus + active

The styles here should not be showing up differently to those in the Umbraco example, so clearly there is a difference between them and the styling needs to handle both. It may be because the Umbraco green banner uses a button group and this doesn't, but that suggests you also need to test the blue banner with a button group to see what happens there.

GovUk.Frontend.ExampleApp/Views/NotificationBanner/Index.cshtml
```diff
@@ -7,13 +7,16 @@
     <p class="govuk-notification-banner__heading">
         A <a class="govuk-notification-banner__link" href="https://design-system.service.gov.uk/components/notification-banner/">notification banner</a> lets you tell the user about something that's not directly relevant to the thing they’re trying to do on that page of the service.
```
sussexrick, line 8: This link has insufficient contrast:

image

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/GovukSummaryListItem.generated.cs
```diff
@@ -62,15 +62,15 @@ public GovukSummaryListItem(IPublishedElement content, IPublishedValueFallback p
 		///</summary>
 		[global::System.CodeDom.Compiler.GeneratedCodeAttribute("Umbraco.ModelsBuilder.Embedded", "10.0.0+e3f4b86")]
 		[global::System.Diagnostics.CodeAnalysis.MaybeNull]
-		[ImplementPropertyType("itemKey")]
```
sussexrick: This is a regression. This property was changed from 'key' to 'itemKey' to avoid a clash with the built-in Umbraco 'Key' property.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/short-heading.config
```diff
@@ -0,0 +1,48 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
sussexrick: This looks like a temporary test page that's crept into the PR?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/GOVUKNotificationBanner.config
```diff
@@ -0,0 +1,55 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
sussexrick: This looks like a duplicate of the file below. The DataType Key attribute identifies the same data type, but the name in this copy is wrong.

GovUk.Frontend.Umbraco/uSync/v9/ContentTypes/govuknotificationbanner.config
```diff
@@ -21,16 +21,32 @@
...
+      <LabelOnTop>false</LabelOnTop>
+    </GenericProperty>
     <GenericProperty>
       <Key>c058f0d1-8ae3-4cac-97d9-b3b561f5e924</Key>
       <Name>Heading</Name>
       <Alias>heading</Alias>
       <Definition>f59f7848-da0b-4eb5-8f57-0b96bca1264b</Definition>
       <Type>Umbraco.TinyMCE</Type>
-      <Mandatory>true</Mandatory>
+      <Mandatory>false</Mandatory>
```
sussexrick, line 46: This probably should still be mandatory, unless Jodi has told you otherwise. But if it isn't you need to update the view because it always renders an h3 at the moment even if there is no content.

sussexrick:
The diff shows a change in the submodule. Did you update it? The version of govuk-frontend needs to stay aligned with that used by govuk-frontend-aspnetcore otherwise there may be unintended consequences.

HarryAmmon:
> The diff shows a change in the submodule. Did you update it? The version of govuk-frontend needs to stay aligned with that used by govuk-frontend-aspnetcore otherwise there may be unintended consequences.

I don't recall updating the submodule, however I have now updated the submodule to version 4.3.0, which is the same as govuk-frontend-aspnetcore uses. Running `git submodule status` on this branch and the main branch of govuk-frontend-aspnet returns the same commit ID

GovUk.Frontend.Umbraco/uSync/v9/ContentTypes/govuknotificationbanner.config
HarryAmmon: Jodi has said there could be scenarios where we don't want a heading but do want a body. I have updated the view so the `h3` isn't rendered if the heading isn't present

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/GOVUKNotificationBanner.config
HarryAmmon: This file has been removed

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/short-heading.config
HarryAmmon: This file has been removed

GovUk.Frontend.ExampleApp/Views/Button/Index.cshtml
HarryAmmon: These styles have been fixed to show either white or black text depending on the background colour and the behaviour is now consistent between the Umbraco and the Govuk example apps

GovUk.Frontend.ExampleApp/Views/NotificationBanner/Index.cshtml
HarryAmmon: Links now have sufficient contrast, they are always white on the green and blue backgrounds

GovUk.Frontend.ExampleApp/Views/NotificationBanner/Index.cshtml
HarryAmmon: Both buttons are hidden when JS is unavailable and when JS is available they both dismiss their banner

GovUk.Frontend.AspNetCore.Extensions/Styles/_tpr-notification-banner.scss
HarryAmmon: The text contrast for unvisited links and the focus + active button state has been resolved.

I have raised the hover contrast issue with Jodi again however she still disagrees. Also when demoing the reversed button in the notification banner, Jodi did not raise any concerns. Maybe this is something to pick up again later?

sussexrick: I've discussed with Jodi and reviewed more WCAG guidance and I now agree that the design passes.

sussexrick:
Regarding the submodule I've set it back to v4.2.0 which is aligned with govuk-frontend-aspnetcore beta 3 which this branch is using. It removes the submodule update from this PR.

sussexrick (approved):

PR #118, opened 2023-02-28
Create NuGet packages

sussexrick:
So far we have mostly been using this repo with project references. This creates NuGet packages and publishes them to nuget.org so that we can do it properly, and have better control of versioning across multiple applications.

**DO NOT MERGE until consuming applications still using a project reference have been updated to use a branch other than develop**

Changes are:
- updated vulnerable and deprecated dependencies, including jQuery Validate
- updated Umbraco to 10.4.0
- updated the pipeline to publish new packages to nuget.org, only when the option to push is manually selected and only from a `develop` or `release` branch.
- automates installing the govuk-frontend submodule to make initial setup easier
- flags out-of-date Umbraco ModelsBuilder models as a build warning
- makes it possible to use GOV.UK and TPR SASS variables and utilities in consuming applications, rather than hard-coding values in plain CSS
- adds support for secondary warning buttons - these were previously mutually exclusive in Umbraco and this is a breaking change
- adds support for custom CSS classes on all components that have a root element (ie everything but typography), and consistently places those classes on the root element (text input was inconsistent - that is a breaking change when updating consuming applications)
- on text input, adds a new setting for the field width based on t-shirt sizes, rather than having to look up a class on GOV.UK
- fixes some CSS issues
- moves the countdown message in the character count component to the right
- adds new partial views for header and footer lockups, to reduce code duplication across TPR applications
- adds more Markdown documentation, and more examples in the Umbraco example application
- adds a switch in the Umbraco example application so you opt to see the plain GOV.UK styles instead of the TPR ones
- adds the value attribute to buttons, which means code can detect which of multiple buttons on a page was clicked (for [AB#145273](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/145273))

[AB#146129](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/146129)


dotcom9 (dismissed):

dotcom9 (approved):

GovUk.Frontend.Umbraco/Models/TprHeaderLockupModel.cs
```diff
@@ -0,0 +1,35 @@
+﻿using Umbraco.Cms.Core.Models;
+using Umbraco.Cms.Core.Models.PublishedContent;
+using Umbraco.Extensions;
+
+namespace GovUk.Frontend.Umbraco.Models
+{
+    public class TprHeaderLockupModel
```
HarryAmmon, line 7: @sussexrick sorry old PR but why is it called Lockup? Is it not a Lookup?

sussexrick: [Lockup is a term from the design world](https://developer.apple.com/design/human-interface-guidelines/components/layout-and-organization/lockups/) for several components combined into one - in this case the skip link, phase banner,  header bar and context bar.

PR #131, opened 2023-03-15
Feature/127262 value for members

dotcom9:


sussexrick (approved):
If github offers the option it would be good to squash commit the to-and-fro and add a new commit message saying what's actually happening.

PR #137, opened 2023-04-03
A partial view to implement GOV.UK guidance on pagination (project refs)

sussexrick:
Pagination support so far has been just importing the styles and using the basic component from govuk-frontend-aspnetcore.

This adds a partial view which implements the paging logic, including guidance from the GOV.UK Design System, so that it does not need to be reimplemented in each application.

It also updates the example page in the ASP.NET example app to use the new partial view and explain what other implementation steps an application would need.

[AB#151609](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/151609)

### DIFF
--- a/GovUk.Frontend.AspNetCore.Extensions/Models/PaginationModel.cs
+++ b/GovUk.Frontend.AspNetCore.Extensions/Models/PaginationModel.cs
@@ -1,0 +1,81 @@
+﻿using Microsoft.AspNetCore.Http;
+using System;
+
+namespace GovUk.Frontend.AspNetCore.Extensions.Models
+{
+    /// <summary>
+    /// Configuration settings for an instance of the GOV.UK Design System Pagination component.
+    /// </summary>
+    public class PaginationModel
+    {
+        /// <summary>
+        /// The current page.
+        /// </summary>
+        public int PageNumber { get; set; } = 1;
+
+        /// <summary>
+        /// The number of items per page. If <c>null</c> then pagination is not active.
+        /// </summary>
+        public int? PageSize { get; set; }
+
+        /// <summary>
+        /// The total number of items to be paginated.
+        /// </summary>
+        public int TotalItems { get; set; }
+
+        /// <summary>
+        /// The total number of pages based on <see cref="TotalItems"/> and <see cref="PageSize"/>
+        /// </summary>
+        public int TotalPages()
+        {
+            if (!PageSize.HasValue)
+            {
+                return 1;
+            }
+            return (int)Math.Ceiling((decimal)TotalItems / PageSize.Value);
+        }
+
+        /// <summary>
+        /// The label for the navigation landmark that wraps the pagination.
+        /// </summary>
+        public string LandmarkLabel = "results";
+
+        /// <summary>
+        /// The link label which adds context to the 'Previous' link when there is a small number of pages.
+        /// </summary>
+        /// <remarks>{{page}} is replaced with the number of the previous page and {{total}} is replaced with the total number of pages</remarks>
+        public string PreviousPageLabel = "{{page}} of {{total}}";
+
+        /// <summary>
+        /// The link label which adds context to the 'Next' link when there is a small number of pages.
+        /// </summary>
+        /// <remarks>{{page}} is replaced with the number of the next page and {{total}} is replaced with the total number of pages</remarks>
+        public string NextPageLabel = "{{page}} of {{total}}";
+
+        /// <summary>
+        /// The aria-label which gives more context to links which point to individual pages by their page number alone.
+        /// </summary>
+        /// <remarks>{{page}} is replaced with the number of the page</remarks>
+        public string PageVisuallyHiddenText = "Page {{page}}";
+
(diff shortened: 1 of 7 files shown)


GovUk.Frontend.AspNetCore.Extensions/Models/PaginationModel.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <summary>
+        /// The total number of pages based on <see cref="TotalItems"/> and <see cref="PageSize"/>
+        /// </summary>
+        public int TotalPages()
+        {
+            if (!PageSize.HasValue || PageSize <= 0 || TotalItems <= 0)
+            {
+                return 1;
+            }
+            return (int)Math.Ceiling((decimal)TotalItems / PageSize.Value);
```
samuelhaa, line 35: Would an explanation as to why we are using `PageSize.Value` rather than `PageSize` in the division help future users?

GovUk.Frontend.AspNetCore.Extensions/Models/PaginationModel.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        /// <summary>
+        /// The name of the query string parameter used to indicate the current page.
+        /// </summary>
+        public string QueryStringParameter { get; set; } = "page";
+
+        /// <summary>
+        /// The GOV.UK Design System recommends a different styles of pagination for small numbers of pages and large numbers of pages. This threshold controls when the style changes.
+        /// </summary>
+        public int LargeNumberOfPagesThreshold { get; set; } = 10;
```
samuelhaa, line 69: I can't find any references to 10 in the Govuk design system. Is 10 an arbitray number? How do we know that 10 is a good number? Does it matter which number we choose?

GovUk.Frontend.AspNetCore.Extensions/Views/Shared/GOVUK/Pagination.cshtml
```diff
@@ -0,0 +1,90 @@
...
+@addTagHelper *, GovUk.Frontend.AspNetCore
+@model GovUk.Frontend.AspNetCore.Extensions.Models.PaginationModel
+@if (Model != null && Model.PageSize.HasValue && Model.TotalItems > Model.PageSize)
+{
+    var path = Model.Path ?? Context.Request.Path;
+    var queryString = Model.QueryString ?? Context.Request.QueryString;
+    var query = HttpUtility.ParseQueryString(queryString.ToString());
+    query.Remove(Model.QueryStringParameter);
+
+    var totalPages = (int)Math.Ceiling((decimal)Model.TotalItems / Model.PageSize.Value);
```
samuelhaa, line 12: There is a method on the `PaginationModel` called `TotalPages()`. Any reason why we can't use that rather than duplicate the calculation?

GovUk.Frontend.AspNetCore.Extensions/Views/Shared/GOVUK/Pagination.cshtml
```diff
@@ -0,0 +1,90 @@
+﻿@using System.Globalization;
+@using System.Web;
+@addTagHelper *, GovUk.Frontend.AspNetCore
+@model GovUk.Frontend.AspNetCore.Extensions.Models.PaginationModel
+@if (Model != null && Model.PageSize.HasValue && Model.TotalItems > Model.PageSize)
```
samuelhaa, line 5: Is it better to use `Model is not null` rather than `Model != null` as the `is` operator doesn't run any custom equality operators. We could create our own equality operator such that `var a = new Model(); a == null` is `true`. Not that any sane person would do that but it is possible.

I can see that we are targeting net6.0 so backwards compatibility would not be a problem.

GovUk.Frontend.AspNetCore.Extensions/Models/PaginationModel.cs
sussexrick: It is arbitrary because they don't define it, and we don't have user testing for this to prove whether it's the right number or not. However it is overridable should we need that.

GovUk.Frontend.AspNetCore.Extensions/Views/Shared/GOVUK/Pagination.cshtml
sussexrick: Good point and this is something I already spotted for #138, which will replace the version you're looking at in the next sprint.  This branch is needed for now but will soon be obsolete.

GovUk.Frontend.AspNetCore.Extensions/Views/Shared/GOVUK/Pagination.cshtml
sussexrick: That's an interesting C#9.0 feature I didn't know, but in this case as you say it's not going to be important. We're in control of the model class it's checking and we know it doesn't have an overridden equality operator.

samuelhaa (approved):

PR #138, opened 2023-04-04
Add 'Pagination' component / Rename 'Back to menu' component (develop)

sussexrick:
## Pagination

This PR adds full support for the [Pagination component](https://design-system.service.gov.uk/components/pagination/).

Pagination support so far has been just importing the styles and using the basic component from govuk-frontend-aspnetcore.

This adds a partial view which implements the pagination logic, including guidance from the GOV.UK Design System, so that it does not need to be reimplemented in each application.

It also adds the component to the Umbraco block list with all properties configurable.

It also updates the example page in the ASP.NET and Umbraco example apps to use the new partial view and explain what other implementation steps an application would need.

## Back to menu

'Back to menu' was incorrectly categorised as a GOV.UK component rather than a TPR component. It is renamed to fix that, and as that is a breaking change the package versions are incremented to v3.0.0. Also renamed some 'TPR back to top' classes to be clearer that it is also a TPR component.

## Testing support for integer properties in Umbraco 

Adds methods for testing integer properties, eg `SetupUmbracoIntegerPropertyValue<T>`, similar to those for other property types. These were added to test the pagination functionality.

## Adds missing composition for 'TPR back to top' component

The 'TPRFooterLockup' partial view was getting custom text for the 'TPR back to top' compoment from an Umbraco field that did not exist. There is now a composition to add it to the relevant document type.

## Documentation improvements

Added screenshots of GOV.UK styling
Added missing pages for 'TPR back to top' and 'TPR back to menu'
Added support for viewing the ASP.NET example app with GOV.UK or TPR styling. Changed equivalent feature in the Umbraco example app to work the same way.
Removes LocalDB configuration from the Umbraco example app, removing a barrier to running it

[AB#151609](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/151609)
[AB#146129](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/146129)


samuelhaa:
Is there a reason why we have three changes in one PR rather than one PR per change?  What does "Back to menu" have to do with Pagination"? Are they related?

GovUk.Frontend.AspNetCore.Extensions/Models/PaginationModel.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        /// <summary>
+        /// The number of items per page.
+        /// </summary>
+        public int PageSize { get; set; } = 10;
+
+        /// <summary>
+        /// The total number of items to be paginated.
+        /// </summary>
+        public int TotalItems { get; set; }
```
samuelhaa, line 24: If `TotalItems` is a non-nullable property, should it not have a default value, like `PageSize` and `PageNumber`? I know `int` defaults to zero but still...

GovUk.Frontend.AspNetCore.Extensions/TagHelpers/TprBackToTopTagHelper.cs
```diff
@@ -11,7 +11,7 @@ namespace GovUk.Frontend.AspNetCore.Extensions.TagHelpers
     /// </summary>
     [HtmlTargetElement(TagName)]
     [OutputElementHint(ComponentGenerator.BackToTopLinkElement)]
```
samuelhaa, line 13: Should it not be `TprBackToTopLinkElement`  rather than `BackToTopLinkElement`?

sussexrick: I thought about that but didn't do it just to have fewer changes in the PR. Unlike the tag name it's an internal member so it's only important that it's clear within the scope of this project. Happy to change it though if you think it's important.

GovUk.Frontend.AspNetCore.Extensions/Models/PaginationModel.cs
sussexrick: That's the point though - it defaults to 0 which is the default I want. The other examples need a default value that's not the same as the default for the type. 

If you set a default value for something that is the default value for the type anyway, the compiler will warn you that it's an unnecessary thing to do.

sussexrick:
> Is there a reason why we have three changes in one PR rather than one PR per change? What does "Back to menu" have to do with Pagination"? Are they related?

They're not related except insofar as they both target v3.0.0 of the packages. They could've been separate PRs. I wrote documentation for Pagination and was reviewing other documentation with a view to meeting [these criteria](https://design-system.service.gov.uk/community/contribution-criteria/#developing-a-community-resource-or-tool). The documentation was missing for that component and in writing it I noticed the inconsistency. Since I am working on a new version of the packages, and am about to actually use them in projects for the first time, it was the right time to get that breaking change in.

The same applies to the 'Back to top' fix - I was writing the missing documentation and noticed the inconsistency.

samuelhaa:
This is minor but should not `Items` be lowercase in both apps? Will the difference in casing affect the usage? Does it have to be called `Items` in umbraco because it ties to a property with Pascal case?

This is `Govuk.Frontend.ExampleApp`

image

And this is `GovUk.Frontend.Umbraco.ExampleApp`


image

README.md
```diff
@@ -31,3 +37,15 @@ This builds on [ASP.NET Core MVC tag helpers for GOV.UK Design System](https://g
 - [Override property values](docs/umbraco/override-property-values.md)
 - [Umbraco unit testing](docs/umbraco/unit-testing.md)
 - [Use SASS for CSS](docs/aspnet/sass.md)
+
+Umbraco support is published on NuGet as [ThePensionsRegulator.GovUk.Frontend.Umbraco](https://www.nuget.org/packages/ThePensionsRegulator.GovUk.Frontend.Umbraco)
+
+![ThePensionsRegulator.GovUk.Frontend.Umbraco on nuget.org](https://img.shields.io/nuget/vpre/ThePensionsRegulator.GovUk.Frontend.Umbraco)
```
samuelhaa, line 43: Will this badge be increased to v3.0 automatically? It sayd 2.0 at the moment.

sussexrick: Yes

sussexrick:
> This is minor but should not `Items` be lowercase in both apps? Will the difference in casing affect the usage? Does it have to be called `Items` in umbraco because it ties to a property with Pascal case?
> 
> This is `Govuk.Frontend.ExampleApp`
> 
> image
> 
> And this is `GovUk.Frontend.Umbraco.ExampleApp`
> 
> image

It's because it's bound to a property, and C# naming conventions mean that property should be pascal case. It works either way - that's the nature of querystring parameters on the web.

samuelhaa (approved):

PR #139, opened 2023-04-06
Do not include character count error in error summary - develop

HarryAmmon:
When a user has exceeded the maximum number of characters allowed in a text area, two error messages appear in the error summary.  These error messages describe the same problem, so only one should be there. 

Also, there was an issue with the second error message where it would not link to the text area.
## Before
image
## After
image

### DIFF
--- a/GovUk.Frontend.AspNetCore.Extensions/GovUk.Frontend.AspNetCore.Extensions.csproj
+++ b/GovUk.Frontend.AspNetCore.Extensions/GovUk.Frontend.AspNetCore.Extensions.csproj
@@ -6,7 +6,7 @@
 		<GenerateEmbeddedFilesManifest>true</GenerateEmbeddedFilesManifest>
 		<IncludeRazorContentInPack>true</IncludeRazorContentInPack>
 		<Nullable>enable</Nullable>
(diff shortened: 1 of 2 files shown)


samuelhaa:
Please update the description and answer why we are need this change and give some background. We know what you have done but not why. We need some wider context as to why the change is needed. If it is part of the gds spec, please include a link to that. This will help future users.

sussexrick (dismissed):

sussexrick:
@HarryAmmon You need to increment the package version in GovUk.Frontend.AspNetCore.Extensions.csproj to 3.0.1

You also need to cherry-pick your commit and create a PR into the `release/2023-03-01-project-references` branch if you want this to move across the board ahead of my NuGet package work.

HarryAmmon:
> @HarryAmmon You need to increment the package version in GovUk.Frontend.AspNetCore.Extensions.csproj to 3.0.1
> 
> You also need to cherry-pick your commit and create a PR into the `release/2023-03-01-project-references` branch if you want this to move across the board ahead of my NuGet package work.

I have updated to 3.0.1

sussexrick (approved):

PR #147, opened 2023-04-18
#154223 fix width from desktop bug for character count fields

HarryAmmon:
Changes:
- Allows the width of a character count component to be changed with the column size property


sussexrick:
You also need to increment the package version to 3.0.3

sussexrick (approved):

PR #153, opened 2023-05-18
Improvements to date validation

sussexrick:
Increments the package version for ThePensionsRegulator.GovUk.Frontend and ThePensionsRegulator.GovUk.Frontend.Umbraco to reflect new features with non-breaking changes.

## Features

Adds a new `[DateRange]` attribute, with examples and documentation. This wraps the standard `[Range]` attribute, but allows it to work with the `DateOnly` type. It also prevents the default client-side validation attributes from being rendered, which do not work with the GOV.UK date input design because they expect one field containing a full date, whereas GOV.UK design uses three fields each with a partial date.

Adds a field to Date input settings in Umbraco to configure the error message for range validation.

Updates the Date input component in Umbraco to re-populate the entered data in a date field when a page fails validation, rather than requiring the user to re-type it. This only works if the data is numeric, because the underlying govuk-frontend-aspnetcore library is using an integer property rather than a string.

Adds a 'Display name' property to Date input settings in Umbraco which, together with some dictionary entries, allows customisation of the error messages for dates which are not parseable or not possible (eg 30 Feb). It is structured so that the defaults will usually be fine if a property name is one word, and if not then just changing the display name should be enough. Dictionary entries to completely change the error messages are unlikely to be needed, but possible to add if they are.

Client-side validation for partially-entered dates does behave differently to server-side. If you enter, for example, a day and month and but no year, on the client-side that will show the error for a required field validation. On the server-side that will show a specific error saying that the year is missing. However this should be acceptable as both messages make sense and an individual user is very unlikely to see both.

## Documentation

Improves README.md at the root of the repo to list the components we support, which is not currently every component in the GOV.UK Design System.

## Bug fixes

Fixes pages in the ASP.NET example app which were displaying two error summary components due to recent improvements in the govuk-frontend-aspnetcore library that we consume.


GovUk.Frontend.AspNetCore.Extensions/Validation/DateRangeAttribute.cs
```diff
@@ -0,0 +1,37 @@
...
+    public class DateRangeAttribute : ValidationAttribute
+    {
+        private readonly RangeAttribute _rangeAttribute;
+
+        /// <summary>
+        /// Create a new <see cref="RangeAttribute"/>
+        /// </summary>
+        /// <param name="minimum">An ISO 8601 date string in the format YYYY-MM-DD</param>
+        /// <param name="maximum">An ISO 8601 date string in the format YYYY-MM-DD</param>
+        public DateRangeAttribute(string minimum, string maximum)
```
samuelhaa, line 23: Would it be useful to have a constructor that takes two `DateTime` objects or is that not necessary?

samuelhaa: No, it is not necessary because an attribute requires that values to be known at compile time meaning we can't give it two objects.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-be-a-real-date.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Dictionary Key="4b578c0a-3139-4086-937c-407c42c7fc3e" Alias="Date must be a real date" Level="0">
+  <Info />
+  <Translations>
+    <Translation Language="de">{0} muss ein echtes Datum sein</Translation>
```
samuelhaa, line 5: Any particular reason why we have the german version as well now? Does it make a difference?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-include-a-day-and-month.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Dictionary Key="9b06ec3e-f82e-4baf-bdb3-f86a9e15f653" Alias="Date must include a day and month" Level="0">
+  <Info />
+  <Translations>
+    <Translation Language="de">{0} muss einen Tag und einen Monat enthalten</Translation>
```
samuelhaa, line 5: Any particular reason why we have the german version as well now? Does it make a difference?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-include-a-day-and-year.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Dictionary Key="eafeca44-f744-488f-b9f4-eb453adf0f15" Alias="Date must include a day and year" Level="0">
+  <Info />
+  <Translations>
+    <Translation Language="de">{0} muss einen Tag und ein Jahr enthalten</Translation>
```
samuelhaa, line 5: Any particular reason why we have the german version as well now? Does it make a difference?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-include-a-day.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Dictionary Key="78605612-0c8e-4fd7-88ab-79629c46daf6" Alias="Date must include a day" Level="0">
+  <Info />
+  <Translations>
+    <Translation Language="de">{0} muss einen Tag enthalten</Translation>
```
samuelhaa, line 5: Any particular reason why we have the german version as well now? Does it make a difference?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-include-a-month-and-year.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Dictionary Key="a9577e33-7d64-4909-a9fd-1f7059c57f3d" Alias="Date must include a month and year" Level="0">
+  <Info />
+  <Translations>
+    <Translation Language="de">{0} muss einen Monat und ein Jahr enthalten</Translation>
```
samuelhaa, line 5: Any particular reason why we have the german version as well now? Does it make a difference?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-include-a-month.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Dictionary Key="2e0c27ef-c31a-4d68-82d6-0143a9d9a9a6" Alias="Date must include a month" Level="0">
+  <Info />
+  <Translations>
+    <Translation Language="de">{0} muss einen Monat enthalten</Translation>
```
samuelhaa, line 5: Any particular reason why we have the german version as well now? Does it make a difference?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-include-a-year.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Dictionary Key="19821494-b040-4f31-b979-31d9fa782ea4" Alias="Date must include a year" Level="0">
+  <Info />
+  <Translations>
+    <Translation Language="de">{0} muss ein Jahr enthalten</Translation>
```
samuelhaa, line 5: Any particular reason why we have the german version as well now? Does it make a difference?

GovUk.Frontend.Umbraco.Tests/UmbracoDateInputModelBinderTests.cs
```diff
@@ -0,0 +1,370 @@
...
+        [TestCase("1", "4", "-1", DateInputParseErrors.InvalidYear)]
+        [TestCase("1", "4", "10000", DateInputParseErrors.InvalidYear)]
+        [TestCase("1", "4", "x", DateInputParseErrors.InvalidYear)]
+        public void Parse_InvalidDate_ComputesExpectedParseErrors(
+            string day, string month, string year, DateInputParseErrors expectedParseErrors)
+        {
+            // Arrange
+
+            // Act
+            var result = UmbracoDateInputModelBinder.Parse(day, month, year, out var dateComponents);
```
samuelhaa, line 263: `UmbracoDateInputModelBinder.Parse()` accepts `null` for day, month and year. Could you please add some extra test cases that tests for `nulls`.

GovUk.Frontend.Umbraco/ModelBinding/UmbracoDateInputModelBinder.cs
```diff
@@ -0,0 +1,189 @@
...
+            if (!missingDay && missingMonth && missingYear)
+            {
+                return string.Format(umbracoDictionary[DictionaryConstants.DateMustIncludeAMonthAndYear] ?? "{0} must include a month and year", displayName).Trim();
+            }
+
+            return string.Format(umbracoDictionary[DictionaryConstants.DateMustBeARealDate] ?? "{0} must be a real date", displayName).Trim();
+        }
+
+        // internal for testing
+        internal static DateInputParseErrors Parse(string? day, string? month, string? year, out DateOnly? date)
```
samuelhaa, line 148: Does `date` need to be nullable here since you are always assigning it a value in line 185?

GovUk.Frontend.Umbraco.Tests/UmbracoDateInputModelBinderTests.cs
```diff
@@ -0,0 +1,370 @@
...
+        [TestCase("0", "4", "2020")]
+        [TestCase("-1", "4", "2020")]
+        [TestCase("32", "4", "2020")]
+        [TestCase("1", "0", "2020")]
+        [TestCase("1", "-1", "2020")]
+        [TestCase("1", "13", "2020")]
+        [TestCase("1", "4", "0")]
+        [TestCase("1", "4", "-1")]
+        [TestCase("1", "4", "10000")]
+        public async Task BindModelAsync_MissingOrInvalidComponents_FailsBinding(string day, string month, string year)
```
samuelhaa, line 114: Do we need test casees for where `day`,`month` and `year` is a string, e.g. `"x"` or some other characters as in the test cases for `Parse_InvalidDate_ComputesExpectedParseErrors()`?

GovUk.Frontend.AspNetCore.Extensions/Validation/DateRangeAttribute.cs
samuelhaa: I don't seem ti find any tests for the daterange attribute itself. Do we need that or is it overed by the other tests implicitly?

samuelhaa (dismissed):
I've run the example app and tried different sets of dates and it seems to work

GovUk.Frontend.AspNetCore.Extensions/Validation/DateRangeAttribute.cs
sussexrick: It's just a wrapper around the [Range] attribute which is tested by Microsoft. We don't need to re-test their code. I took the view that the actual code in this attribute was simple enough that the effort of writing the tests would outweigh the value.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-be-a-real-date.config
sussexrick: I was testing that the multi-lingual support works.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-include-a-day-and-month.config
sussexrick: I was testing that the multi-lingual support works.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-include-a-day-and-year.config
sussexrick: I was testing that the multi-lingual support works.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-include-a-day.config
sussexrick: I was testing that the multi-lingual support works.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-include-a-month.config
sussexrick: I was testing that the multi-lingual support works.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-include-a-year.config
sussexrick: I was testing that the multi-lingual support works.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/date-must-include-a-month-and-year.config
sussexrick: I was testing that the multi-lingual support works.

GovUk.Frontend.Umbraco/ModelBinding/UmbracoDateInputModelBinder.cs
sussexrick: If there are errors then on line 185 it is assigned the default value of `Nullable<DateOnly>` which is `null`.

GovUk.Frontend.AspNetCore.Extensions/Validation/DateRangeAttribute.cs
sussexrick: Tests added now though

GovUk.Frontend.Umbraco.Tests/UmbracoDateInputModelBinderTests.cs
sussexrick: Done

GovUk.Frontend.Umbraco.Tests/UmbracoDateInputModelBinderTests.cs
sussexrick: Done

samuelhaa (approved):

PR #155, opened 2023-05-26
AB#146422 add unconfirmed task list section

HarryAmmon:
Adds "unconfirmed" as a task list status 
image


sussexrick:
This is the wrong approach, as this enum is the list of statuses that are part of the GOV.UK Design System.

I have [provided support for custom task statuses](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/docs/components/task-list.md) in plain ASP.NET by doing something like this:

`<govuk-task-list-task-status status="Completed">Unconfirmed</govuk-task-list-task-status>`

However that's not currently something you can configure in Umbraco, so the right approach would involve adding Umbraco support for that.

The [current contents of the `<govuk-task-list-task-status />` tag](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/GovUk.Frontend.Umbraco/Views/Partials/GOVUK/GovUkTaskList.cshtml) uses a dictionary value by default, and falls back to a plain English version of the enum. That allows customising the name used for a given enum value.

This use case requires another status, or effectively two sub-statuses within the 'In progress' status. So I think it would have to work with a similar fallback approach:

- if the task has status text configured in Umbraco, use that
- second preference, use the application-wide dictionary value
- if that's missing, use the plain English version of the enum

HarryAmmon:
I have added a new property on the 'task' block called 'Status text'. Setting a value here will override the value from the dictionary which makes the status independent from the status text.

This property can be overrode in code.

GovUk.Frontend.Umbraco.ExampleApp/Controllers/TaskListController.cs
```diff
@@ -31,6 +31,7 @@ public override IActionResult Index()
                                                          x.Settings.Value<string>(nameof(GovukTaskSettings.CssClasses))!.Contains("yet-another-thing")) as OverridableBlockListItem;
             if (target != null)
             {
+                target.Content.OverrideValue(nameof(GovukTask.StatusText), "Done");
```
sussexrick: I would expect this property to appear in the settings of 'Task list Task' rather than the content, because that's where the status field is. I would expect 'Status text' to be after 'Status' because it customises the effect of setting 'Status'.

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/GovukTask.generated.cs
```diff
@@ -65,6 +65,14 @@ public GovukTask(IPublishedElement content, IPublishedValueFallback publishedVal
 		[ImplementPropertyType("link")]
 		public virtual global::Umbraco.Cms.Core.Models.Link Link => this.Value<global::Umbraco.Cms.Core.Models.Link>(_publishedValueFallback, "link");
 
+		///<summary>
+		/// Status text: If populated, this will override the text that comes from the dictionary when using the task status setting
```
sussexrick: The convention for help text in this library is "Defaults to xxx if left blank", so I would suggest rewriting the help text to "Defaults to the value of 'Status' in sentence case if left blank. Default status text can be customised in the dictionary."

GovUk.Frontend.Umbraco/Views/Partials/GOVUK/GovUkTaskList.cshtml
```diff
@@ -20,6 +20,7 @@
         {
             var hint = GovUkTypography.Apply(task.Content.Value<string>("hint"), new TypographyOptions { RemoveWrappingParagraph = true });
             var link = task.Content.Value<Link>("link")?.Url;
+            var statusText = GovUkTypography.Apply(task.Content.Value<string>("statusText"), new TypographyOptions { RemoveWrappingParagraph = true });
```
sussexrick: I would expect this to be a plain text field, in which case you don't need to apply GOV.UK typography. Unless you think there is a genuine use case for including HTML?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/task-list.config
```diff
@@ -77,7 +77,7 @@
...
...nt\": \"\"\r\n    },\r\n    {\r\n      \"contentTypeKey\": \"28cfc6cc-1e02-4d9e-8232-b2f921026042\",\r\n      \"udi\": \"umb://element/d299b34cca2e47079734a25d20cc9009\",\r\n      \"taskName\": \"Yet another thing to do\",\r\n      \"link\": \"[{\\\"name\\\":\\\"#\\\",\\\"url\\\":\\\"#\\\"}]\",\r\n      \"hint\": \"<p>This thing has an <strong>important deadline</strong> you need to know about. It is set to 'Completed' by code in <em>TaskListController.</em></p>\"\r\n    }\r\n  ],\r\n  \"settingsData\": [\r\n    {\r\n      \"contentTypeKey\": \"dac47aa8-2d8c-4024-8b72-118d15916b4c\",\r\n      \"udi\": \"umb://element/bab17a84a7fb4a3f98c7c62ca24ed6b5\",\r\n      \"cssClasses\": \"\",\r\n      \"status\": \"[\\\"Not started\\\"]\"\r\n    },\r\n    {\r\n      \"contentTypeKey\": \"dac47aa8-2d8c-4024-8b72-118d15916b4c\",\r\n      \"udi\": \"umb://element/59ad6b4519614f2bac5f1673e62ab178\",\r\n      \"status\": \"[\\\"Incomplete\\\"]\"\r\n    },\r\n    {\r\n      \"contentTypeKey\": \"dac47aa8-2d8c-4024-8b72-118d15916b4c\",\r\n      \"udi\": \"umb://element/13b1896f2f094ccb96daeecd8e9033d1\",\r\n      \"status\": \"\",\r\n      \"cssClasses\": \"yet-another-thing\"\r\n    }\r\n  ]\r\n}"
```
sussexrick, line 80: The uSync data for the content type is missing. There is no 'status text' property without it.

docs/components/task-list.md
```diff
@@ -74,6 +74,8 @@ Add the 'Task list' component to a block list, and add 'Task' components to the
 
 Adding formatting (for example, bold text) in the task name is not currently supported in Umbraco.
 
-The status can be set in the settings of each 'Task', but usually you would [override the status property](/docs/umbraco/override-property-values.md) in code. Setting a status other than those in the `TaskListTaskStatus` enum is not currently supported.
+The status text content can be set in the content of each 'Task' using the 'Status text' property. If this value is not set then the dictionary value assigned to the `TaskListStatusEnum` will be used. If no dictionary value is available then the text of the enum will be used. Usually you would override [this content property](/docs/umbraco/override-property-values.md) in code.
```
sussexrick: You're talking about the dictionary value before the dictionary value is explained. Also, because the need to use 'status text' should be the exception rather than the rule, it makes sense to explain it last. 

You've already got the explanation of the default text and dictionary below. I would remove the paragraph you have, and add a simpler paragraph lower down after the bit about the dictionary. Something like "Use the 'Status text' property to override the status text for an individual task when using the dictionary entry is not flexible enough for your application."

sussexrick (changes requested):

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govuktasksettings.config
sussexrick, line 1: You need to copy this update to the GovUK.Frontend.Umbraco project, otherwise it won't be applied to consuming projects and will get overwritten here as soon as you build.

GovUk.Frontend.Umbraco.Testing/UmbracoDictionaryFactory.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Moq;
+using Umbraco.Cms.Core.Models;
+using Umbraco.Cms.Core.Services;
+
+namespace GovUk.Frontend.Umbraco.Testing
+{
+    public static class UmbracoDictionaryFactory
```
sussexrick: To be consistent with [similar existing functionality](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/GovUk.Frontend.Umbraco.Testing/UmbracoPublishedElementExtensions.cs) I would expect:

- the class to be called `LocalizationServiceExtensions`
- the method to be called `SetupDictionaryItem` or `SetupUmbracoDictionaryItem` ([Item is Umbraco terminology](https://docs.umbraco.com/umbraco-cms/fundamentals/data/dictionary-items/))
- the method should return the mock, to allow a fluent interface
- the method should have two overloads to accept either the mock or the mocked object

sussexrick (changes requested):

GovUk.Frontend.Umbraco.ExampleApp/Controllers/TaskListController.cs
sussexrick: Not done here?

GovUk.Frontend.Umbraco.Testing/UmbracoDictionaryFactory.cs
sussexrick: And the documentation...

sussexrick (dismissed):

sussexrick (approved):

PR #163, opened 2023-06-27
Update to release version of govuk-frontend-aspnetcore

sussexrick:


samuelhaa:
I can see that the version number of govuk-frontend-aspnetcore-extensions was increased to 3.1.2 last week but that was because of a bugfix.  Do we need to bump the version number for this as well?

sussexrick:
Not yet. There are other changes to come.

samuelhaa (approved):

PR #165, opened 2023-06-27
When using TPR styles, links within button groups should be bold 

sussexrick:
When using TPR styles, all links should be semi-bold, but GOV.UK styles were resetting these to `normal`.

[AB#151933](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/151933)


AlexBeesley:
Looks good, to use this will `_tpr-button-group.scss` need to be added in the backoffice custom styles section for the button group component, or will these styles be picked up automatically?

AlexBeesley (approved):

sussexrick:
@AlexBeesley They'll be picked up automatically.

PR #167, opened 2023-06-29
Add companies house validation attribute

samuelhaa:
A valid companies house should consist of

- 8 numbers or
- 2 letters and 6 numbers or
- 1 letter and 7 numbers

This validates a string against those criterion

[AB#155184](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/155184)


sussexrick (changes requested):
1. You should remove `Is` from the name to be consistent with `UkPostcodeAttribute`
2. Looking online the correct name is `Company Registration Number` but it's commonly referred to as `Company Number`, so I would suggest `CompaniesHouseCompanyNumberAttribute` for the name. Companies House may have other types of number it assigns.
3. Also looking online, I see no evidence for a "one character seven numbers" format, so I wonder whether that is actually a Companies House company number, or something else like a charity number? It would be worth double-checking that with our DBAs as I know we combine data sources.
4. You should add documentation to the [Text input](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/docs/components/text-input.md) page, similar to that for postcodes (and please fix the incorrect h1 while you're there)
5. Optional: It would be good to add a working example in the example apps too (postcode is demonstrated on the 'Text input' page)

samuelhaa:
1 and 2. I've renamed it 
3. I've asked the question and waiting for a reply.
4. Added documentation, similar to UkPostcode. Changed the h1 to h2s
5. I get a 404 when trying to access the backoffice of the example app. I've tried the usual `https://localhost:44326/umbraco`

sussexrick:
3. OK, we'll wait for the response.
4. Sorry, I wasn't clear. The heading levels were correct so you should put them back. I was referring to the h1 saying 'Date input' even though it's about the 'Text input' component
5. There are two example apps. 44326 is the one that doesn't use Umbraco. The Umbraco one is on port 44350.
6. 'Enter a valid company number' should be 'Enter a company number in the correct format, like 12345678 or AA123456'. Obviously that changes if the third format is also valid.

samuelhaa:
4. I changed it to and h1 and renamed it Text input
5. I changed the default error message

sussexrick (approved):

docs/components/text-input.md
```diff
@@ -17,3 +17,22 @@ Error messages for text input components are configurable in Umbraco on the sett
...
 ![Settings for a postcode using a text input component](/docs/images/uk-postcode-settings.png)
+
+
+## Companies house company number
+
+### Validating company numbers
+
+Use the `[CompaniesHouseCompanyNumber]` attribute to validate company numbers using the Text input component. This uses a regular expression rather than a check against a database, so some false positives may be allowed.
+
+For an example of this in use, see the 'Text input' page in either the ASP.NET and Umbraco example apps.
```
sussexrick, line 28: You're stating there are examples in both example apps, but you haven't added those examples.

sussexrick (changes requested):

docs/components/text-input.md
samuelhaa: They are there now

sussexrick (approved):

PR #169, opened 2023-06-29
Umbraco block list helper methods refactor

sussexrick:
This contains breaking changes and sees major version numbers of packages increase. This is a good time to do it as the next version to be published will have v1.0.1 of `govuk-frontend-aspnetcore` which also has breaking changes from the release candidate we currently target.

- Moves Umbraco functionality for filtering and overriding block lists out of the `GovUk.Frontend.Umbraco` package and into a new package called `ThePensionsRegulator.Umbraco`. This is a new home for Umbraco helper functionality that is unrelated to the GOV.UK Design System, enabling it to be used on sites that are not using the GOV.UK Design System (in TPR's case, the intranet)
- Reduces duplication in extension methods:
   - only one implementation of recursing through blocks
   -  `FindBlock\FindBlocks` is to be found on collections of blocks, not repeated in other places like content nodes
   - content nodes get the previously-private ability to `FindBlockLists`, from where they can use all the variants of `FindBlock\FindBlocks` rather than just those that were previously re-implemented 
- Adds support for `IPublishedValueFallback` rather than just passing `null`. We're not currently using fallback, but it's an important part of Umbraco functionality [allowing properties to be inherited from parent content or other languages](https://our.umbraco.com/apidocs/v8/csharp/api/Umbraco.Web.Models.PublishedContent.PublishedValueFallback.html).
- Deletes `RemoveBlockListSettingsErrorsActionFilter` which appears to work around a bug that is no longer reproducible in Umbraco 10.4.2


ThePensionsRegulator.Umbraco.Tests/BlockListModelExtensionsTests.cs
```diff
@@ -0,0 +1,194 @@
...
+            // Assert
+            Assert.That(results.Count(), Is.EqualTo(2));
+            Assert.Contains(blockList.BlocksToMatch[0], results);
+            Assert.Contains(blockList.BlocksToMatch[1], results);
+        }
+
+        private static (BlockListModel BlockList, IList<BlockListItem> BlocksToMatch) CreateBlockListHierarchyWithMultipleMatchingBlocks()
+        {
+            var matchingBlockContent1 = new Mock<IOverridablePublishedElement>();
+            matchingBlockContent1.Setup(x => x.GetProperty("MyTextProperty")).Returns(UmbracoPropertyFactory.CreateTextboxProperty("MyTextProperty", "value"));
```
HarryAmmon: Can the literal `"MyTextProperty"` be passed made a constant that is referenced throughout these tests?

If someone changed "MyTextProperty" in one place and not others, these tests would fail giving the impression that "MyTextProperty" is important to the methods under test when what is actually important is that they are the same.

HarryAmmon (changes requested):

ThePensionsRegulator.Umbraco.Tests/BlockListModelExtensionsTests.cs
sussexrick: Done

HarryAmmon (dismissed):

HarryAmmon (approved):

PR #173, opened 2023-07-06
When you access a key not in session state it should return null

sussexrick:
In testing the new fake was throwing a `KeyNotFoundException` which was not faking correctly the behaviour of the real thing.


ThePensionsRegulator.Umbraco.Testing/FakeSessionState.cs
samuelhaa, line 1: Any reason why we can't use the [DistributedSession Class](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.session.distributedsession?view=aspnetcore-7.0) instead?

HarryAmmon (dismissed):

ThePensionsRegulator.Umbraco.Testing/FakeSessionState.cs
sussexrick: Good suggestion - done. 

@samuelhaa @samuelhaa @ancientyouthman Ready for review again.

samuelhaa (approved):

PR #176, opened 2023-07-17
It should be possible to configure the header and context bars to not allow HTML

sussexrick:
It should be possible to configure the header and context bars to not allow HTML when they're known not to require it.

[AB#150712](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/150712)

### DIFF
--- a/GovUk.Frontend.Umbraco/Models/TprHeaderLockupModel.cs
+++ b/GovUk.Frontend.Umbraco/Models/TprHeaderLockupModel.cs
@@ -16,6 +16,10 @@ namespace GovUk.Frontend.Umbraco.Models
         public bool ShowSkipLink { get; init; } = true;
         public bool ShowPhaseBanner { get; init; } = true;
         public bool ShowHeader { get; init; } = true;
+        public bool HeaderBarContentAllowHtml { get; init; } = true;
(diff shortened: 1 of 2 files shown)


GovUk.Frontend.Umbraco/Models/TprHeaderLockupModel.cs
```diff
@@ -16,6 +16,10 @@ public TprHeaderLockupModel(IPublishedContent settings)
         public bool ShowSkipLink { get; init; } = true;
         public bool ShowPhaseBanner { get; init; } = true;
         public bool ShowHeader { get; init; } = true;
+        public bool HeaderBarContentAllowHtml { get; init; } = true;
```
samuelhaa, line 19: The `init` keyword means that a value can only be assigned during construction. That makes it immutable after construction. How does that make it configurable? It can't be changed.

sussexrick: They can be changed in an object initializer, which is how the other similar properties are already being used.

`new TprHeaderLockupModel(settings) { HeaderBarContentAllowHtml = false };`

samuelhaa (approved):

PR #181, opened 2023-07-24
Non-breaking changes to reduce boiler-plate code

sussexrick:
## Adds a `PropertyValueConverter` for `OverridableBlockListModel` 

This means that the models generated by Umbraco have `OverridableBlockListModel` properties where they used to have `BlockListModel` properties. 

You no longer need to create a separate property on your view model to use `OverridableBlockListModel`. You can now override property values directly on the generated model. In read-only use cases you may no longer need a separate view model class at all.

```csharp
// Before
viewModel.Blocks = new OverridableBlockListModel(viewModel.Page.Blocks!);
var block = viewModel.Blocks.FindBlockByContentTypeAlias("myAlias");

// After
var block = viewModel.Page.Blocks.FindBlockByContentTypeAlias("myAlias");
```

## Makes the setter for `OverridableBlockListModel.Filter` public

Previously you could only set the filter via the constructor, but since the model may now be constructed by Umbraco you can now set it by the property as well.

```csharp
// Before
viewModel.Blocks = new OverridableBlockListModel(viewModel.Page.Blocks, model =>
    model.Where(block =>
        block.Content.ContentType.Alias != GovukRadio.ModelTypeAlias ||
        ((GovukRadio)block.Content).Value != "not-relevant")
    );

// After
viewModel.Page.Blocks!.Filter = model =>
    model.Where(block =>
        block.Content.ContentType.Alias != GovukRadio.ModelTypeAlias ||
        ((GovukRadio)block.Content).Value != "not-relevant");
```

## Adds a `TypeConverter` to `OverridableBlockListModel`

This means that both of the following now work where previously you could only use the first one:

```csharp
var model1 = CurrentPage.Value<BlockListModel>("blocks");
var model2 = CurrentPage.Value<OverridableBlockListModel>("blocks");
```

## Adds an indexer to `OverridableBlockListModel`

This allows you to use `model.Blocks[0]` or iterate over `model.Blocks` with a `for` loop.

## Adds extension method overloads with `OverridableBlockListItem`

In many cases this removes the need to cast the result of methods. For example: 

```csharp
// Before
((OverridableBlockListItem)model.Blocks.FindBlockByContentTypeAlias("myAlias")).Settings.OverrideValue("myProperty", "some value")

 // After
model.Blocks.FindBlockByContentTypeAlias("myAlias").Settings.OverrideValue("myProperty", "some value")
```

[AB#146149](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/146149)


ThePensionsRegulator.Umbraco/BlockLists/OverridableBlockListComposer.cs
HarryAmmon, line 1: I can't see this being used in the Example app, does it need to called directly? I've read [in the docs](https://docs.umbraco.com/umbraco-cms/implementation/composing#example-creating-a-composer-to-listen-for-contentsavingnotification) that it needs to implement `IComposer`, which this does.

I assume when `.AddComposers()` is called anything that implements `IComposer` has its `Compose` method called but how does it know what order to compose it in? Is there anything preventing the default Umbraco composer that adds `BlockListPropertyValueConverter` running after this composer?

HarryAmmon (approved):

ThePensionsRegulator.Umbraco/BlockLists/OverridableBlockListComposer.cs
sussexrick: Not sure exactly how the Umbraco internals work, but I am following the docs for replacing a built-in PropertyValueConverter: https://docs.umbraco.com/umbraco-cms/extending/property-editors/property-value-converters

PR #182, opened 2023-07-31
Support numbered list style types in Umbraco 

sussexrick:
In the Umbraco rich text editor, if you enable the use of ordered lists you enable the selection of different styles of numbering:

image

These numbering styles were being stripped by TinyMCE configuration that disallows the `style` attribute. To enable them, allow the `style` attribute in TinyMCE configuration and strip it out when `GovUkTypography` applies GOV.UK classes instead. Before stripping it, apply classes if one of the style attribute values we're looking for exists.

[AB#165186](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/165186)

### DIFF
--- a/GovUk.Frontend.Umbraco.Tests/GovUkTypographyTests.cs
+++ b/GovUk.Frontend.Umbraco.Tests/GovUkTypographyTests.cs
@@ -130,6 +130,37 @@ namespace GovUk.Frontend.Umbraco.Tests
         }
 
         [Test]
+        public void Style_attribute_is_removed_from_ordered_lists()
+        {
+            var html = "<ol style=\"list-style-type: lower-alpha;\"><li>Item 1</li><li>Item 2</li></ol><ol style=\"color: red;\"><li>Item 3</li><li>Item 4</li></ol>";
+
+            var result = GovUkTypography.Apply(html);
+
+            var doc = new HtmlDocument();
+            doc.LoadHtml(result);
+            Assert.AreEqual(2, doc.DocumentNode.SelectNodes("//ol").Count);
+            Assert.Null(doc.DocumentNode.SelectNodes("//ol[@style]"));
+        }
+
+        [TestCase("lower-alpha")]
+        [TestCase("lower-greek")]
+        [TestCase("lower-roman")]
+        [TestCase("upper-alpha")]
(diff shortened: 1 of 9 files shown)


GovUk.Frontend.Umbraco.Tests/GovUkTypographyTests.cs
```diff
@@ -129,6 +129,37 @@ public void List_classes_are_applied_to_ordered_lists()
             Assert.AreEqual(2, doc.DocumentNode.SelectNodes("//ol[contains(@class,'govuk-list--number')]").Count);
         }
 
+        [Test]
+        public void Style_attribute_is_removed_from_ordered_lists()
```
samuelhaa, line 133: I first thought that the purpose of this test was to test that any non-permitted styles were removed so I got a bit confused.  Would it be possible to change the title to "**All_style attributes_are**_removed_from_ordered_list"? (If I haven't misunderstood the purpose of the test)

samuelhaa:
This doesn't seem to work for me. I can do the default numbering
image

But the numbering is not displayed when I set it to Lower alpha. Am I missing anything?

image
image

samuelhaa (changes requested):

GovUk.Frontend.Umbraco.Tests/GovUkTypographyTests.cs
sussexrick: The _attribute_ is the HTML, ie `style=""`. The _value_ of the style attribute is a semi-colon separated list of CSS _property: value_ pairs.

It is correct to say that the style attribute (singular) is being removed. A test for what you're talking about would be called `All_CSS_properties_are_removed_from_style_attribute_on_ordered_list", but that might imply that the empty attribute itself remains.

sussexrick:
> This doesn't seem to work for me. I can do the default numbering ... But the numbering is not displayed when I set it to Lower alpha. Am I missing anything?

You probably have an old cached stylesheet. Try Ctrl+F5. There's no cache-busting on the example app, but there is on the apps we deploy.

samuelhaa (approved):

PR #185, opened 2023-08-01
Improve support for finding blocks by class and updating classes in controllers

sussexrick:
**This is a follow-up to #181, so that should be merged before reviewing this.**

For solutions using `ThePensionsRegulator.GovUk.Umbraco` this PR adds the ability to do things like:

```csharp
var block = blockList.FindBlockByClass("my-class-name");
if (block.ClassList().Contains("change-me")) 
{
    block.ClassList().Add("my-new-class");
}

var blocks = blockList.FindBlocksByClass("my-class-name");
foreach (var block in blocks)
{
    block.ClassList().Remove("my-old-class");
}
```

There are also versions to find blocks by classes in the 'CSS classes for row' and 'CSS classes for column' fields.

Full functionality is only available for an `OverridableBlockListItem`, but read-only methods such as `block.ClassList().Contains("my-class-name")` will also work for `BlockListItem`.

Help text is also added in the back office to explain where classes are applied to.

### DIFF
--- a/GovUk.Frontend.Umbraco.ExampleApp/Controllers/SummaryCardController.cs
+++ b/GovUk.Frontend.Umbraco.ExampleApp/Controllers/SummaryCardController.cs
@@ -9,25 +9,21 @@ using Umbraco.Cms.Web.Common.PublishedModels;
 
 namespace GovUk.Frontend.Umbraco.ExampleApp.Controllers
 {
-	public class SummaryCardController : RenderController
-	{
-		public SummaryCardController(ILogger<RenderController> logger, ICompositeViewEngine compositeViewEngine, IUmbracoContextAccessor umbracoContextAccessor) : base(logger, compositeViewEngine, umbracoContextAccessor)
-		{
-		}
+    public class SummaryCardController : RenderController
+    {
+        public SummaryCardController(ILogger<RenderController> logger, ICompositeViewEngine compositeViewEngine, IUmbracoContextAccessor umbracoContextAccessor) : base(logger, compositeViewEngine, umbracoContextAccessor)
+        {
+        }
 
-		[ModelType(typeof(SummaryCard))]
-		public override IActionResult Index()
-		{
-			var viewModel = new SummaryCard(CurrentPage, null);
+        [ModelType(typeof(SummaryCard))]
+        public override IActionResult Index()
+        {
+            var viewModel = new SummaryCard(CurrentPage, null);
 
-			// Override content in the block list
-			var target = viewModel.Blocks!.FindBlockByClass("full-name");
-			if (target != null)
-			{
-				target.Content.OverrideValue(nameof(GovukSummaryListItem.ItemValue), "Sarah Smith");
-			}
+            // Override content in the block list
+            viewModel.Blocks!.FindBlockByClass("full-name")?.Content.OverrideValue(nameof(GovukSummaryListItem.ItemValue), "Sarah Smith");
 
-			return CurrentTemplate(viewModel);
-		}
-	}
+            return CurrentTemplate(viewModel);
+        }
+    }
 }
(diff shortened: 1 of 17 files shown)


ThePensionsRegulator.Umbraco/TokenList.cs
```diff
@@ -0,0 +1,237 @@
...
+		private void Tokenise()
+		{
+			var raw = _isReadOnly ? _publishedElement?.Value<string>(_propertyName) : _overridablePublishedElement!.Value<string>(_propertyName);
+			_tokens.Clear();
+			if (!string.IsNullOrEmpty(raw))
+			{
+				raw = _regex.Replace(raw, _separator).Trim();
+				_tokens.AddRange(raw.Split(_separator));
+			}
+			else
```
HarryAmmon: Is this `else` really needed? The list will already be empty if this `else` statement is reached

ThePensionsRegulator.Umbraco/TokenList.cs
```diff
@@ -0,0 +1,237 @@
...
+			{
+				_tokens.Clear();
+			}
+			_tokensCurrent = true;
+		}
+
+		/// <summary>
```
HarryAmmon, line 55: This documentation describes getting a token index, not getting a token by index

ThePensionsRegulator.Umbraco.Tests/TokenListTests.cs
```diff
@@ -0,0 +1,196 @@
...
+			var result = new string[tokenList.TokenList.Count];
+			tokenList.TokenList.CopyTo(result, 0);
+
+			Assert.That(result[0], Is.EqualTo("example-a"));
+			Assert.That(result[1], Is.EqualTo("example-b"));
+			Assert.That(result[2], Is.EqualTo("example-c"));
+		}
+
+		[Test]
+		public void Can_find_by_index()
```
HarryAmmon: This tests that an index is returned for a given class, not that the correct class is found by it's index.

GovUk.Frontend.Umbraco.ExampleApp/Controllers/SummaryCardController.cs
```diff
@@ -1,34 +1,33 @@
...
 using Microsoft.Extensions.Logging;
-using ThePensionsRegulator.Umbraco.BlockLists;
 using Umbraco.Cms.Core.Web;
 using Umbraco.Cms.Web.Common.Controllers;
 using Umbraco.Cms.Web.Common.PublishedModels;
 
 namespace GovUk.Frontend.Umbraco.ExampleApp.Controllers
 {
-    public class SummaryCardController : RenderController
-    {
```
HarryAmmon: Changed from spaces to tabs, is this intentional? Doesn't matter what one we use, as long as it's consistent with the rest of the project

HarryAmmon (changes requested):

ThePensionsRegulator.Umbraco.Tests/TokenListTests.cs
sussexrick: Well spotted. Fixed.

GovUk.Frontend.Umbraco.ExampleApp/Controllers/SummaryCardController.cs
sussexrick: I think it came from the github conflicts editor. I've fixed it to spaces.

ThePensionsRegulator.Umbraco/TokenList.cs
sussexrick: Well spotted. Fixed.

ThePensionsRegulator.Umbraco/TokenList.cs
sussexrick: Well spotted. Fixed.

HarryAmmon (approved):

PR #187, opened 2023-08-18
Apply automatic host updating to links everywhere

sussexrick:
[AB#164201](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/164201) notes that `TprHostUpdater` is used to update links in 'Text' blocks, but misses links in other rich text blocks such as hint text. It also misses links in a 'MultiUrlPicker' property editor.

Rather than trying to catch them all individually this refactors the code, in a non-breaking change, to apply that update to all rich text and multi-URL picker values as they're retrieved from Umbraco.

## Changes
- Introduces and documents a new concept of an `IPropertyValueFormatter`. 
- Moves the application of `GovUkTypography` formatting and running `TprHostUpdater` from the moment of output into new property value formatters that run earlier, automatically. The folder/namespace structure follows that of the related concept of `IPropertyValueConverter` in Umbraco itself.
- Because different properties need different formatting options for `GovUkTypography`, creates several new property editor types as hooks for the different sets of requirements. Each property editor is targeted by a different `IPropertyValueFormatter`. Creates/updates data types to use these new property editors. Updates element types for various blocks to use the new data types.
- Creates `RichTextEditorPropertyValueConverter` (and for GOV.UK projects, `GovUkRichTextEditorPropertyValueConverter`) and updates `OverridablePublishedElement` to run registered `IPropertyValueFormatter` instances on compatible properties. This ensures that the two processes are run any time a rich HTML field is accessed, or any time it is overridden using `OverridablePublishedElement.OverrideValue`. It can't be forgotten.
- Creates `MultiUrlPickerPropertyValueConverter` to do the same for multi-URL picker fields.
- Adds shortcut methods for mocking rich text properties in unit tests


GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govuksummarylistitem.config
```diff
@@ -58,7 +58,7 @@
       <Name>Value</Name>
       <Alias>itemValue</Alias>
       <Definition>0729c57a-748f-482d-abbc-fcd2976c9fc6</Definition>
```
HarryAmmon, line 60: Does it matter that the definition hasn't changed for this and govuktask, tprcontextbar1 - 3, tprfooter and tprheader but has for the other content types above?

sussexrick: Well spotted, but no it doesn't because they already had their own data types.

HarryAmmon (approved):

PR #188, opened 2023-08-25
Make it easy to display data from databases and other data sources

sussexrick:
Where we have needed to populate an unknown number of items into a block (eg a select list of countries, or search results displayed as radio buttons) we have so far used a "blocks before" and "blocks after" pattern. With this pattern we drop out of Umbraco and back to plain ASP.NET to accomplish the task, but then we have to wire up things that should work as standard (like error messages) to custom properties in Umbraco so that they remain editable.

The PR adds extension methods and documentation for the list components - radios, checkboxes, select and summary list - which make it easy to override a list of child blocks at runtime. For example, to populate a select component with a list of countries from a database:

```csharp
var block = viewModel.Page.Blocks.FindBlockByContentTypeAlias(GovukSelect.ModelTypeAlias);
block.Content.OverrideSelectOptions(
     myDatabaseCountries.Select(c => new SelectOption(c.CountryId, c.CountryName)), 
     _publishedSnapshotAccessor
);
```

This required the use of Umbraco's `PublishedSnapshotAccessor` so extra properties for working with that are added to `UmbracoTestContext`. This will therefore lead to two new versions:

* ThePensionsRegulator.GovUk.Frontend.Umbraco v4.2.0
* ThePensionsRegulator.Umbraco.Testing v4.2.0

This work also led to repeating a lot of element type aliases and property aliases, so I added them as constants and replaced existing magic strings.

### DIFF
--- a/GovUk.Frontend.Umbraco/Views/Partials/GOVUK/GovUkSelect.cshtml
+++ b/GovUk.Frontend.Umbraco/Views/Partials/GOVUK/GovUkSelect.cshtml
@@ -11,7 +11,10 @@
 @using Umbraco.Cms.Core.Strings;
 @using Umbraco.Extensions
 @{
-    var options = Model.Content.Value<OverridableBlockListModel>("options")!;
+    var options = Model.Content.Value<OverridableBlockListModel>(PropertyAliases.SelectOptions)!;
+
+    var wtf = options.FilteredBlocks();
+
     var modelPropertyName = Model.Settings.Value<string>(PropertyAliases.ModelProperty);
     ModelStateEntry? modelStateEntry = null;
     if (!string.IsNullOrEmpty(modelPropertyName))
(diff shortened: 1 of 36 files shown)


GovUk.Frontend.Umbraco/Views/Partials/GOVUK/GovUkSelect.cshtml
```diff
@@ -11,7 +11,10 @@
 @using Umbraco.Cms.Core.Strings;
 @using Umbraco.Extensions
 @{
-    var options = Model.Content.Value<OverridableBlockListModel>("options")!;
+    var options = Model.Content.Value<OverridableBlockListModel>(PropertyAliases.SelectOptions)!;
+
+    var wtf = options.FilteredBlocks();
```
HarryAmmon: Not 100% sure this is used and if so, it could have a more descriptive name

sussexrick: Oops, that was just debugging code. Thank you for spotting that.

HarryAmmon (approved):

PR #191, opened 2023-08-29
Separate TPR components and branding from GOV.UK for ASP.NET without Umbraco

sussexrick:
This is step 1 of moving TPR-specific components and branding into separate NuGet packages, so that GOV.UK ones do not come with too much TPR baggage. 'GOV.UK frontend extensions' has strayed beyond its original purpose to become a TPR design system, and this brings it back to that purpose again. It also prepares us for an expansion of TPR-specific components and styling as our use of the library expands with upcoming projects.

Utility functions that aren't specific to TPR but are just useful for building applications, eg `INonceProvider`, remain in the GOV.UK project for anyone to take advantage of.

The package versions are shown as beta at the moment. This will be cleaned up in a later PR which tackles a more thorough update to the documentation than the minimal changes shown here.


ThePensionsRegulator.Frontend/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+        public static IServiceCollection AddTprFrontendExtensions(
+            this IServiceCollection services,
+            Action<GovUkFrontendAspNetCoreOptions> options)
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            services.AddGovUkFrontendExtensions(options);
```
HarryAmmon, line 24:
```suggestion
            services.AddTprFrontendExtensions(options);
```

Should this not be `AddTprFrontend`?

HarryAmmon: No it should be how it is, misunderstood so marking as resolved

HarryAmmon (dismissed):

HarryAmmon (approved):

PR #202, opened 2023-09-01
The divider text for radios and checkboxes should not wrap at 40px

sussexrick:
You can have divider text between radio buttons and checkboxes. The text is usually "or" which is fine, but in Umbraco we provide the ability to customise it. If you do, it'll probably wrap badly, like this:

Divider text wrapping at 40px

This PR allows the text to be longer:

Divider text not wrapping


sussexrick:
By the way the snyk failure needs to be addressed but it's nothing to do with the changes here, so I'll make that a separate PR.

HarryAmmon (dismissed):

bentpr (approved):

PR #208, opened 2023-09-11
Changes to the govuk-validation script that allow the use of custom prefixes

KyleAlbiston-TPR:
The update title function now uses a data-govuk-error-prefix attribute which can be used to set custom prefixes for the page title when an error occurs. 
A default prefix of "Error: " is used when the data-govuk-error-prefix attribute is not implemented.


GovUk.Frontend.AspNetCore.Extensions/wwwroot/govuk/govuk-validation.js
```diff
@@ -32,8 +32,15 @@ function createGovUkValidator() {
     /**
      * Ensures the page title is prefixed with 'Error: ' when there is at least one .govuk-error-message displayed
      */
-    updateTitle: function () {
-      const prefix = "Error: ";
+      updateTitle: function () {
+        const getPrefix = document.querySelector(".govuk-error-prefix");
```
sussexrick: You can simplify this with `document.querySelector("[data-govuk-error-prefix]")`. This will find the first element with that attribute, so you don't need to put a class on the element in order to find it.

GovUk.Frontend.AspNetCore.Extensions/wwwroot/govuk/govuk-validation.js
```diff
@@ -32,8 +32,15 @@ function createGovUkValidator() {
...
+        let prefix = ""
+          if (getPrefix == null) {
+              prefix = "Error: "
+          }
+          else {
+              prefix = getPrefix.getAttribute("data-govuk-error-prefix", "&nbsp;");
```
sussexrick: There should be no need for the second parameter on `getAttribute`.

GovUk.Frontend.ExampleApp/Views/TextInput/Index.cshtml
```diff
@@ -12,6 +13,7 @@
 
 <form asp-controller="TextInput" asp-action="Post" method="post" novalidate>
     <span class="govuk-caption-xl">Example application</span>
+        <div class="govuk-error-prefix" data-govuk-error-prefix=@Localizer["Error:"]></div>
```
sussexrick: You don't need to insert a new element for this data. You could put the attribute on any element. In this view the `form` element would be a good place to put it.

Also `Localizer` is a variable so should by convention be lowercase.

GovUk.Frontend.AspNetCore.Extensions/wwwroot/govuk/govuk-validation.js
```diff
@@ -32,8 +32,15 @@ function createGovUkValidator() {
     /**
      * Ensures the page title is prefixed with 'Error: ' when there is at least one .govuk-error-message displayed
      */
-    updateTitle: function () {
-      const prefix = "Error: ";
+      updateTitle: function () {
```
sussexrick, line 35: You need to add/update tests for this function: https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/GovUk.Frontend.AspNetCore.Extensions/__tests__/govuk-validation-updateTitle.tests.js

sussexrick (changes requested):

GovUk.Frontend.AspNetCore.Extensions/wwwroot/govuk/govuk-validation.js
```diff
@@ -30,10 +30,17 @@ function createGovUkValidator() {
...
 
     /**
-     * Ensures the page title is prefixed with 'Error: ' when there is at least one .govuk-error-message displayed
+     * Ensures the page title is prefixed with the data-govuk-error-prefix attribute, or 'Error: ' when there is at least one .govuk-error-message displayed
      */
-    updateTitle: function () {
-      const prefix = "Error: ";
+      updateTitle: function () {
+        const getPrefix = document.querySelector("[data-govuk-error-prefix]")
+        let prefix = ""
```
HarryAmmon: There's inconsistent spacing here. Also make sure you end lines with `;` in JS. If you don't the interpreter guesses where the line of code ends which can lead to some odd bugs

HarryAmmon (dismissed):

sussexrick (approved):

sussexrick (changes requested):
This fixes the issue for client-side validation, but if there is a server-side error (either because the validation is not supported on the client or JavaScript is disabled) then it will fall back to the hard-coded English text implemented by [TitleTagHelper](https://github.com/gunndabad/govuk-frontend-aspnetcore/blob/1f1363e9cb4df87840008e2c7294c000c1cac782/src/GovUk.Frontend.AspNetCore/TagHelpers/TitleTagHelper.cs)

I've [logged this as an issue](https://github.com/gunndabad/govuk-frontend-aspnetcore/issues/270) with the base library. In the meantime I would suggest we have a copy of `TitleTagHelper` in our own library that does support localisation through the use of an attribute, eg `<title error-prefix="Error: ">My page title</title>`. A consuming application would feed the localised value to the `error-prefix` attribute, or it would fall back to "Error: " if the attribute were not set.

It would make sense if this not only updated the server-side rendered page title, but also rendered the `data-govuk-error-prefix` attribute for the client-side code to pick up. In other words the title tag would look as follows:

In Razor:
 `<title error-prefix="Error: ">My page title</title>`

In HTML, with no server-side error:
 `<title data-govuk-error-prefix="Error: ">My page title</title>`

In HTML, with a server-side error:
 `<title data-govuk-error-prefix="Error: ">Error: My page title</title>`

sussexrick:
@KyleAlbiston-TPR The base implementation now includes the ability to customise the error prefix on the server-side. That isn't available for us to import yet, but you can start by copying the implementation: https://github.com/gunndabad/govuk-frontend-aspnetcore/pull/271/files

GovUk.Frontend.AspNetCore.Extensions/TagHelpers/TitleTagHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+        [HtmlAttributeNotBound]
+        [ViewContext]
+        [DisallowNull]
+        public ViewContext? ViewContext { get; set; }
+        /// <inheritdoc/>
+        public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
+        {
+            await base.ProcessAsync(context, output);
+
+            if (_options.PrependErrorToTitle && !string.IsNullOrEmpty(ErrorPrefix) && !ViewContext!.ModelState.IsValid)
```
sussexrick: You could modify this slightly to render the attribute for the client-side code, then you could avoid the duplication that's happening in `Layout.cshtml` and just add `gfa-error-prefix` instead of both that and `data-govuk-error-prefix`.

```csharp
if (_options.PrependErrorToTitle && !string.IsNullOrEmpty(ErrorPrefix))
{
    if (!ViewContext!.ModelState.IsValid)
    {
        output.PreContent.Clear();
        output.PreContent.Append(ErrorPrefix + " ");
    }
    output.Attributes.Add("data-govuk-error-prefix", ErrorPrefix);
}
```

GovUk.Frontend.AspNetCore.Extensions/__tests__/govuk-validation-global.tests.js
```diff
@@ -10,6 +10,7 @@ describe("govuk-validation with only jQuery Validator mocked", () => {
   it("should add an error to both the error message and summary", () => {
     document.body.innerHTML = `
         <main>
+    <title gfa-error-prefix="Error: " data-govuk-error-prefix="Error: ">@ViewData["Title"]</title>
```
sussexrick: You don't need `gfa-error-prefix` in these examples because it's a server-side attribute. If you view the source of the rendered page in your browser you'll see it's gone by the time JavaScript gets to look at it. 

Also a `title` element would never occur within `main`, so this should look more like:

```html
<head>
    <title></title>
</head>
<body>
    <main>
        ...
    </main>
</body>
```

sussexrick (changes requested):

GovUk.Frontend.AspNetCore.Extensions/__tests__/govuk-validation-global.tests.js
```diff
@@ -8,7 +8,10 @@ const _mockValidator = {
 
 describe("govuk-validation with only jQuery Validator mocked", () => {
   it("should add an error to both the error message and summary", () => {
-    document.body.innerHTML = `
+      document.body.innerHTML = `
+        <head>
```
sussexrick, line 12: You should make this change to the example below as well.

GovUk.Frontend.AspNetCore.Extensions/wwwroot/govuk/govuk-validation.js
```diff
@@ -30,10 +30,19 @@ function createGovUkValidator() {
...
 
     /**
-     * Ensures the page title is prefixed with 'Error: ' when there is at least one .govuk-error-message displayed
+     * Ensures the page title is prefixed with the data-govuk-error-prefix attribute, or 'Error: ' when there is at least one .govuk-error-message displayed
      */
-    updateTitle: function () {
-      const prefix = "Error: ";
+      updateTitle: function () {
+          const titleTag = document.getElementsByTagName("title")[0];
+          const getPrefix = titleTag.getAttribute("gfa-error-prefix");
```
sussexrick: You don't need to check for `gfa-error-prefix` as it's a server-side attribute that will never be there when JavaScript sees it.

sussexrick (changes requested):

sussexrick (dismissed):
Approving, but you could tidy up the test cases in `govuk-validation-updateTitle.tests.js`. The `gfa-error-prefix` attribute is in two of them, but would never really be there because it's server-side only.

sussexrick (approved):

PR #209, opened 2023-09-18
Support block grid as well as block list

sussexrick:
So far this project has only worked with the block list editor in Umbraco. This PR adds support for the block grid editor, opening up more layout possibilities for future projects.

It is at parity with block list support. Filtering and overriding blocks is fully supported.

- Add a new demo page for the block grid in the example app, and block grids on some other pages (these will be further revised as documentation is updated in following PRs). Renames the existing `BlockList` example page to `FilterAndOverrideBlocks` which applies to both block editors.
- Filters now use an interface that is common to block list items and block grid items. Views for each component also use the interface.
- Some namespaces and classes change from `BlockList` to `Blocks`
- Backoffice views are configured so that the backoffice interface is somewhat WYSIWIG. More can be done on this, but a good level of support is there for v1. `govuk-umbraco-rich-text-editor.css` becomes `govuk-umbraco-backoffice.css` and now supports this additional functionality. This shows part of the _backoffice_ view of the example page:
Backoffice view of block grid

- There is a lot more unit testing of `FindBlock` and `FindBlocks` extension methods. However the number of overloads is reduced to only support overridable blocks. These extension methods now run on interfaces that are common to both block list and block grid, but supporting both the base types and the overridable types meant having to use a long syntax to specify the overridable type with every call to avoid ambiguity. With ModelsBuilder now building overridable types by default it's very rare to encounter the base types (I couldn't find an example in our current applications) so I removed the base type support. 
- Task list summary logic moves out of the `BlockList.cshtml` view and into an `ActionFilter` that can handle block lists or block grids
- `RemoveSettingsErrorsActionFilter` deals with spurious validation errors that are thrown up by the block grid. This used to happen for the block list but got fixed, so I have revived the old code.
- There is a new 'Image' block in the TPR-specific project which is copied from the Umbraco example code for block grid, and allows adding images to the page.
- Block grid support is added to the testing project.


sussexrick:
Closing PR to create smaller PRs instead with the same content.

PR #214, opened 2023-09-28
Add areas and GOV.UK grid support to the block grid editor

sussexrick:
**This PR is a follow-up-to #213. It should be reviewed after that PR completes.** 

- Filters now use an interface that is common to block list items and block grid items. Views for each component also use the interface
- Support overriding, filtering and validating blocks within areas, which enables multi-column layouts using the block grid (an initial two-column layout is included, but others can be added as requirements emerge)
- Supports the GOV.UK grid using the Umbraco grid editor, by copying the approach used by the block list to create the grid HTML


sussexrick:
Closed due to the target branch being wrong.

PR #217, opened 2023-09-29
Change BlockLists namespaces to Blocks, ready for block grid support

sussexrick:


ThePensionsRegulator.Umbraco/Blocks/PublishedContentExtensions.cs
ancientyouthman, line 1: Are the arguments the right way round for the ArgumentException  ?

ancientyouthman: You may want:
`throw new ArgumentException($"{nameof(content)}.{nameof(content.Properties)} cannot be null", nameof(content));`

sussexrick: Well spotted, no they weren't.

ancientyouthman (approved):

PR #219, opened 2023-10-02
Update block views to use interface compatible with block list and block grid

sussexrick:
On the task list view it also applies filtering to the tasks, which previously were ignoring the filter.

### DIFF
--- a/GovUk.Frontend.Umbraco/Services/GovUkFieldsetErrorFinder.cs
+++ b/GovUk.Frontend.Umbraco/Services/GovUkFieldsetErrorFinder.cs
@@ -2,34 +2,34 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using ThePensionsRegulator.Umbraco;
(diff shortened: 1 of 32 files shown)


GovUk.Frontend.Umbraco/Services/GovUkFieldsetErrorFinder.cs
```diff
@@ -2,8 +2,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using ThePensionsRegulator.Umbraco.Blocks;
 using Umbraco.Cms.Core.Models.Blocks;
```
DeniseG123: Umbraco.Cms.Core.Models.Blocks isn't used - the using statement can be removed

sussexrick: Removed

DeniseG123 (approved):

PR #221, opened 2023-10-02
Create overridable block grid types and add support to testing project

sussexrick:
- Create overridable block grid classes following the same pattern as the existing classes for block lists. 
- Add tests, including new tests for the existing block list classes.
- Adds support for block grids to the Umbraco testing package, similar to that which already exists for block lists.


ThePensionsRegulator.Umbraco.Tests/Blocks/OverridableBlockGridModelTests.cs
```diff
@@ -0,0 +1,254 @@
...
+            // Act
+            var model = new OverridableBlockGridModel(blockGrids.ParentBlockGrid, null);
+            model.Filter = filter;
+
+            // Assert
+            Assert.That(model.Filter, Is.EqualTo(filter));
+
+            var childBlockGrid = model[0].Content.Value<OverridableBlockGridModel>(PROPERTY_ALIAS_CHILD_BLOCKS);
+            Assert.That(childBlockGrid!.Filter, Is.EqualTo(filter));
+        }
```
HarryAmmon, line 140: What about the grandchild?

HarryAmmon (changes requested):

ancientyouthman (dismissed):

ThePensionsRegulator.Umbraco.Tests/Blocks/OverridableBlockGridModelTests.cs
sussexrick: Extra test added for both block list and block grid types.

HarryAmmon (approved):

PR #231, opened 2023-10-16
Add block grid example to pagination page in example app

sussexrick:
Proves that pagination works with block grid, where it is necessary to override the total results property to get pagination to work.


samuelhaa:
I understand that the purpose of this is to demonstrate the pagination with a blockgrid. However, the difference between a blocklist and a blockgrid it's not visually obvious and that is confusing to me. 

Is it possible to add more blocks to the grid so that the difference between the blocklist and blockgrid is apparent? Just something visually demonstrate that this is a blockgrid and it is different to a blocklist?

For example:
image

sussexrick:
@samuelhaa Updated to this:

Preview of pagination in a block grid

samuelhaa (dismissed):

samuelhaa (approved):

PR #232, opened 2023-10-16
Add custom backoffice views for the block grid and block list editors

sussexrick:
- Backoffice views are configured for both block grid and block list so that the backoffice interface is somewhat WYSIWIG. More can be done on this, but a good level of support is there for v1. 
- Inline editing is disabled on some block list data types, because it's not compatible with using this type of preview.
- govuk-umbraco-rich-text-editor.css becomes govuk-umbraco-backoffice.css and now supports this additional functionality.
- Copies of App_Plugins files are excluded from the example app to avoid duplicates, since they are copied there on every build.

For example, the backoffice view of the 'Error message' example page now looks like this:

Error message page backoffice view

For comparison, the actual page looks like this:

Error message page


HarryAmmon (dismissed):
Happy to approve as it. Only feedback would be that in the back office I found a bit tricky to visually identify where one block ended and another began. I understand the benefit of WYSIWYG approach for previewing content. 

Is it possible to add some visual cue that a block begins/ ends? Perhaps a dotted line or re-use of the blue line that appears when you drag a block?

image

sussexrick:
@HarryAmmon Good suggestion, it now looks like this:

Backoffice with divider line between blocks

I've re-used the dotted line style that Umbraco puts around the 'Add content' button.

HarryAmmon (dismissed):
This is better though this change has removed the bottom border for the error summary component in the back office. If it can be added back that would be good but don't feel like it takes away from the WYSIWYG experience too much

image

sussexrick:
Fixed:

Error summary component with border restored

HarryAmmon (approved):

PR #235, opened 2023-10-26
Add tpr styling to govuk-panel (Success panel) AB#172468

samuelhaa:


samuelhaa:
@sussexrick Would this work for the Success panel?

sussexrick:
@samuelhaa You were on the right track. I've added another commit with some further fixes:

* we can use `govuk-panel` rather than `govuk-panel--confirmation` since the font size and spacing changes you were making would also apply to other variations
* you didn't need a special class for bold text, just use the bold button in the rich text editor
* if you leave out the description our bottom margin was overriding the default, creating excess padding between the heading and the bottom of the box
* if you add a second paragraph to the description it adds paragraph tags, which is correct, but the styling broke completely
* in the backoffice the preview of a link within a panel was not the correct colour, and there was no spacing above the heading

I've added extra examples to the Panel page in both example apps to highlight these configurations.

samuelhaa (commented):
Looks good

sussexrick (approved):

PR #237, opened 2023-10-30
Avoid using extension method to access properties

HarryAmmon:
When creating an `OverrideableBlockListModel` with a mocked `IPublishedElement`, `GetValue(alias)` would throw the exception `The type initializer for 'Umbraco.Extensions.FriendlyPublishedElementExtensions' threw an exception. ----> System.ArgumentNullException : Value cannot be null. (Parameter 'provider')` when converting a block model to an overrideable block model. 

To avoid this the constructor for `OverrideableBlockListModel` now gets the value of block being converted the 'long way' which avoids calling the mocked value.


sussexrick (approved):
Approving, but that is into ThePensionsRegulator.Umbraco v2.0.0 which is already published. 

If you want to wait for v2.0.0 to be applied to your application (which is in progress, but not done) then you can publish v2.0.1 and upgrade from there. 

If you want this in straight away you'll need to start from the v1.2.2 tag, make this fix again because `OverridableBlockModel` was just `OverridableBlockListModel` back then, and publish v1.2.3 that you could use right away.

PR #241, opened 2023-11-14
fix: Create property group in project targets to map $UserProfle to $User for linux

### LINKED ISSUE #240: Nuget package target paths fail on linux
I am adding the project to an umbraco project using:

```
dotnet add package ThePensionsRegulator.GovUk.Frontend.Umbraco
```

When I run the project using 
```
dotnet run
```

I get an error
```
Building...
/root/.nuget/packages/thepensionsregulator.govuk.frontend.umbraco/5.0.0/buildTransitive/ThePensionsRegulator.GovUk.Frontend.Umbraco.targets(54,3): error MSB3733: Input file "/.nuget/packages/thepensionsregulator.govuk.frontend.umbraco/5.0.0/thepensionsregulator.govuk.frontend.umbraco.nuspec" cannot be opened. Could not find a part of the path '/.nuget/packages/thepensionsregulator.govuk.frontend.umbraco/5.0.0/thepensionsregulator.govuk.frontend.umbraco.nuspec'. [/root/dev/mermaid-workflow/mermaid-cms/mermaid-cms.csproj]

The build failed. Fix the build errors and run again.
```

The reason for the is the use of `$(UserProfile)` in the ThePensionsRegulator.GovUk.Frontend.Umbraco.targets file. This is, as far as I am aware, only set on windows machines. Linux uses `$(HOME)`

To fix it you can add:

```
<PropertyGroup>
  <UserProfile Condition=" '$(OS)' == 'Windows_NT' ">$(UserProfile)</UserProfile>
  <UserProfile Condition=" '$(OS)' != 'Windows_NT' ">$(HOME)</UserProfile>
</PropertyGroup>
```

Near the top of the target file.

If anyone else gets this and needs a work round you can also add this in the PropertyGroup section of your csproj file

```
    <UserProfile Condition="'$(UserProfile)' == ''">$(HOME)</UserProfile>
```

PS - this is an awesome project. I will try and get round to contributing and providing a pull request for this, but I need a little time to get my head round it all. Also I did a quick search in the repo and the change may be required for the other nuget targets?

jonnymuir:
Fixes #240

I've created a property group to set UserProfile to PATH for if not on windows.

For the projects I've changed I've upped the version number to 5.0.1. I'm not sure if this was what you wanted or not, or whether that will generate new nuget packages in your CI/CD.

Hope this helps - as mentioned in the issue description, it isn't critical, as it is workable round-able, so feel free to ignore if you feel it too risky.

### DIFF
--- a/GovUk.Frontend.AspNetCore.Extensions/Directory.Build.targets
+++ b/GovUk.Frontend.AspNetCore.Extensions/Directory.Build.targets
@@ -6,6 +6,10 @@
 		 -->
 	<Target Name="GovUkFrontend_BuildSass" BeforeTargets="BeforeBuild">
 		<PropertyGroup>
+			<UserProfile Condition=" '$(OS)' == 'Windows_NT' ">$(UserProfile)</UserProfile>
+			<UserProfile Condition=" '$(OS)' != 'Windows_NT' ">$(HOME)</UserProfile>
+		</PropertyGroup>
+		<PropertyGroup>
 			<GovUkFrontEndFolder>$(MsBuildThisFileDirectory)..\lib\govuk-frontend\src</GovUkFrontEndFolder>
 		</PropertyGroup>
 		<Message Text="You are using the following version of GOV.UK Frontend. You can run 'git submodule update' to check for updates." Condition="Exists('$(GovUkFrontEndFolder)')" />
--- a/GovUk.Frontend.AspNetCore.Extensions/GovUk.Frontend.AspNetCore.Extensions.csproj
+++ b/GovUk.Frontend.AspNetCore.Extensions/GovUk.Frontend.AspNetCore.Extensions.csproj
@@ -4,7 +4,7 @@
 		<TargetFramework>net6.0</TargetFramework>
 		<AddRazorSupportForMvc>true</AddRazorSupportForMvc>
 		<Nullable>enable</Nullable>
-		<Version>5.0.0</Version>
+		<Version>5.0.1</Version>
 		<!-- Set PackageValidationBaselineVersion to the current major version, eg if you're publishing 1.1.1 set it to 1.0.0. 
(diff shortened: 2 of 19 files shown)


samuelhaa:
I tried to build your branch but I get the following error:
`~/.nuget/packages/aspnetcore.sasscompiler/1.58.1/runtimes/win-x64/src/dart.exe: Exec format error`

I am assuming because it is looking for the windows version of dart rather than the linux version. Is this related or is it a different issue?

Branch: jonnymoo/govuk-frontend-aspnetcore-extensions
Distributor ID:    Ubuntu 
Description:    Ubuntu 22.04.3 LTS 
Release:    22.04 
Codename:    jammy 

dotnet --version 
8.0.100

jonnymuir:
Ah, yeh, that’s a good point. I built the project on windows (in visual studio) to test the build, so I guess if you want the CI build to run on Linux there is an extra work to do. (Like you say - targeting non windows exe’s for dart).

My issue wasn’t with the original build, it was with the nuget restores which used the target file. I.e for  deploying the package on Linux. This doesn’t need to recompile the sass files, so I don’t hit that problem.

There may well be some more hidden gotcha’s I’ll find later on, and I’ll keep you updated if and when i hit them.

samuelhaa:
I've checked the `.targets` file and the `csproj` files. I've built and run the tests on windows and it works fine. I haven't run it through the pipeline though. @sussexrick Is it possible to test it via the pipeline just to make sure that it doesn't break anything?

sussexrick (approved):

sussexrick:
@jonnymoo Thank you vary much for contributing. No worries about the versioning - our CI has a manual trigger to publish to nuget.org. 
@samuelhaa Thank you for doing the testing. I ran the CI build and it works fine.

I'll look to get #244 through code review, then publish out some new versions with this change included.

PR #243, opened 2023-11-21
_tpr-image.scss was being published in the NuGet package by mistake

sussexrick:
`_tpr-image.scss` is compiled into tpr.css which is included in ThePensionsRegulator.Frontend.Umbraco as a static file. `_tpr-image.scss` itself doesn't need to be included, and is getting copied to consuming projects where it's not wanted.

There's no version bump because that's already in the `develop` branch.

The other change in this PR is removing one instance of a duplicate setting in `GovUk.Frontend.Umbraco.csproj` that I noticed while comparing the two projects.

### DIFF
--- a/GovUk.Frontend.Umbraco/GovUk.Frontend.Umbraco.csproj
+++ b/GovUk.Frontend.Umbraco/GovUk.Frontend.Umbraco.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk.Razor">
+﻿<Project Sdk="Microsoft.NET.Sdk.Razor">
 
 	<PropertyGroup>
 		<TargetFramework>net6.0</TargetFramework>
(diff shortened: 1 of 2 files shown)


samuelhaa:
I've looked inside the nuget package  [ThePensionsRegulator.GovUk.Frontend.Umbraco](https://nuget.info/packages/ThePensionsRegulator.GovUk.Frontend.Umbraco/4.3.0) and I can't find `_tpr-image.scss`.

I've built NE and looked for that file, hoping that it would be copied to the consuming project but I can't find it.

Where in the package is it? How would I test this?

sussexrick:
@samuelhaa It's in the v5.0.0 package under contentFiles. NE is on an earlier version.

samuelhaa:
I can see the `_govuk-list.scss` but not `_tpr-image.scss` Am I missing anything here?

image

sussexrick:
@samuelhaa You're looking at the wrong package. It's in ThePensionsRegulator.Frontend.Umbraco. Your screenshot shows ThePensionsRegulator.**GovUk**.Frontend.Umbraco.

samuelhaa (approved):

PR #246, opened 2023-11-30
Add read-only support to Date input, Text input and Textarea components

sussexrick:
[AB#173404](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/173404) [AB#172103](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/172103) both require read-only text input components. 

This PR adds Umbraco support for making a date input, text input or textarea block read-only. It adds a new composition that is applied to the settings of all three components.

The [HTML `readonly` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly) only applies to `input` and `textarea` elements, and then not to certain types of `input` like `input type="file"`.

Because this involves repeating the tags in the views, as far as possible I moved logic out of those tags and up to the top of the views. I also moved hard-coded property aliases into the `PropertyAliases` class so that they're available for reference by consuming projects.

TPR styling is added for read-only fields matching the specification in [AB#173404](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/173404).

Relevant package versions increment the minor version for a non-breaking feature release.


HarryAmmon (changes requested):
This looks good, would be nice to see examples of this working added to the example project as this acts as the documentation

sussexrick:
Examples added to both example apps, on the 'Date input', 'Text input' and 'Textarea' pages.

I also fixed a backoffice bug I found where textarea previews displayed the field label twice:

Duplicated label

HarryAmmon (approved):

PR #249, opened 2023-11-30
Add the ability to hide the day field for a Date input in Umbraco

sussexrick:
[AB#173233](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/173233) requires a 'Date input' component that only asks for the month and year. This PR makes that an option in Umbraco.

Month and year only

It is a follow-up to #246 to avoid an inevitable merge conflict from that one, so it should not be merged until that one has already been completed. The code can be reviewed now by looking at the relevant commit.

The base `<govuk-date-input />` component that we use does not support this option, so we need to use the approach we took for adding validation attributes. This involves creating a tag helper that wraps the original component, reads the HTML output by that component, and updates that HTML using `HtmlAgilityPack`.

The `UmbracoDateInputModelBinder` is updated to know when it should not expect a day field, and in that case it uses 1 for the day so that a `DateTime` or `DateOnly` object can be created. This is why this change is for Umbraco projects only - the `DateInputModelBinder` in the base project doesn't do this, and always expects a day.

The backoffice preview reflects the selection made. 

The Umbraco example app has an example of a month and year date field.


GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/GovukDateInputSettings.generated.cs
```diff
@@ -104,13 +111,6 @@ public GovukDateInputSettings(IPublishedElement content, IPublishedValueFallback
 		[ImplementPropertyType("modelProperty")]
 		public virtual object ModelProperty => global::Umbraco.Cms.Web.Common.PublishedModels.GovukModelProperty.GetModelProperty(this, _publishedValueFallback);
 
-		///<summary>
```
HarryAmmon: Is removing the readonly property intentional?

GovUk.Frontend.Umbraco/HtmlGeneration/DateInputHtmlEnhancer.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            if (dayEnabled) { return html; }
+
+            var document = new HtmlDocument();
+            document.LoadHtml(html);
+
+
+            var wrapperForDayInput = document.DocumentNode.SelectSingleNode($"//div[{WithClass("govuk-date-input__item")} and div[{WithClass("govuk-form-group")} and input[{EndsWith("@id", ".Day")}]]]");
+
+            if (wrapperForDayInput != null)
```
HarryAmmon: Prefer the `is not` syntax for checking null for the same reason we use `string` not `String`

HarryAmmon (changes requested):

sussexrick:
@HarryAmmon Comments addressed, ready for another look

HarryAmmon (approved):

PR #257, opened 2023-12-07
Dependent fields should only be validated when their parent option is selected

sussexrick:
Both the 'Radios' and 'Checkboxes' components support dependent or conditional fields which appear when a particular radio button or checkbox is selected. If you bind a property to one of those dependent fields, and then add validation to that property, the validation should only fire if the relevant radio button or checkbox is selected. This is already working client-side, but not server-side.

This PR adds an `ActionFilter` which runs after validation has happened, but before the model is passed to the controller. It updates the validation result to meet this requirement. `ModelState` for any dependent fields which should not be validated is marked as 'Skipped'.

The types of fields which could appear as dependent fields also needed updating. If `ModelState` contained an error message they always displayed it, but now they only display it if the `ModelState` for that field is marked as `Invalid`. If it is `Skipped` the error message is not displayed. This is also true for the 'Error summary' component.

This PR also updates the 'Checkboxes' page in the example app to include an `[Required]` example of this validation.

In testing you may notice that the validation message incorrectly displays as the field name. That's a separate PR, #255.


GovUk.Frontend.Umbraco/ServiceCollectionExtensions.cs
```diff
@@ -37,13 +36,12 @@ public static IServiceCollection AddGovUkFrontendUmbraco(this IServiceCollection
...
             services.AddTransient<IUmbracoPaginationFactory, UmbracoPaginationFactory>();
             services.AddSingleton<IConfigureOptions<MvcOptions>, ModelBindingMvcConfiguration>();
-            services.AddSingleton<IConfigureOptions<MvcOptions>, RemoveSettingsErrorsMvcConfiguration>();
+            services.AddSingleton<IConfigureOptions<MvcOptions>, ValidationMvcConfiguration>();
             services.AddTransient<ITaskListTaskStatusProvider, TaskListTaskStatusProvider>();
             services.AddTransient<IPropertyValueFormatter, GovUkTypographyPropertyValueFormatter>();
             services.AddTransient<IPropertyValueFormatter, NoParagraphPropertyValueFormatter>();
             services.AddTransient<IPropertyValueFormatter, NoParagraphInversePropertyValueFormatter>();
             services.AddTransient<IPartialViewPathProvider, GovUkPartialViewPathProvider>();
-            services.AddTransient<IDateInputHtmlEnhancer, DateInputHtmlEnhancer>();
```
HarryAmmon: Was removing this service intentional?

HarryAmmon (dismissed):

GovUk.Frontend.Umbraco/ServiceCollectionExtensions.cs
sussexrick: Well spotted, no it wasn't!

HarryAmmon (dismissed):

HarryAmmon (approved):

PR #259, opened 2023-12-11
Allow SetInitialValue to be used in a SurfaceController 

sussexrick:
Currently `SetInitialValue()` can throw an exception if used in a surface controller, if the field was previously marked as invalid. 

Looking at the .NET source, everything else `MarkFieldSkipped` does is also done by `SetModelValue`, so this should also be a fraction faster.

[AB#178662](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/178662)


sussexrick:
Abandoned as #257 means using `SetInitialValue` and therefore setting a field to skipped will prevent an error message from being displayed.

PR #273, opened 2024-02-26
Add tests for parsing month names

sussexrick:
Parsing month names in the 'Date Input' component was added with govuk-frontend 4.6 but unit tests were not added for it.


HarryAmmon (dismissed):
Very minor but would be good to see a test for leap years, if that's supported. 29 February 2023 would be invalid, for example

sussexrick:
@HarryAmmon Leap year tests added

HarryAmmon (approved):

PR #276, opened 2024-03-12
Upgrade to .NET 8 / Umbraco 13.2.0

sussexrick:
- Upgrades Umbraco from 10.x to the latest LTS release, 13.2.0. 
- Upgrade to .NET 8 (required by Umbraco 13).
- Upgrade some other packages (eg uSync, AspNetCore.SASSCompiler).

Ideally the upgrade would've targeted Umbraco 13.0.3 rather than 13.2.0 for wider compatibility with downstream projects. However in 13.0.0, with the upgrade to TinyMCE, Umbraco introduced a new private type for a rich text editor value, which made extending rich text property editor more difficult and fixing our tests for our rich text property editor impossible. An interface was introduced for this in 13.2.0, so this PR targets that version.

When upgrading my own installation I was getting a locking error on `NuCache.Content.db`. If you get the same the solution is to delete the `GovUk.Frontend.Umbraco.ExampleApp\umbraco\Data\TEMP\NuCache` folder and set `UpgradeUnattended` to `false` in `appsettings.json`. Upgrading manually is one button click and does not hit the locking issue.

Some classes have changed because Umbraco 13 has deprecated a number of constructors and scheduled them for removal in future versions, so using the approved constructors gets us ahead of that change.

The renaming of backoffice HTML files to `xxxPreview.html` was the only way I could find to update Umbraco's cache of back office previews. The JSON data structure of rich text editor data has changed and several previews needed updating. Nothing except renaming could clear this cache - rebooting, rebuilding, deleting temp files, clearing database and memory caches... 

I have not bumped the versions of any packages yet because I will also be upgrading to govuk-frontend 5.x as part of this upgrade, and it'll make more sense to bump versions with that as it'll affect every project, not just Umbraco ones.


samuelhaa:
I'm getting boot failed when when I try to launch the `GovUk.Frontend.Umbraco.ExampleApp`.  Have you come across this before? I have deleted the umbraco/ folder in the hope that it would trigger a new installation but it still fails to boot
 
`GovUk.Frontend.ExampleApp` works fine.

image

sussexrick:
@samuelhaa There can be many reasons for that. Take a look at the logs in /umbraco/logs to find out what the error message is.

samuelhaa:
Joel had the same error on the Submit project.
My solution was to delete the connectionstring in appsettings and delete the umbraco/ folder

samuelhaa:
I've run the app and the upgrade works

samuelhaa (approved):

samuelhaa:
Do we not need to bump the version number on the Extensions project? Is that done elsewhere?

sussexrick:
I have not bumped the versions of any packages yet because I will also be upgrading to govuk-frontend 5.x as part of this upgrade, and it'll make more sense to bump versions with that as it'll affect every project, not just Umbraco ones.

ancientyouthman:
Just FTR @samuelhaa  and @sussexrick, my issue on the Submit Umbraco project was due to the UmbracoDbsn connection string (pointing to local SQL) sneaking back into the appsettings.json somehow (I had previously deleted it - i think!). Deleting it again fixed the issue.

PR #279, opened 2024-03-22
Optionally adds data-prevent-double-click="true" attribute to a button.

MarkyYeah:
If a user double clicks a button, this will make the button ignore the second click.


HarryAmmon:
Is the target branch for this PR correct? I was expecting this to come from develop and back into develop.

PR #290, opened 2024-05-08
Add support for the accordion component

sussexrick:
This GOV.UK component has not been supported until now, and gives content designers more flexibility in authoring content. See [Accordion component](https://design-system.service.gov.uk/components/accordion/) on the GOV.UK Design System.

This PR:

- Creates an example page in the ASP.NET example app demonstrating support for `<govuk-accordion />` in `govuk-frontend-aspnetcore`
- Also adds a 'Back link' component to the ASP.NET example app to make navigation easier
- Adds 'Accordion' component styles to `tpr.css` in `ThePensionsRegulator.Frontend` package
- Adds Umbraco support for the 'Accordion' component in all block list and block grid data types that accept a wide range of content
- Adds a backoffice preview for the 'Accordion' component
- Adds an example page in the Umbraco example app, with content identical to the example page in the ASP.NET example app

## Example page

Example page

## Backoffice preview

Backoffice preview including Accordion component

## Editing an accordion section

Umbraco interface for editing an Accordion section


GovUk.Frontend.Umbraco/App_Plugins/GOVUK/blocks/views/GovUkAccordion.html
HarryAmmon, line 1: Should this file be suffixed with "Preview"? Other files that are intended for the WYSIWYG content editor often have that prefix

HarryAmmon:
In the content editor, I am able to select a heading level of 6 and then select a heading level of 3 for the rich text content in the content section. This results in HTML where a `h3` is after a `h6`. Whilst the way it's been done means the HTML is valid (I think), I imagine it might be confusing as the h3 content is related to the h6 content but the different heading levels don't make this apparent. 

Could the rich text element have the available CSS classes restricted to only include p tags? 

image

GovUk.Frontend.Umbraco/App_Plugins/GOVUK/blocks/views/GovUkAccordion.html
sussexrick: That was just the only way I could find to clear the cache for those files the last time they changed.

sussexrick:
@HarryAmmon Thank you for the review. You raise a good point, which I've partially addressed. I've limited the heading levels available for selection as accordion headings to just H2 and H3, which are the most likely candidates. This means you could still create an H2 incorrectly nested below an H3, but I didn't go further for the following reasons:

1. If you look at existing content on our website, it's quite likely that an accordion heading will have a significant amount of text below it, and that text will benefit from supporting subheadings.
2. The rich text editor only goes down to H3 at the moment, but before extending that anywhere in that library we need to do a big exercise in supporting the latest TPR heading styles standards, because they all changed. That's a task on its own, out of scope for the accordion support.
3. Creating separate style lists for different scenarios isn't supported yet, because we only have one backoffice stylesheet, so that's quite an involved task that I don't think would actually solve the problem you've raised.
4. I looked at validating the editor on save to check the selected heading level and compare it to the HTML, but because it's a block editor you only get the JSON. The internal classes that parse the JSON and turn it into block objects work on published content, but validation only has access to unpublished content, so it looks like it would require manual parsing of the JSON and that would be complex and brittle.
5. I don't want to get into trying to do the same validation client-side when all the client-side code is going to get binned when we upgrade from Umbraco 13 to 14+, where the backoffice is rewritten without angularjs.

HarryAmmon (approved):

PR #293, opened 2024-05-14
Feature/189853 validate file type and size

ancientyouthman:


DeniseG123:
All scenarios have been tested - works as expected

DeniseG123 (approved):

sussexrick:
@ancientyouthman It would be good to check the `ContentType` property is valid too. It should be a little harder to fake than a file extension.

PR #311, opened 2024-07-30
Feature/190378 make choose file secondary button

KomalPolavarapu-tpr:


sussexrick (commented):
As you tab through the buttons on the 'File upload' page in the Umbraco example app, the text "No file chosen" shifts position as its adjacent button gains and loses focus (in MS Edge). I would expect that text to remain still.

ancientyouthman (dismissed):

KomalPolavarapu-tpr:
> As you tab through the buttons on the 'File upload' page in the Umbraco example app, the text "No file chosen" shifts position as its adjacent button gains and loses focus (in MS Edge). I would expect that text to remain still.

This has been now fixed

sussexrick:
This is the button style with your changes:

Proposed button

This is a standard secondary button:

Secondary button

As you can see they're still quite different. You can make it much more like a standard secondary button:

Modified file upload button

I added the following styles to DevTools to get the screenshot above:

Proposed styles

`element.style` in this case is the `input[type=file]` element and it affects the 'No file chosen' text, while the rest affects the button. You could set these styles in the SASS to get a better result, but in the SASS I would expect to see them using existing variables to set the font size, font family, colour and spacing (`$govuk-gutter`).

KomalPolavarapu-tpr:
Made the relevant changes

ThePensionsRegulator.Frontend/Styles/_tpr-button.scss
```diff
@@ -19,14 +19,38 @@ $tpr-button-padding: 20px;
     border-radius: 8px;
     box-shadow: none;
     background: $govuk-button-background-colour url(/_content/ThePensionsRegulator.Frontend/tpr/chevron_right_white_24dp.svg) right center no-repeat;
+    padding: 8px 10px 7px;
```
sussexrick: You're setting the left and right padding unnecessarily here, because the values are immediately overwritten by the next properties. 

Also, your intention is to set the top and bottom padding of `.govuk-file-upload::file-selector-button` but you're also resetting `.govuk-button`, so if you're going to put it here I would use the [original setting which you can see on lines 73 and 74 in govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/button/_index.scss).

ThePensionsRegulator.Frontend/Styles/_tpr-button.scss
```diff
@@ -19,14 +19,38 @@ $tpr-button-padding: 20px;
...
     padding-left: $tpr-button-padding;
     padding-right: $tpr-button-padding *2;
 }
+input[type="file"] {
```
sussexrick: All of the references in this file to `input[type="file"]` should be scoped to `.govuk-file-upload`.

ThePensionsRegulator.Frontend/Styles/_tpr-button.scss
```diff
@@ -19,14 +19,38 @@ $tpr-button-padding: 20px;
...
+    font-size: 1rem;
+    outline: 3px solid transparent;
+    border-color: transparent;
+    box-shadow: inset 0 0 0 2px transparent;
+    padding: 3px 3px 1px 3px;
+    font-family: $govuk-font-family;
+}
+
+input[type="file"]::file-selector-button {
```
sussexrick: Rather than re-specifying all these properties, a simpler approach might be to do

```css
.govuk-file-upload::file-selector-button {
   @extend .govuk-button;
   @extend .govuk-button--secondary;
}
```

...so that all the styles can be set on the secondary button and just imported into the file-upload button. Not sure if this will give you what you need, but if it works it could remove a lot of lines.

ThePensionsRegulator.Frontend/Styles/_tpr-button.scss
```diff
@@ -19,14 +19,38 @@ $tpr-button-padding: 20px;
...
     padding-left: $tpr-button-padding;
     padding-right: $tpr-button-padding *2;
 }
+input[type="file"] {
+    font-size: 1rem;
+    outline: 3px solid transparent;
+    border-color: transparent;
+    box-shadow: inset 0 0 0 2px transparent;
+    padding: 3px 3px 1px 3px;
+    font-family: $govuk-font-family;
```
sussexrick: You also need to set `color: $govuk-text-colour` here as the "No file chosen" text is too dark. You can see that it's not the same as the text above it:

image

sussexrick (changes requested):

ThePensionsRegulator.Frontend/Styles/_tpr-button.scss
KomalPolavarapu-tpr: @sussexrick  I have created a new sass for file upload and moved all related code there, please have a look

ThePensionsRegulator.Frontend/Styles/_tpr-button.scss
KomalPolavarapu-tpr: I have corrected the padding properties that are being overridden. and the padding that I set is the one for .govuk.button. I have added that again here to apply the same properties to the fileupload button, but now I have separated that into its own scss

ThePensionsRegulator.Frontend/Styles/_tpr-button.scss
```diff
@@ -55,6 +49,7 @@ a.govuk-button--secondary:link, a.govuk-button--secondary:visited {
 .govuk-file-upload::file-selector-button:hover,
 .govuk-button--secondary:link:active,
 .govuk-button--secondary:visited:active,
+input[type="file"]::file-selector-button:hover,
```
sussexrick: You shouldn't need `input[type="file"]::file-selector-button:hover` and `.govuk-file-upload::file-selector-button:hover`. 

If the component doesn't have `.govuk-file-upload` applied we shouldn't be styling it.

Also, you've moved all the other file upload styles to `_tpr-file-upload.scss` (which is good) but why not this?

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
```diff
@@ -0,0 +1,36 @@
+﻿.govuk-file-upload {
+    font-size: 1rem;
```
sussexrick: Use variables rather than hard-coding. To be the same as `.govuk-button` set body text font size with `@include govuk-font($size: 19, $line-height: 19px);` rather than setting `font-size` directly. (see https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/button/_index.scss)

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
```diff
@@ -0,0 +1,36 @@
...
+    outline: 3px solid transparent;
```
sussexrick: Use variables rather than hard-coding. `3px` here should be `$govuk-focus-width`

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
```diff
@@ -0,0 +1,36 @@
...
+    border-color: transparent;
+    box-shadow: inset 0 0 0 2px transparent;
+    padding: 3px 3px 1px 3px;
```
sussexrick: This padding can be removed. It actually causes the button to be misaligned relative to other buttons on the page.

buttons not aligned

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
```diff
@@ -0,0 +1,36 @@
...
+    font-family: $govuk-font-family;
+    color: $govuk-text-colour
+}
+
+.govuk-file-upload::file-selector-button {
+    padding-top: 8px;
+    padding-bottom: 7px;
+    padding-left: $tpr-button-padding;
```
sussexrick, line 18: I would expect top and bottom padding to be the same as `.govuk-button` and therefore set in the same way. Top is `(govuk-spacing(2) - $govuk-border-width-form-element)` and bottom is `(govuk-spacing(2) - $govuk-border-width-form-element - ($button-shadow-size / 2))`. See https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/button/_index.scss

When govuk-frontend changes we want to adapt automatically and consistently. Hard-coded values prevent that.

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
```diff
@@ -0,0 +1,36 @@
...
+    padding-right: $tpr-button-padding *2;
+    font-family: $govuk-font-family;
+    border: 1px solid $tpr-colour-royal-blue;
```
sussexrick: Don't hard-code, use `$govuk-border-width-form-element`

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
```diff
@@ -0,0 +1,36 @@
...
+    margin-right: 1rem;
```
sussexrick: Don't hard-code, use `govuk-spacing(2)`

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
```diff
@@ -0,0 +1,36 @@
...
+    background-color: $tpr-colour-tangerine-yellow;
+    color: $tpr-colour-black;
+    border-left-color: transparent;
+    border-right-color: transparent;
+    border-top-color: transparent;
+}
+
+.govuk-file-upload:focus-within,
+.govuk-file-upload:focus {
+    outline: 3px solid #ffdd00;
```
sussexrick: Don't hard-code, use `$govuk-focus-width`. 

Where has `#ffdd00` come from? It should be `$govuk-focus-colour` which is not the same colour.

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
```diff
@@ -0,0 +1,36 @@
...
+    border-color: #0b0c0c;
```
sussexrick: `border-color` appears to be having no effect here and can be removed.

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
```diff
@@ -0,0 +1,36 @@
...
+    box-shadow: inset 0 0 0 2px #0b0c0c;
```
sussexrick: This colour should be set to `$govuk-focus-text-colour`

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
```diff
@@ -0,0 +1,36 @@
...
+    padding: 3px 3px 1px 3px;
```
sussexrick: See earlier comment about removing this padding. You wouldn't need to re-specify it on focus anyway.

ThePensionsRegulator.Frontend/Styles/_tpr-button.scss
sussexrick: Did you try this?

sussexrick (changes requested):

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
KomalPolavarapu-tpr: Without the padding the Removed

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
KomalPolavarapu-tpr: done changed

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
KomalPolavarapu-tpr: done removed

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
KomalPolavarapu-tpr: Corrected this

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
KomalPolavarapu-tpr: If the padding is removed, there is no gap between the shadow and the button
With padding 
image
Without padding
image

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
KomalPolavarapu-tpr: done

ThePensionsRegulator.Frontend/Styles/_tpr-button.scss
KomalPolavarapu-tpr: removed

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
KomalPolavarapu-tpr: done

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
KomalPolavarapu-tpr: removed

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
sussexrick: This should be `$govuk-focus-width` because it's here so that it can be made visible on focus.

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
sussexrick: You're still not using `$govuk-focus-width` for the width of the focus ring.

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
```diff
@@ -0,0 +1,43 @@
...
+    padding: govuk-px-to-rem(3) govuk-px-to-rem(3) govuk-px-to-rem(1) govuk-px-to-rem(3);
+    font-family: $govuk-font-family;
+    color: $govuk-text-colour
+}
+
+.govuk-file-upload::file-selector-button {
+    padding-top: (govuk-spacing(2) - $govuk-border-width-form-element);
+    padding-bottom: (govuk-spacing(2) - $govuk-border-width-form-element - ($button-shadow-size / 2));
+    padding-left: $tpr-button-padding;
+    padding-right: $tpr-button-padding *2;
```
sussexrick: Your left and right padding should both be `$tpr-button-padding`. You've got one-sided spacing because you've copied values from a button with a chevron, but there is no chevron here.

image

KomalPolavarapu-tpr: yes that's right, I have applies the same style as button but changed now as suggested

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
KomalPolavarapu-tpr: changed as discussed and suggested

ThePensionsRegulator.Frontend/Styles/_tpr-file-upload.scss
KomalPolavarapu-tpr: changed as discussed and suggested

ThePensionsRegulator.Frontend/Styles/_tpr-button.scss
KomalPolavarapu-tpr: just extending them didn't give me what is needed so customised them accordingly

sussexrick (approved):

PR #314, opened 2024-08-05
Add BinaryFileValidators for Excel and PDF #195851

samuelhaa:
Because
- We need to validate the file headers to ensure that the file is what
  it says it is.

This is using the the [FileSignatures](https://github.com/neilharvey/FileSignatures) repo on Github for the signature validation

Each file type will have a a certain set of bytes that defines the file
type. We search the file for that signature to determine whether the uploaded file is a valid file.

A Google search on "File signatures" would give the file signatures for the most common file types.

We define the `[AllowedFileTypeAttribute]` that replaces `[AllowedFileExtensionAttribute]`

Ref:
- [File signature validation on Microsoft.com](https://learn.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads?view=aspnetcore-5.0#file-signature-validation)

AB #195851


GovUk.Frontend.AspNetCore.Extensions.Tests/GovUk.Frontend.AspNetCore.Extensions.Tests.csproj
```diff
@@ -22,4 +31,23 @@
     <ProjectReference Include="..\GovUk.Frontend.AspNetCore.Extensions\GovUk.Frontend.AspNetCore.Extensions.csproj" />
   </ItemGroup>
 
+
+
+	<!--<Target Name="CopyYourFolderAfterPublish" AfterTargets="Publish">
```
sussexrick: Can all this unused MSBuild code be removed?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/file-upload.config
```diff
@@ -131,9 +131,8 @@
       "fileTypes": ".csv",
       "errorMessagePrefix": "",
       "cssClassesForRow": "govuk-grid-row--tpr-divider",
-      "errorMessageMaxFileSize": "",
-      "errorMessageAllowedFileExtensions": "",
-      "cssClassesForColumn": ""
+      "cssClassesForColumn": "",
+      "ErrorMessageAllowedFileTypes": ""
```
sussexrick: The custom error messages are missing in your examples.

unhelpful errors

You're also not filtering the file types on your two new examples. The two existing examples do this.

Select file dialog showing 'all files' filter


You're also missing a capital E on Excel and a fullstop at the end of your sentence, and PDF should be all-caps.

GovUk.Frontend.AspNetCore.Extensions/ServiceCollectionExtensions.cs
```diff
@@ -2,6 +2,7 @@
 using GovUk.Frontend.AspNetCore.Extensions.ModelBinding;
 using GovUk.Frontend.AspNetCore.Extensions.Security;
 using GovUk.Frontend.AspNetCore.Extensions.Validation;
+using GovUk.Frontend.AspNetCore.Extensions.Validation.BinaryFileValidators;
```
sussexrick: This is an unused declaration

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
```diff
@@ -1,97 +1,81 @@
...
-			<Validation></Validation>
-			<Description><![CDATA[Sets the message displayed if the file uploaded exceeds the maximum size]]></Description>
-			<SortOrder>0</SortOrder>
-			<Tab Alias="errorMessages">Error messages</Tab>
-			<Variations>Nothing</Variations>
-			<MandatoryMessage></MandatoryMessage>
-			<ValidationRegExpMessage></ValidationRegExpMessage>
-			<LabelOnTop>false</LabelOnTop>
-		</GenericProperty>
-		<GenericProperty>
```
sussexrick, line 64: You've removed the `errorMessageMaxFileSize` property. I don't think you meant to because other code still refers to it.

GovUk.Frontend.AspNetCore.Extensions/Validation/BinaryFileValidators/FileType.cs
```diff
@@ -0,0 +1,206 @@
+﻿using System.Collections.ObjectModel;
```
sussexrick: It would be better to consume the NuGet package rather than selectively copy the code. By copying it we're taking on the maintenance of something that is being actively maintained elsewhere, while simultaneously limiting our options by not including all the other file types that other use cases may want. You've also neglected to copy the unit tests along with the code - we wouldn't expect to unit test this code if we consumed the package where it's already tested.

GovUk.Frontend.AspNetCore.Extensions.Tests/AllowedFileTypesAttributeTests.cs
```diff
@@ -0,0 +1,74 @@
...
...DiRl98UpscE2RygAVHBKxES5RWUqJtoTO4YDmw0sTcGeI272VsGmfhNtqXDgLJq7K8lnAgCDXUl+kUKKbEVU8fDa2jHfjwcXdMDKzVt5S8s4b4lvqnszlibKj4frDglZyLium2YF+yo6MulZy3amuNhzUH68Z4BCXfBuoezLC0jXEZtepp1YOlmAt0f3htV6L4bRAGnEr0JjsTiLEG29SMtU9IWT/F/IwtAKGSbJiGYzn3zmv3RS9HAxfnxiFgAmHhHHHnyAP+ajYm0zvEyznxyDDxTjjbgW86c843XplP+id7HbtkwpGFU/XDhWd8SLt4awhe13k+VNvWZKj5BU7rPg3UPW8y+yFk3Zqwh/rV878wPP7j9MP18npRfi75XWczJd/+sv4LAAD//wMAUEsBAi0AFAAGAAgAAAAhAKRTxc9OAQAACAQAABMAAAAAAAAAAAAAAAAAAAAAAFtDb250ZW50X1R5cGVzXS54bWxQSwECLQAUAAYACAAAACEAtVUwI/QAAABMAgAACwAAAAAAAAAAAAAAAACHAwAAX3JlbHMvLnJlbHNQSwECLQAUAAYACAAAACEAOK1CBWYDAACrCAAADwAAAAAAAAAAAAAAAACsBgAAeGwvd29ya2Jvb2sueG1sUEsBAi0AFAAGAAgAAAAhAI2H2nDgAAAALQIAABoAAAAAAAAAAAAAAAAAPwoAAHhsL19yZWxzL3dvcmtib29rLnhtbC5yZWxzUEsBAi0AFAAGAAgAAAAhAI8wCwz3AQAApQMAABgAAAAAAAAAAAAAAAAAXwwAAHhsL3dvcmtzaGVldHMvc2hlZXQxLnhtbFBLAQItABQABgAIAAAAIQDBFxC+TgcAAMYgAAATAAAAAAAAAAAAAAAAAIwOAAB4bC90aGVtZS90aGVtZTEueG1sUEsBAi0AFAAGAAgAAAAhAHmhgGykAgAAUgYAAA0AAAAAAAAAAAAAAAAACxYAAHhsL3N0eWxlcy54bWxQSwECLQAUAAYACAAAACEA4ysQi0gBAABxAgAAEQAAAAAAAAAAAAAAAADaGAAAZG9jUHJvcHMvY29yZS54bWxQSwECLQAUAAYACAAAACEAYUkJEIkBAAARAwAAEAAAAAAAAAAAAAAAAABZGwAAZG9jUHJvcHMvYXBwLnhtbFBLBQYAAAAACQAJAD4CAAAYHgAAAAA=";
```
sussexrick: It's not clear why you have actual files to run tests and also these byte strings?

Also, I can see what you're doing, but `ExcelDataProvider` is not a clear name for something that provides a PowerPoint file.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
```diff
@@ -1,97 +1,81 @@
...
-		</Compositions>
-		<DefaultTemplate></DefaultTemplate>
-		<AllowedTemplates />
-	</Info>
-	<Structure />
-	<GenericProperties>
-		<GenericProperty>
-			<Key>cac445e1-9d87-452e-94de-a70e43ff09ff</Key>
-			<Name>Allowed file types</Name>
-			<Alias>errorMessageAllowedFileExtensions</Alias>
```
sussexrick, line 35: You've removed the `errorMessageAllowedFileExtensions` property but `ApplicationBuilderExtensions.cs` still refers to it and the validator that uses it still exists, you've only marked it as obsolete.

I had a chat with Joel about whether this validator should replace his earlier effort, and we agreed that the validation you're putting in place is better. However, Joel's suggestion was to incorporate the file extension check into this validator as a first pass which can fail fast, before reading into a `MemoryStream` for a further check. I agree with this as it'll give faster feedback to the user and helps to guard against an attacker uploading potentially malicious files into our memory. 

I can see you had a similar idea and have added an overload for `IsMatch` that checks the extension, but the suggestion here is that you would fail the validator as soon as the extension test fails, without ever doing the binary check in that case. Plus if you're consuming the package this logic would be in the validator rather than an `IsMatch` overload.

GovUk.Frontend.AspNetCore.Extensions/Validation/AllowedFileExtensionsAttribute.cs
```diff
@@ -6,9 +6,11 @@
 
 namespace GovUk.Frontend.AspNetCore.Extensions.Validation
```
sussexrick: I would expect to see some documentation added to https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/docs/components/file-upload.md similar to https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/docs/components/text-input.md

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
```diff
@@ -1,97 +1,81 @@
...
+  <GenericProperties>
+    <GenericProperty>
+      <Key>02492b18-6710-4c30-872c-8019f4ba037c</Key>
+      <Name>File Type</Name>
+      <Alias>errorMessageFileType</Alias>
+      <Definition>0cc0eba1-9960-42c9-bf9b-60e150b429ae</Definition>
+      <Type>Umbraco.TextBox</Type>
+      <Mandatory>false</Mandatory>
+      <Validation></Validation>
+      <Description><![CDATA[Error message for file type validation]]></Description>
```
sussexrick: image

Your property display name should be "File types" with a lowercase t like the one above it, and plural because the validator supports multiple types at once.

The description you have is not helpful as it just repeats the section heading and property heading. "Sets the message displayed if an uploaded file is not an allowed file type."

Lastly, your new property should be after Required on this list. Required is always shown first for all the component types.

sussexrick (changes requested):

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
samuelhaa: I can't move Required to the top because the ordering is locked.

image

GovUk.Frontend.AspNetCore.Extensions/Validation/BinaryFileValidators/FileType.cs
samuelhaa: Are you saying we should rather use FileSignature package that this is based on?

The reason for not adding the package is to reduce dependencies. Didn't we upgrade a site recently with a lot of deprecated dependencies? The package also contains a lot of types we don't need and the implementation doesn't exactly suit our needs but if you prefer to add the dependency, we can do that. That's ok.

GovUk.Frontend.AspNetCore.Extensions.Tests/GovUk.Frontend.AspNetCore.Extensions.Tests.csproj
samuelhaa: Removed

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/file-upload.config
samuelhaa: - Capitalised E in Excel.  
- Added punctuation at the end of the hints.
- Capitalised PDF
- Added custom error messages

image

- Added filter for Excel files

image

- Added filter for PDF files

image

GovUk.Frontend.AspNetCore.Extensions/ServiceCollectionExtensions.cs
samuelhaa: Removed unused usings

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
samuelhaa: Added MaxFileSize back. It seems to remove it by itself somehow because it was removed after I clicked Save serveral times now.
image

GovUk.Frontend.AspNetCore.Extensions.Tests/AllowedFileTypesAttributeTests.cs
samuelhaa: Changed it to use real files. Renamed the `DataProvider`

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
samuelhaa: Moved the extension validation to the `AllowedFileTypesAttribute`

The reason why `AllowedFileExtensionsAttribute` is marked as Obsolete is because I don't know where it is used and do not want to create a breaking change. Marking is Obsolete allows people to migrate in an orderly fashion and we can delete it in a few months or so when necessary.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
sussexrick: That's fine, I meant it's at the top of the error messages. Error message prefix is a little different.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
sussexrick: Yes, I understand your thinking on the use of `[Obsolete]`. 

However, `ApplicationBuilderExtensions.cs` connects the obsolete `AllowedFileExtensionsAttribute`  to a property with the alias `errorMessageAllowedFileExtensions`, so by removing that property you've broken it.

In this case I would recommend removing it because it was only added in v6.0.0-alpha004 and we haven't published v6.0.0 yet. It's the nature of alphas that features come and go, and projects use them at their own risk.

GovUk.Frontend.AspNetCore.Extensions/Validation/BinaryFileValidators/FileType.cs
sussexrick: Yes, I am making that recommendation. Dependencies vs maintaining extra code yourself is always a trade-off. In this case it's a small but relatively complex domain which is being actively maintained and doesn't bring in a load of further dependencies. It's also highly likely that we'll need other file types supported in other scenarios. That seems like a good trade.

Why do you say it doesn't exactly suit our needs?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
sussexrick: It's not there. Are you updating the copies in both `GovUk.Frontend.Umbraco` and `GovUk.Frontend.Umbraco.ExampleApp`? If you only update the example app, the build will copy the old version back from `GovUk.Frontend.Umbraco`. You need to copy your changes there.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
```diff
@@ -30,68 +30,68 @@
 	<Structure />
 	<GenericProperties>
 		<GenericProperty>
-			<Key>cac445e1-9d87-452e-94de-a70e43ff09ff</Key>
-			<Name>Allowed file types</Name>
-			<Alias>errorMessageAllowedFileExtensions</Alias>
+			<Key>02492b18-6710-4c30-872c-8019f4ba037c</Key>
+			<Name>File types</Name>
+			<Alias>ErrorMessageAllowedFileTypes</Alias>
```
sussexrick: The casing of this alias is inconsistent with all others. It should be `errorMessageAllowedFileTypes`.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
samuelhaa: Removed `[AllowedFileExtensions]`

GovUk.Frontend.AspNetCore.Extensions/Validation/BinaryFileValidators/FileType.cs
samuelhaa: Added the Nuget package

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
samuelhaa: Added it back. It's done in this commit: [Add MaxFileSize back to FileUploadSettings](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/pull/314/commits/19775c7dfdebe0fd9c069075ca1884220b8cc8ee)

GovUk.Frontend.AspNetCore.Extensions/Validation/AllowedFileExtensionsAttribute.cs
samuelhaa: Added documentation

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
sussexrick: I can see you've fixed this but you need to re-export the uSync data for the example app as it still has the old version in page data. Otherwise that change will creep into someone's unrelated PR.

docs/components/text-input.md
```diff
@@ -53,3 +53,23 @@ Error messages for text input components are configurable in Umbraco on the sett
 For a `[RegisteredCharityNumber]` attribute set the 'Text input width' setting to 'small' and use the 'Pattern' error message.
 
 ![Settings for a postcode using a text input component](/docs/images/registered-charity-number-settings.png)
+
+
+## Allowed file fypes
```
sussexrick: Looks like this is a draft you forgot to remove?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
samuelhaa: Did a Clean Export of the ExampleApp and copied the `FileUploadsettings.config` to the `GovUk.FrontEnd.Umbraco` project

docs/components/text-input.md
samuelhaa: Removed the AllowedFileType section from TextInput.md

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukfileuploadsettings.config
sussexrick: It still didn't update the file upload page, but I've done that and will approve this PR.

sussexrick (approved):

PR #315, opened 2024-08-08
Bugfix/200464 adding no chevron button

Syed-kainos:
updating version


Syed-kainos:
Creating a new branch for version update

PR #318, opened 2024-08-21
Upgrade to govuk-frontend v5.2.0 and GovUk.Frontend.AspNetCore v2.2.0

### LINKED ISSUE #302: Update to release 2.0.0 of govuk-frontend-aspnetcore
Test that existing components still work/look as intended.

sussexrick:
`options.AddImportsToHtml` no longer exists as the scripts that triggered our content security policy are rendered in a different way by the base library.

Setting `z-index: 1` for radios and checkboxes was a bug in GOV.UK Frontend v5.1.0, fixed in v5.2.0.

Closes #302.


GovUk.Frontend.Umbraco/ServiceCollectionExtensions.cs
```diff
@@ -19,8 +19,7 @@ public static class ServiceCollectionExtensions
     {
         public static IServiceCollection AddGovUkFrontendUmbraco(this IServiceCollection services)
         {
-            // Avoid adding scripts which require 'unsafe-inline' in the content security policy
-            return services.AddGovUkFrontendUmbraco(options => { options.AddImportsToHtml = false; });
+            return services.AddGovUkFrontendUmbraco(options => { });
```
ancientyouthman, line 22: do we need to pass the options delegate parameter at all ?

ancientyouthman: (here and elsewehre)

sussexrick: Yes, otherwise it would call the same overload in an infinite loop.

It's worth retaining the overload that allows options to be passed in case a consuming project wants to do that.

ancientyouthman (dismissed):

ancientyouthman (approved):

PR #324, opened 2024-09-12
Support three and four column grids 

sussexrick:
Up to this point our block grid support in Umbraco has only included a range of two-column layouts. This PR adds support for three and four column layouts using equal-width columns. A three column layout is required for [AB#204415](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/204415), and one existing TPR application uses three and four-column layouts with block list.

## Backoffice layout options

A selection of layouts to choose from in the backoffice

## Backoffice view of content in columns

Content in three and four column grids in the Umbraco backoffice

## Frontend view of content in columns

This is a screenshot of the updated grid page in the Umbraco example app.

Content in three and four column grids

## Deprecating grid layout in block list

This PR also deprecates the use of block list to create grids, in favour of using block grid which is a better user experience. Grid support only exists in block lists because the block grid editor did not exist when we needed it.

This deprecation is just a change to a label visible in the backoffice only. In a future release we could remove this support and the maintenance overhead of the code that supports it, but only once all TPR systems are no longer using it.

Updated label in backoffice


HarryAmmon (approved):
Happy to approve as is. 

Only suggestion would be that as a grid layout in block list is being depreciated, the content for the grid page on the example app could be updated to show that this feature is deprecated. Given this page acts as the documentation for the feature, I would expect to see that information there

PR #331, opened 2024-10-01
Update task list to GOV.UK standard

sussexrick:
Our task list implementation was based on pre-release code from GOV.UK. With the release of GOV.UK Frontend 5.0 it became an official component, with HTML and CSS that differed from our pre-release version.

Our base library `govuk-frontend-aspnetcore` is already at GOV.UK Frontend 5.2, but it doesn't implement this component, so it's still on us to do that.

This PR:
- Renames the `GovUk.Frontend.AspNetCore.Extensions.Tests` project to `GovUk.Frontend.AspNetCore.Extensions.UnitTests`, and introduces a new test project `GovUk.Frontend.AspNetCore.Extensions.ConformanceTests`
- Adds a new dependency on the `govuk-frontend` npm package, because it has test fixtures which aren't in the repo that we already consume as a submodule.
- Adds conformance tests for the task list component. These fetch the GOV.UK test fixtures from the `govuk-frontend` npm package, parses the JSON into classes and then feeds the settings from those tests into our task list generator. It then compares the resulting HTML with the GOV.UK expected HTML. In this way we can be sure we now conform to the standard.

   This code is adapted from James Gunn's code in `govuk-frontend-aspnetcore`, and when tests fail you get useful feedback like this:
   
  Failed conformance test
  
  We can adopt a similar approach for testing our own TPR components at some point (not for this PR).

- Removes GOV.UK pre-release SASS code for the task list, as we can now rely on the release version. Updates TPR SASS code to match the new HTML.
- Adds a new option to the task list status tag helper to decide whether to render the status within a GOV.UK tag component. For GOV.UK styles this is optional, but for TPR styles it's required so we always do it.
- Refactors the existing HTML generator for task list so that it's more readable, not one massive method.
- Refactors the method which turns a task status enum into English text. This was in an internal method of our HTML generator that had been exposed and used by a view in the Umbraco project - a hack. It's now an extension method on the enum.
- Updates documentation.

**You may need to rebuild the new conformance tests project in order to get the tests to pass. I've logged this as #332.**

The end result is a task list that looks the same as before, but using GOV.UK HTML and CSS:

Task list with TPR styling


GovUk.Frontend.AspNetCore.Extensions.ConformanceTests/Assert.cs
```diff
@@ -0,0 +1,82 @@
...
+                var sb = new StringBuilder();
+                sb.AppendLine($"{nameof(Assert)}.{nameof(HtmlEqual)}() failure");
+                sb.AppendLine();
+
+                foreach (var diff in diffs)
+                {
+                    DiffConverter.Append(diff, sb);
+                }
+
+                throw new InvalidOperationException(sb.ToString());
```
HarryAmmon: Throwing an `InvalidOperationException()` here makes me think there is something wrong with the test code itself, maybe the test code is in a state where this class doesn't know how to make an assertion. 

Instead, we could use NUnits `Assert.Fail(sb.ToString());`. This generates a more uniform message in Test explorer and would be the message I would expect to see if my implementation of a component no longer conforms

GovUk.Frontend.AspNetCore.Extensions.ConformanceTests/TextOrHtmlHelper.cs
```diff
@@ -0,0 +1,24 @@
+using Microsoft.AspNetCore.Html;
+using System.Text.Encodings.Web;
+
+namespace GovUk.Frontend.AspNetCore.Extensions.ConformanceTests
+{
+    public static class TextOrHtmlHelper
+    {
+        public static IHtmlContent? GetHtmlContent(string? text, string? html)
+        {
+            if (html != null)
```
HarryAmmon: IMO it's better is use `is not` for null checks, as the `!=` operator can be overiden where `is not` can't be. Same argument for using the `string` alias instead of `String`

HarryAmmon (changes requested):

GovUk.Frontend.AspNetCore.Extensions.ConformanceTests/TextOrHtmlHelper.cs
HarryAmmon: And `is not` is used in other parts of the solution

GovUk.Frontend.AspNetCore.Extensions.ConformanceTests/Assert.cs
sussexrick: Agreed and updated.

GovUk.Frontend.AspNetCore.Extensions.ConformanceTests/TextOrHtmlHelper.cs
sussexrick: Agreed and updated.

HarryAmmon (approved):

PR #333, opened 2024-10-02
Replace and document the extension point for custom rich text editor data types

sussexrick:
Currently there is an undocumented setup for custom rich text editor data types which involves many levels of inheritance. A specific site would have to create a `PropertyValueConverter` that inherits from TPR, which inherits from GOV.UK, which inherits from our core project, which inherits from Umbraco. Each time you inherit you have to know the internals of the class you're inheriting so you remove it from DI and re-register the same aliases as before, as well as your new alias(es).

This PR replaces that with just one level of inheritance - our core project inherits from Umbraco. To add a new rich text property editor you implement a simple interface that returns only your new alias(es), and register your implementation with DI. It's also documented. 

[AB#206408](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/206408)


MarkyYeah (approved):
Looks good to me - approved!

PR #342, opened 2024-10-07
Publish v6.0.0, and start publishing all packages together

sussexrick:
Upgrade all packages to v6.0.0, representing the upgrade of Umbraco to v13.x and GOV.UK Frontend to v5.x (which we've been using for a while under the alpha versions).

Move the version number out of each individual package into a single file for the repo. This means it only has to be updated in one place, and all packages are published together at the same version. Running the pipeline gets simpler - no more selecting the wrong package from a list with similar names.

With this change sometimes packages will publish a new version with no changes (because there are changes elsewhere... but we do that now), but it also means that you'll never need to install a new version of a transitive package because the top-level one isn't upgraded, or remember to upgrade the top-level one to avoid installing the transitive one. You'll just be able to install/upgrade the top-level package and know you have the latest version of everything that was built and tested together.

Remove use of `<PackageValidationBaselineVersion>` which has not raised the warnings it promises to.


samuelhaa (commented):
I'm getting this error when I compile

`Error (active)	MSB3073	The command "yarn install" exited with code 9009.	GovUk.Frontend.AspNetCore.Extensions.ConformanceTests	C:\dev\govuk\govuk-frontend-aspnetcore-extensions\GovUk.Frontend.AspNetCore.Extensions.ConformanceTests\Directory.Build.targets	10`

samuelhaa (commented):
Together with `'yarn' is not recognized as an internal or external command`

samuelhaa:
Happy otherwise.

sussexrick:
@samuelhaa These dependencies have always been required to run the JavaScript tests, and were documented on the 'Run tests' page. I've now repeated that on the 'Contributing' docs, and got the project file detecting the situation and displaying more helpful error messages. I tried getting it to do some of the install automatically, but commands that work fine outside of MSBuild were not working when called from MSBuild.

If you don't have Node.js installed, instead of the "code 9009" error you will now get this:

Error Node.js not installed

If you have Node.js but not Yarn Classic you'll get this:

Error Yarn Classic not installed

samuelhaa (approved):

PR #351, opened 2024-10-30
Document how to unit test a block has been overridden

### LINKED ISSUE #347: Some properties created with UmbracoBlockListFactory do not have their values overwritten by OverrideValue() whilst some do
When calling `.Content.OverrideValue(alias, value)` on a `Link` property as part of a unit test, where the blocks have been created using `UmbracoBlockListFactory`, the URL (and other properties) are changed, this is the expected functionality.

When calling `.Content.OverrideValue(alias, value)` on `text` property of a `GovukTypography` block as part of a unit test, where the blocks have been created using `UmbracoBlockListFactory`, the original value is retained. This is inconsistent with the behaviour of overriding a `Link`

This means as part of a unit test, you can Assert that a `link.Url` has the correct, overriden value but on a text property you can only verify that `.Content.OverrideValue(alias, value)` was called with the correct parameters.

Anecdotally we have had trouble unit testing that legends, hints and text are overriden properly, this may explain why.

I will try to create a branch with tests that demonstrate this bug, I will update this issue if I do

HarryAmmon:
Closes #347


docs/umbraco/unit-testing.md
```diff
@@ -152,6 +152,29 @@ dictionary
     .AddDictionaryValue("myKey", "myValue", 1); // Setting languageId to 1 - English US
 ```
 
+## Testing content is overridden  
+Blocks in Umbraco can be overridden with blocks generated by your code. This is typically done when populating a summary list with data the user has entered or replacing a token in a string with a value. You should write tests to prove this happens correctly.
```
sussexrick: While this first sentence is true, the more common scenario and the one you're describing is that "Properties in Umbraco can be overridden with values generated by your code."

Also, I misread the heading as saying that "testing content" had been overridden. "Test that content is overridden" would be clearer.

sussexrick (approved):

PR #354, opened 2024-11-11
Support boxed content and more flexible multi-column layouts

sussexrick:
# Objective of this PR

We need to support adding boxed content in a block grid in an editor-friendly way - specifically two types of box used on our corporate website:

* a solid box with a grey background
* a grey border around the standard white background

It was already possible to add the first style of box, but only within a block list by applying a well-known HTML class to a grid row.

We need to not break the existing support for adding a dividing line after the box.

# Implementation
## Box support for TPR styles

The `.tpr-box` class already existed, but could only be applied to a `.govuk-grid-row`, which would style its child `.govuk-grid-column` as a box. This is still supported, making this a non-breaking change. However it is now possible to use the class more simply, which is good for non-Umbraco applications:

```html
<div class="tpr-box">
    <p class="govuk-body">My content here</p>
</div>
```

A new page is added to the ASP.NET example app demonstrating boxes and divider lines, similar to the one in the Umbraco app.

In the Umbraco block grid, for TPR data types only because this is a TPR style, you can now add a 'Box' layout:

Adding a box layout

Once you've added the layout, you can set the style to either 'Solid' or 'Bordered' in the settings (the default is solid).

Settings for a box

In the absence of a documented TPR standard for the border width, I used the `$govuk-border-width` variable to remain consistent with other components (although only 'Details' and 'Error summary' use this). It is set to `5px` by default, whereas our current TPR website uses `4px`, so this is a point for design review.

A box can be added inside a column of a multi-column layout, or a multi-column layout can be added inside a box. You can even have boxes within boxes, and this is all reflected in the backoffice previews:

Backoffice preview of multiple columns and nested boxes

## Improved column support for GOV.UK and TPR styles

To match the support in the block list for adding a divider after a box, a setting for adding a custom class to a grid row was added to all grid layouts.

To enable this even for single-column layouts, new single column layouts are available for all the supported widths - one quarter, one third, one half, two thirds, three quarters and full width. In TPR styles these new single column layouts also allow a box to be created at any of those widths, by adding a box inside the layout.

There are lots of GOV.UK document types related to building grids now, so I gathered them in a new 'Grid' subfolder.

To enable multi-column layouts inside a box in TPR styles, all layouts are now enabled within areas. In either GOV.UK or TPR styles this means nested multi-column layouts are also possible, eg two equal columns inside the two-thirds column of a two-thirds/one-third layout.

The example page on dividers and boxes in the Umbraco app is updated with many more examples to demonstrate and test this support:

Examples of boxes and lines in the example app

There's also an important bug fix - one third width columns were being rendered as one quarter wide.


bentpr (approved):

JodiB-TPR:
Howdy, Rick. The only ones that **weren't** designed at **4px** were the Details component's vertical bar and that of the Error state of a field / fieldset. The rest (e.g., Error summary, errored input field border, inner border of input text fields, radios and checkboxes when in focus state) were designed at 4px **because** of the bordered box on TPR's site. That is what was designed when the site was rebranded in 2018/19 so that's what I went with for consistency. The only reason Details and the Error state bars are 5px is because they weren't something we had on TPR, were taken directly from GOV, and they're not complete borders like the rest.

Here's what's been designed and used/documented in any prototypes I've done:

image

(Yellow Focus on input fields is 3px and inner border is 4px per GOV's. TPR Error field border is 4px because red alone is used on GOV (i.e., they keep the border 2px in error state) to denote the field error, so we made the field's border fatter to help distinguish it for users with protanopia as an extra affordance.)


image



image

(Yellow Focus and inner border on radios and checkboxes are 4px per GOV's)



So, as you can see, there's even inconsistency on GOV's! :)

Issue #359, opened 2024-11-19
Remove obsolete methods in v7.0.0


sussexrick:
The following static methods have been made obsolete and should be removed in the next major version. They were only ever intended for internal use, but were public types making this a breaking change.

`GovUkFieldsetErrorFinder.FindFieldsetErrors()`
`GovUkGridClassBuilder.BuildGridRowClass()`
`GovUkGridClassBuilder..BuildGridColumnClass()`

PR #360, opened 2024-11-20
Add tests for block rendering

sussexrick:
## Objectives of this PR

- We have important code that decides when and how to render row and column elements, which should be covered by tests but isn't.
- To avoid later duplication of code by making `BlockGrid.cshtml` as small as possible. Upcoming work will require a TPR-specific version of `BlockGrid.cshtml`.
- No change to behaviour.

## Implementation

- Move the logic that decides what to render out of the views and into a new `BlockViewService`, and add tests.
- Make `GovUkGridClassBuilder` and `GovUkFieldsetErrorFinder` injectable rather than static classes so that they can be injected into `BlockViewService`. Make the static methods obsolete but retain them to make this a non-breaking change. Created #359 for later removal of these methods.
- Added support for grid areas to our testing library.
- Added support for non-nullable bool to our testing library, because `.Value<bool>("field")` was not working, only `.Value<bool?>("field")` worked.


GovUk.Frontend.Umbraco/Blocks/BlockViewService.cs
```diff
@@ -0,0 +1,183 @@
...
+            // If this block is a fieldset and there is a fieldset-level error, add extra classes to show that the entire fieldset is in an error state.
+            // But only if the 'legendIsPageHeading' setting is false, otherwise it's done in GovUkFieldset.cshtml.
+            var fieldsetErrors = _fieldsetErrorFinder.FindErrors(block, modelState);
+            string? fieldsetErrorClasses = null;
+            if (fieldsetErrors.Any())
+            {
+                var legendIsPageHeading = block.Settings?.Value<bool>(PropertyAliases.FieldsetLegendIsPageHeading) ?? false;
+                if (!legendIsPageHeading)
+                {
+                    fieldsetErrorClasses = "govuk-form-group govuk-form-group--error";
```
villarin: Text used in multiple files. Maybe worth using a constant here?

sussexrick: Done

villarin (approved):
Code looks good and all tests passing. 
A bug related to givuk styles crasing the block grid rendering pipeline will be addressed on a separate issue/ticket

PR #363, opened 2024-12-05
Feature/203745 add breadcrumbs govuk styles

villarin:
**Adding Gov UK Breadcrumbs support**

Enable users understand and move between the multiple levels of a website.

Include in layout for content sections with breadcrumb trail as below:

`<govuk-breadcrumbs collapse-on-mobile="true">`
`<govuk-breadcrumbs-item href="/">Home</govuk-breadcrumbs-item>`
`<govuk-breadcrumbs-item href="/about">About</govuk-breadcrumbs-item>`
`<govuk-breadcrumbs-item >About - Mission</govuk-breadcrumbs-item>`
`</govuk-breadcrumbs>`


ZebSadiq (dismissed):
Looks ok to me.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukbreadcrumbs.config
```diff
@@ -0,0 +1,52 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
sussexrick: Supporting the breadcrumbs component as a block is the wrong approach. 

Blocks are for content that appear in the middle of the page, but the breadcrumbs component will appear at the top, above the content and above the block list/block grid. The [guidance on the GOV.UK page for the component](https://design-system.service.gov.uk/components/breadcrumbs/#how-it-works) says "Always place breadcrumbs at the top of a page, before the `<main>` element."

The `<govuk-breadcrumbs>` tag helper would be used on the layout page, not as a block. If you look at `_Views\Shared\_Layout.cshtml` you'll see `<govuk-back-link>` being used in this way already.

It won't be relevant now, but for when you're creating future components that are blocks, we always have a settings type for each component, and the settings go on there. You've added the `CSS classes` composition and the 'Is dark background?' setting on the content document type here.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/GOVUKGridColumnBlockList.config
```diff
@@ -295,6 +295,18 @@
       "stylesheet": null,
       "thumbnail": null,
       "view": null
+    },
```
sussexrick: It won't be relevant here now as this should not be implemented as a block, but for future components this was the wrong block list to use. It should be removed from here.

This one controls which components can be added in a grid column block, which sits inside a grid row block, which is used on a block list. All of that is deprecated in favour of using block grid, which did not exist when those block types were created.

Generally for a new component I'll look at all the existing block list and block grid data types, understand where those data types are used, and consider whether the new component would be appropriate there. For example, in a summary list block list, only summary list item blocks are appropriate. If the component you're adding is a TPR-specific one, it's not appropriate to add it to any GOV.UK block lists or block grids.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukbreadcrumbs.config
```diff
@@ -0,0 +1,52 @@
...
+    </Compositions>
+    <DefaultTemplate></DefaultTemplate>
+    <AllowedTemplates />
+  </Info>
+  <Structure />
+  <GenericProperties>
+    <GenericProperty>
+      <Key>c615169e-0405-4c00-a9c5-eee3b3d3a747</Key>
+      <Name>Is Dark Background?</Name>
+      <Alias>isDarkBackground</Alias>
```
sussexrick: You don't need to bother with dark background support, as we don't need it. Since any developer consuming our package would be expected to put `<govuk-breadcrumbs>` on their layout page they can apply the `govuk-breadcrumbs--inverse` class themselves.

To note for future implementations though, our property names are in "Sentence case" not "Title Case", as are our property descriptions. Our property descriptions always finish with a fullstop (or other punctuation), and we don't include a description at all if it doesn't add anything that isn't said in the name.

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/_breadcrumbs.cshtml
```diff
@@ -0,0 +1,25 @@
...
+@model BreadcrumbsViewModel;
+@addTagHelper *, GovUk.Frontend.AspNetCore
+
+@inject IPublishedUrlProvider PublishedUrlProvider
+@inject IUmbracoPublishedContentAccessor PublishedContext
+@{
+    var currentPage = PublishedContext.PublishedContent;
+    var selection = currentPage.Ancestors();
+    if (!selection.Any()) return;
+    var cssClasses = Model != null && Model.IsDarkBackground
```
sussexrick: You don't need to bother with dark background support, as we don't need it. I don't think you'll need this code either, but `Model.CssClasses` was only going to get applied on a dark background.

GovUk.Frontend.Umbraco.ExampleApp/wwwroot/css/site.css
```diff
@@ -3,7 +3,7 @@
 
 .govuk-grid-row--dark-background {
     background: gray;
-    padding-top: 15px;
+    padding: 15px;
```
sussexrick: This class is in use on the Buttons page, and you've added unwanted extra padding to the example there, so this should be reset.

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/_breadcrumbs.cshtml
```diff
@@ -0,0 +1,25 @@
+﻿@using GovUk.Frontend.Umbraco.Models
+@using ThePensionsRegulator.Umbraco
+@using Umbraco.Cms.Core.Routing
+@using Umbraco.Extensions
+
+@model BreadcrumbsViewModel;
+@addTagHelper *, GovUk.Frontend.AspNetCore
+
+@inject IPublishedUrlProvider PublishedUrlProvider
```
sussexrick: You may still need to inject these when you move this code to `_Layout.cshtml`. When you do, the injected variables should follow the `camelCase` naming convention for variables.

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/_breadcrumbs.cshtml
```diff
@@ -0,0 +1,25 @@
...
+    var selection = currentPage.Ancestors();
+    if (!selection.Any()) return;
+    var cssClasses = Model != null && Model.IsDarkBackground 
+        ? $"{Model.CssClasses} govuk-breadcrumbs--inverse" 
+        : string.Empty;
+}
+<govuk-breadcrumbs collapse-on-mobile="true" class="@cssClasses">
+    @foreach (var item in selection.OrderBy(x => x.Level))
+    {
+        <govuk-breadcrumbs-item href="@item.Url(PublishedUrlProvider)" link-target="_blank">@item.Name</govuk-breadcrumbs-item>
```
sussexrick: There's no need to include a link target. We don't want these links to open in a new tab, as that's rarely the right thing to do - see [GOV.UK guidance on opening links in a new tab](https://design-system.service.gov.uk/styles/links/#opening-links-in-a-new-tab).

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/_breadcrumbs.cshtml
```diff
@@ -0,0 +1,25 @@
...
+@inject IUmbracoPublishedContentAccessor PublishedContext
+@{
+    var currentPage = PublishedContext.PublishedContent;
+    var selection = currentPage.Ancestors();
+    if (!selection.Any()) return;
+    var cssClasses = Model != null && Model.IsDarkBackground 
+        ? $"{Model.CssClasses} govuk-breadcrumbs--inverse" 
+        : string.Empty;
+}
+<govuk-breadcrumbs collapse-on-mobile="true" class="@cssClasses">
```
sussexrick: This is still rendering a `<div>` rather than `<nav>`. We spoke about logging an issue in the `govuk-frontend-aspnetcore` library we consume, but actually this change only came in with [GOV.UK Frontend 5.4.1](https://github.com/alphagov/govuk-frontend/releases/tag/v5.4.1) and we are on GOV.UK Frontend 5.2.0.

I would suggest that we leave this for now, awaiting the implementation of GOV.UK Frontend 5.4.1 or better in `govuk-frontend-aspnetcore`.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/home.config
```diff
@@ -40,7 +40,7 @@
...
... draft specification)</li>\n<li><a href=\"/{localLink:umb://document/a8cfa09981654cfeb3d1179472290d0c}\">textarea</a></li>\n<li><a href=\"/{localLink:umb://document/acbabf457b3e4b53a825f7b6185f8840}\">text input</a> (demonstrates localisation)</li>\n<li><a href=\"/{localLink:umb://document/0e8a5a377a3341ff83867099db8db229}\">warning text</a></li>\n</ul>\n<p>You can <a href=\"/{localLink:umb://document/dd8e62c73e924bf28803ec2d080626e8}\">filter and override blocks</a> from code.</p>\n<p>Outside of the block list we also implement:</p>\n<ul>\n<li><a href=\"https://design-system.service.gov.uk/components/phase-banner/\">phase banner</a></li>\n<li>back to top (<a href=\"https://github.com/alphagov/govuk-design-system-backlog/issues/5\">proposed for the GOV.UK Design System</a>, implemented to a TPR design on every page)</li>\n</ul>\n<p>This example application also includes some TPR brand components:</p>\n<ul>\n<li>Back to menu (at the bottom of this page)</li>\n<li><a href=\"/{localLink:umb://document/868203ebff42473989acd6297ab8bc4a}\">Divider lines and boxes</a></li>\n<li><a href=\"/{localLink:umb://document/bd2933539783471ea3a238b761102d14}\">TPR header and footer</a></li>\n</ul>",
```
sussexrick: The link text going to 'breadcrumbs' needs to start with a lowercase b. All the other links to components are lowercase.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/breadcrumbs.config
```diff
@@ -0,0 +1,139 @@
...
+        "settingsUdi": "umb://element/3b223663cdcc4f7d9543f6cb083b304f"
+      }
+    ]
+  },
+  "contentData": [
+    {
+      "contentTypeKey": "e0876897-88f4-4dcf-ae0a-ec5a37e3a185",
+      "udi": "umb://element/e3ad23e23b6f438ea04a43419ecaf5af",
+      "text": {
+        "markup": "<h2 class=\"govuk-heading-l\">Breadcumbs&nbsp;</h2>\n<p>Assist users understand and move between the multiple levels of a website.</p>\n<p>The breadcrumbs components is based on <a href=\"https://design-system.service.gov.uk/components/breadcrumbs/\">Gov.Uk design guidelines</a></p>",
```
sussexrick: When you create pages for components in the example app you need to:
- include a 'Caption' block, which you can copy from other pages
- include a 'Page heading' block, which you can copy from other pages. Copying it will automatically give you the line underneath the heading, which is missing on these pages. 
- you've added an h2 heading here which is the wrong level. You won't need this when you have 'Page heading'.
- 'Breadcrumbs' is misspelled in the title of each page.

If you're going to link back to the component page on the GOV.UK Design System website (we don't usually), then the link text should be the title of the page you're linking to.

Given that the 'Breadcrumbs' component should not be used as a block, the way you should include it on these pages is to:
* Go to `Views\Shared\_Layout.cshtml` and add a new `@RenderSection("navigation", required: false)`. 
* The back link should appear by default on most pages, as it does now, so [use IsSectionDefined]( https://stackoverflow.com/questions/6001929/default-content-for-section-in-razor) to make the existing code the default content for the section.
* On the template for your breadcrumbs pages use `@section navigation { <govuk-breadcrumbs>...</govuk-breadcrumbs> }` to swap the back link for a breadcrumb trail. No page should use both back link and breadcrumb trail together.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/contentpage.config
```diff
@@ -0,0 +1,30 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
sussexrick: Each document type to demonstrate a component is named after that component, so 'Content page' is the wrong name here; it should be 'Breadcrumbs'. We do this so that when we need to demonstrate code interacting with the component we can add a controller specifically for that page.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/contentpage.config
```diff
@@ -0,0 +1,30 @@
...
+    <IsListView>False</IsListView>
+    <Variations>Nothing</Variations>
+    <IsElement>false</IsElement>
+    <HistoryCleanup>
+      <PreventCleanup>False</PreventCleanup>
+      <KeepAllVersionsNewerThanDays></KeepAllVersionsNewerThanDays>
+      <KeepLatestVersionPerDayForDays></KeepLatestVersionPerDayForDays>
+    </HistoryCleanup>
+    <Compositions>
+      <Composition Key="e8386da2-bf9c-43e2-ac0e-02f33f6b6c2d">govukGridColumn</Composition>
```
sussexrick: Our example pages add a `blocks` property with a block list or block grid directly on the document type, they don't add this property via a composition. As it happens the document type you chose is not meant to be a composition, and it has the wrong block list data type on it. You should be using 'GOV.UK Block list' or 'GOV.UK Block grid' for an example page.

sussexrick (changes requested):
You also need to add a link in README.md to [Breadcrumbs](https://github.com/gunndabad/govuk-frontend-aspnetcore/blob/main/docs/components/breadcrumbs.md) just after the link to 'Back link'.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukbreadcrumbs.config
villarin: Agree and removed now.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/GOVUKGridColumnBlockList.config
villarin: reverted.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukbreadcrumbs.config
villarin: Removed properties as component will be standalone and for inclusion in layouts.

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/_breadcrumbs.cshtml
villarin: removed now

GovUk.Frontend.Umbraco.ExampleApp/wwwroot/css/site.css
villarin: reverted

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/_breadcrumbs.cshtml
villarin: changed vars to camelCase now

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/_breadcrumbs.cshtml
villarin: the art of copy paste ;)  removed now

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/_breadcrumbs.cshtml
villarin: agree. maybe we should trial  upgrading to 5.4.1 at some point.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/home.config
villarin: done

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/breadcrumbs.config
villarin: Thanks for the feedback! All actioned now

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/contentpage.config
villarin: Sorted! Although I would like to discuss why not having a multi component page to try out variation of components and their impact on the block grid.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/contentpage.config
villarin: done

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/_breadcrumbs.cshtml
sussexrick: We're relying on a package that's targeting 5.2. 

To upgrade to 5.4.1 we'd need to submit PRs to that library to update it first.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/contentpage.config
sussexrick: That's fine. We do that for dividers and lines for example. We can create others if it's helpful.

sussexrick:
@villarin This just needs `Views\Shared\_breadcrumbs.cshtml` to be renamed `Views\Shared\GOVUK\GovUkBreadcrumbs.cshtml` and then it's good to go.

villarin:
> @villarin This just needs `Views\Shared\_breadcrumbs.cshtml` to be renamed `Views\Shared\GOVUK\GovUkBreadcrumbs.cshtml` and then it's good to go.

Perfect! All done now.

sussexrick (approved):

PR #364, opened 2024-12-13
Support full-width boxes

sussexrick:
## Objective of this PR

TPR applications need support for full-width boxes that break out of the limited width page wrapper defined by the class `govuk-width-container`. An example of the layout in use can be found on https://www.thepensionsregulator.gov.uk/en/employers, where a background extends the full width of the browser:

full width box on Employers page on TPR website

## Approach

To achieve this we need to render  `<div class="govuk-width-container">` around the components that should be limited in width, and not around full-width components.

 `<div class="govuk-width-container">` has hitherto been included on layout views inherited by every page, and most TPR applications don't need this support, so to require the width container to be removed would be a disruptive breaking change, and increase the size of rendered pages with many extra `<div>` elements.

Instead this PR creates a new opt-in setting that leaves the existing behaviour as default, so that this change is not disruptive to existing applications.

Since it is part of the GOV.UK grid system changing the rendering of the width container is added as a feature at GOV.UK level, but currently used only to implement the rendering of full-width boxes for TPR.

## Implementation

- Moved a lot of logic from existing views that render individual blocks into a service. This logic was not previously covered by unit tests, but now it is (some of this work was in #360). This has simplified the view to the point that a single view can be shared by blocks in a block list and a block grid.
- When the new `RenderWidthContainerForBlocks` setting is enabled, `<div class="govuk-width-container">`  is rendered around each block. This respects and extends the existing behaviour of grid rows, where consecutive blocks with the same configuration are rendered inside a single container rather than repeating the container for each block.
- A new extension point `IBlockViewInterceptor` is added and documented which allows consuming code to interrupt and change which container elements are rendered.
- This new extension point is used by `TprBoxViewInterceptor` to prevent a width container being rendered for a full-width TPR box.
- Components which render a child block list or block grid set a new `RenderWidthContainer = false` option to prevent nested width container elements being rendered inside width container elements.
- The previous block grid implementation of 'TPR box' is split into 'TPR box' and 'TPR nested box'. The first one can be added at the root of a block grid and includes the full-width option. The second can only be added within grid areas and does not include the full-width option. This prevents the full-width option being selected within a grid area where the box could not be rendered at full-width.
- TPR box components get a new choice of background colour, to reflect the different background colour used on some full-width boxes. The colour is not exactly the same as the current TPR website, but is agreed with the TPR design team and is a close match based on a standard TPR colour.

TPR box settings

## A new major version

This is mostly implemented as a non-breaking change. However, since the first part of the work was done in #360 and has now been released as v6.2.0, it is a breaking change from classes which are present in that version, though they are certainly not used in the wild.

This requires a version bump to v7.0.0, and I've taken advantage of this to make two more breaking changes:
- removed some static methods scheduled for removal in v7.0.0 which are now available as instance methods (see #359)
- moved the `RenderGrid` and `ChildColumnsDefaultToFullWidth` settings out of `ThePensionsRegulator.Umbraco` into `GovUk.Frontend.Umbraco` since they relate to implementing the GOV.UK grid system, and `ThePensionsRegulator.Umbraco` is supposed to be a core library that is just Umbraco and not even GOV.UK. These were only ever for internal use, so it's unlikely they're in use in the wild.

## Documentation updates

- updated documentation for 'TPR box' 
- documented the new `IBlockViewInterceptor` extension point
- documented how to support full-width content by changing the rendering of the width container
- updated/corrected documentation for starting a new application to reflect feedback from TPR developers, and changes that have come in since it was last reviewed
- updated the Umbraco example app to support full-width content and included an example of a full-width box

Example of full-width box


ThePensionsRegulator.Frontend.Umbraco/Services/TprBoxViewInterceptor.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            bool currentBlockIsFullWidthBox = BlockIsFullWidthBox(blockViewModel.CurrentBlock);
+
+            if (currentBlockIsFullWidthBox)
+            {
+                blockViewModel.OpenWidthContainer = false;
+                blockViewModel.CloseWidthContainer = false;
+            }
+
+            bool? previousBlockIsFullWidthBox = blockViewModel.PreviousBlock is not null ? BlockIsFullWidthBox(blockViewModel.PreviousBlock) : null;
```
villarin, line 30: I cannot see any nullable inputs here. So maybe worth simplifying this section using bool instead of bool? variables i.e. previousBlockIsFullWidthBox / nextBlockIsFullWidthBox?

villarin (commented):
Code looks good to me. I would only suggest the review of nullable bool variables that are not dependent on external inputs. Will do some local testing to check the full width boxes shortly.

villarin (approved):
- Code compiling and working locally as expected.
- Usync running succesfully
- Can see the different boxes behaviour including full width, solid and bordered. 

Some recommendations made on usage of bool vs bool? variables.

ThePensionsRegulator.Frontend.Umbraco/Services/TprBoxViewInterceptor.cs
sussexrick: The `null` is representing that the first block doesn't have a previous block, and the last block doesn't have a next block.

When two consecutive blocks are determined to be the same, in terms of their row and column requirements, they are merged into a single row/column. We can't use `bool` because then we'd have `false` instead of `null`, and `null != false` but `false == false`. The first block might be seen as the same as a previous one that doesn't exist, or the last might be the same as a next block that doesn't exist, and the rendering would be wrong.

PR #365, opened 2024-12-20
Enabling culture variation for shared components localisation

villarin:
**Enabling culture variation for shared components localisation**

**Objective**: Ensure shared components can be localised - specifically header, and footer assets.

Only text and rich text fields are in scope for this PR. Image, link and relation fields can be looked into on a case by case basis.


Directory.Build.props
```diff
@@ -1,5 +1,5 @@
 <Project>
 	<PropertyGroup>
-        <Version>7.0.0</Version>
+        <Version>7.1.0</Version>
```
sussexrick: 7.0.0 isn't released yet so you don't need to increment it

sussexrick (changes requested):
Looks good aside from the version number

Directory.Build.props
villarin: file update removed now

sussexrick (approved):

PR #366, opened 2024-12-20
#209217 support gov.uk section break

Ethan-Lambourne:
In this PR:
- `Section Break` and `Section Break Settings` element types created in the Example App;
- Created a partial view for the Section break that takes into account the visibility (`hidden`) and `size` variable settings;
- Used Example App to test Section break after adding them to `GOVUK Block List`, `GOVUK Block Grid`,  `TPR Block List`, and `TPR Block Grid`.
- Once all features were tested and considered working I copied the necessary config files over to the `GovUk.Frontend.Umbraco` and `ThePensionsRegulator.Frontend.Umbraco` projects respectively.

I've made a note that the `Corporate Website Block Grid` is to be modified to accept the new `GOVUK Section Break` Block and the current styles in the Corporate Websites `site.scss` file for the `contentbreak` will be removed. As these changes need to be put through on the `Websites-CorporateWebsite-Umbraco` repository.

[AB#209217](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/209217)


GovUk.Frontend.Umbraco/GovUk.Frontend.Umbraco.csproj
```diff
@@ -176,6 +176,11 @@
 		<Content Update="uSync\v9\ContentTypes\govukreadonly.config">
 		  <CopyToOutputDirectory>Never</CopyToOutputDirectory>
 		</Content>
+		<Content Update="uSync\v9\ContentTypes\govuksectionbreak.config">
+		  <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
sussexrick: These settings are wrong, and this will cause a build error in a consuming application. Note that all the files around it are set to `<CopyToOutputDirectory>Never</CopyToOutputDirectory>` so this one should be too. You can update this file directly, or select the file in Visual Studio and press F4 to get the Properties dialog and change it there.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govuksectionbreak.config
```diff
@@ -0,0 +1,25 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<ContentType Key="5d33cd38-b819-4afd-9d4a-499ee73799ab" Alias="govukSectionBreak" Level="2">
+  <Info>
+    <Name>Section break</Name>
+    <Icon>icon-inactive-line color-black</Icon>
```
sussexrick: I can see why you chose `icon-inactive-line` but because it's diagonal it doesn't really suggest the purpose of the component. I think `icon-remove` would be more appropriate, which is a horizontal line (albeit a short and wide one).

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/GOVUKSectionBreakSizeRadioButtonList.config
```diff
@@ -0,0 +1,24 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<DataType Key="223a3ac0-b842-42c9-abe0-6dd1fc151187" Alias="GOVUKSectionBreakSizeRadioButtonList" Level="1">
+  <Info>
+    <Name>GOVUKSectionBreakSizeRadioButtonList</Name>
```
sussexrick: This should be a plain English name with spaces, "GOV.UK Section break size", and it should be in the DataTypes/GOV.UK folder in the Umbraco backoffice (Settings > Data Types > right-click your data type > Move).

I appreciate this wasn't in the ticket you picked up but there is actually a fourth size, 'Small', which is what happens when no other size is specified. See [Section break on the GOV.UK design system](https://design-system.service.gov.uk/styles/section-break/).

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/GovUkSectionBreak.cshtml
```diff
@@ -0,0 +1,28 @@
...
+@using Microsoft.AspNetCore.Mvc.ModelBinding;
+@using ThePensionsRegulator.Umbraco.Blocks;
+@using ThePensionsRegulator.Umbraco;
+@using Umbraco.Extensions
+@{
+    var hidden = Model.Settings.Value("hidden");
+    var size = Model.Settings.Value("size");
+}
+
+@if (hidden is false)
```
sussexrick: This view isn't doing the right thing in a few ways. I think you need to start by reading:

* [GOV.UK guidance for the section break component](https://design-system.service.gov.uk/styles/section-break/)
* [Horizontal line tag](https://expertbeacon.com/html-horizontal-line-tag-a-complete-guide-for-developers/)

Firstly, you should be using an `<hr />` element instead of a `<div></div>` because it has a sematic meaning that's important for accessibility.

Secondly, setting "hidden" shouldn't prevent the component from rendering - you could just delete it if you wanted that. Rather it should render the `<hr />` element without the `govuk-section-break--visible` class, so that it creates space without rendering a line, and importantly still creates the thematic break for accessibility.

Thirdly, if this view were to render an `<hr />` without  `govuk-section-break--visible`, then the line would still be visible. You need to apply an additional class `govuk-section-break` every time, regardless of other settings.

Finally, I appreciate this wasn't in the ticket you picked up but there is actually a fourth size, 'Small', which is what happens when none out of medium, large and extra-large is selected. See [Section break on the GOV.UK design system](https://design-system.service.gov.uk/styles/section-break/).

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govuksectionbreaksettings.config
```diff
@@ -0,0 +1,66 @@
...
+  <GenericProperties>
+    <GenericProperty>
+      <Key>0e8f7ed9-427b-44c7-9e55-f08771123dbb</Key>
+      <Name>Hidden</Name>
+      <Alias>hidden</Alias>
+      <Definition>92897bc6-a5f3-4ffe-ae27-f2e7e33dda49</Definition>
+      <Type>Umbraco.TrueFalse</Type>
+      <Mandatory>false</Mandatory>
+      <Validation></Validation>
+      <Description><![CDATA[Toggle visibility on/off, not hidden by default.]]></Description>
```
sussexrick: We only add a description where it tells you something that isn't in the name, so I don't think you need this.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govuksectionbreak.config
```diff
@@ -0,0 +1,25 @@
...
+    <Thumbnail>folder.png</Thumbnail>
+    <Description>A thin grey horizontal line used to break up sections of text</Description>
```
sussexrick: I'd like the description for this component to explain the semantic meaning of it for accessibility (I've linked to more about that from another comment). A good description would be "A thematic break between sections of content. DO NOT use a section break for purely aesthetic reasons."

sussexrick (changes requested):
This is an excellent first effort Ethan. You've understood how to add a component to this repo and where to put all the files, and that's not easy because there are a lot of places to update.

Most components have a backoffice preview and you haven't added one, which I think is a good choice. In this case it's more helpful not to have one because the component's appearance is invisible or just a line that looks like a lot of other lines that already appear in the backoffice.

In addition to the comments on specific files there are a couple more things to look at:

1. When the section break line is shown and TPR styles are in use it is a thin grey line, yes, but it's not a TPR thin grey line. The colour is different. 

      The ticket you picked up talks about using [the `tpr-divider` mixin](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/ThePensionsRegulator.Frontend/Styles/_tpr-grid.scss) but I've had a look at that and its not going to be helpful because it adds spacing that would conflict with your size options. 
      
      However, in that same file `_tpr-grid.scss` you should be specifying that `.govuk-section-break` has `border-color: $tpr-colour-very-light-grey`.

2. You need to document what you've done. In README.md add it to the list of components we support, configured entirely in Umbraco, by adding a link to the [GOV.UK Section break page](https://design-system.service.gov.uk/styles/section-break/).

    In the Umbraco example app you need to add examples of the different settings. I would suggest you add them to the 'Divider lines and boxes' page.

Ethan-Lambourne:
@sussexrick I believe I have acted upon all of the comments above in my latest commits

README.md
```diff
@@ -18,6 +18,7 @@ We add support for:
   - [Summary card](/docs/components/summary-card.md)
   - [Summary list](/docs/components/summary-list.md)
   - [Warning text](https://github.com/gunndabad/govuk-frontend-aspnetcore/blob/main/docs/components/warning-text.md)
+  - [Section break](https://design-system.service.gov.uk/styles/section-break/)
```
sussexrick: This is not in alphabetical order

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govuksectionbreaksettings.config
sussexrick: You've still left a description in here which isn't adding anything, so remove it. Users can see that the default is hidden because the hidden toggle is visibly turned off.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govuksectionbreak.config
sussexrick: The settings document type should also have a description - see any other settings type in the GOV.UK folder. It's normally very similar so I would recommend "Settings for a thematic break between sections of content."

GovUk.Frontend.Umbraco/GovUk.Frontend.Umbraco.csproj
sussexrick: You've still got settings which are different to other files, so I would suggest tidying that up although they do work. Of more concern is that your other two new uSync files (`govuksectionbreaksettings.config` and `GOVUKSectionBreakSize.config`) are still being copied to the output directory. You need to change all of them to be `<CopyToOutputDirectory>Never</CopyToOutputDirectory>`.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govuksectionbreaksettings.config
```diff
@@ -0,0 +1,66 @@
...
+    </GenericProperty>
+    <GenericProperty>
+      <Key>d8ea0c24-695f-47fb-9aaa-11a145c1dd2e</Key>
+      <Name>Size</Name>
+      <Alias>size</Alias>
+      <Definition>223a3ac0-b842-42c9-abe0-6dd1fc151187</Definition>
+      <Type>Umbraco.RadioButtonList</Type>
+      <Mandatory>false</Mandatory>
+      <Validation></Validation>
+      <Description><![CDATA[Defaults to small]]></Description>
```
sussexrick: To be consistent with other properties that have defaults, this should be `Defaults to 'Small' if left blank.`

However, setting the default to 'Large' would make more sense. This would mean changing the `GovUkSectionBreak.cshtml` view as well as the text here.

The reason 'Large' would make more sense is because the default should be whatever editors are likely to use the most. Since this is a thematic break between sections on a page, doing that without any space is going to be an unusual requirement, so they'd have to change the setting almost every time. They're normally going to want some space around the dividing line. The current TPR website has an equivalent style which this will replace. It's called `.contentbreak` and it adds 30px of space, the same as the 'Large' setting here.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/TPRBlockGrid.config
```diff
@@ -1248,6 +1248,27 @@
...
+      "areas": [],
+      "backgroundColor": null,
+      "columnSpanOptions": [],
+      "contentElementTypeKey": "5d33cd38-b819-4afd-9d4a-499ee73799ab",
+      "editorSize": "medium",
+      "forceHideContentEditorInOverlay": false,
+      "groupKey": "fe4459ab-dada-4c2f-97f2-1b70aee27310",
+      "iconColor": null,
+      "inlineEditing": false,
+      "label": "",
```
sussexrick: In each example where you add this to the configuration for a block list or block grid data type you could add the following label: `Section break ({{$settings.size}}{{$settings.hidden ==  1 ? ', hidden' : ''}})`.

image

This means that the backoffice view will change from this:

image

To this, which is more helpful:

image

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govuksectionbreaksettings.config
```diff
@@ -0,0 +1,66 @@
...
+    <IsListView>False</IsListView>
+    <Variations>Nothing</Variations>
+    <IsElement>true</IsElement>
+    <HistoryCleanup>
+      <PreventCleanup>False</PreventCleanup>
+      <KeepAllVersionsNewerThanDays></KeepAllVersionsNewerThanDays>
+      <KeepLatestVersionPerDayForDays></KeepLatestVersionPerDayForDays>
+    </HistoryCleanup>
+    <Folder>GOV.UK</Folder>
+    <Compositions />
```
sussexrick: You need to add three compositions here (which you do by clicking the 'Compositions' button in the UI for this document type):

1. `Grid (row classes)` which automatically adds support for adding a custom CSS class on the grid row.
2. `Grid (column classes)` which automatically adds support for adding a custom CSS class on the grid column, and importantly allows changing the section break from its default size of two-thirds wide to another size (eg full width) if needed.
3. `CSS classes` which should be on every component and allows adding a custom CSS class to the component, in this case the `hr` element. You'll need to implement yourself in `GovUkSectionBreak.cshtml` but you can find examples in most files in the same folder.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/divider-lines-and-boxes.config
sussexrick, line 1: I like what you've done with the example page, adding every possible variation. There are a few changes needed though:

1. The 'Section breaks' h1 should be an h2, just because it's a sub-section of the page and not the main page heading.
2. For a similar reason where you have "Small:", "Medium:", "Large:" and "Extra large:" these are acting as headings for the content to follow so they should be proper headings, for both sets of these headings. As sub-sections of 'Section breaks' which is an h2, they should be h3. Remove the colon from each one as headings don't have that.
3. You say "to break up headings or sections of content" but that isn't clear _why_ you're breaking them up. Wherever possible in this app we re-use text from the GOV.UK Design System itself, so in this case it should say "to create a thematic break between sections of content."
4. You've got "sections breaks" when you meant "section breaks".
5. "the larger the size..." should be a new sentence.
6. "effects" should be "affects" and "Allowing you to..." isn't the start of a sentence - I'd suggest you deal with both of these and make it snappier by going with "A 'Hidden' toggle allows you to include any size of section break spacing without the grey line."

sussexrick (changes requested):

Ethan-Lambourne:
@sussexrick I have acted upon all of your current comments in the latest commit `97e0786`

sussexrick (dismissed):

sussexrick (approved):

PR #367, opened 2024-12-27
Automate the divider below h1 headings on TPR pages

sussexrick:
## Objective of this PR

When you create a TPR page you must always remember to apply a class to the grid row that represents the page heading, otherwise a required decorative line is missing. This is complicated by different classes being required where a form component or fieldset is setting the page heading.

This wastes business analyst and developer time and is also frequently forgotten, which wastes testers' time and creates extra work items to resolve it, which restarts the cycle and wastes more time.

This PR automates adding the class so that we can avoid this extra work.

## Implementation

An `IBlockViewInterceptor` implementation called `TprDividerViewInterceptor` looks for the right kinds of block and applies the relevant class just before the block is rendered. It ensures that the block is rendered in its own grid row so that there is an element to add the class to. 

It does not do this for layout blocks in a block grid because they take care of their own grid rendering, and letting the code alter the rendering of these blocks leads to invalid HTML and unexpected results.

The substantive change in this PR is really just the `TprDividerViewInterceptor` class and its tests:

- The uSync updates are just the manually-specified class being removed from example app pages because it's no longer needed. Some of these pages haven't been saved in a while, so they also serialise differently due to improvements in uSync itself.
- The Razor file updates are because `HtmlClassNames` is renamed to `GovUkClassNames` and made public so that TPR code can refer to it. 

## Other minor bugs fixed

- `ElementTypeAliases.cs` should have all element type aliases but some were missing, so I added them
- the text on the breadcrumb trail example page is updated to match that on the GOV.UK design system website
- in the example app the breadcrumb trail example was not correctly aligned, due to changes that came in for supporting full-width boxes
- the filename did not match the class for `TprBoxViewInterceptorTests`


ThePensionsRegulator.Frontend.Umbraco.Tests/Services/TprDividerViewInterceptorTests.cs
```diff
@@ -0,0 +1,220 @@
...
+				Mock.Get(blockViewModel.NextBlock.Settings).SetupUmbracoBooleanPropertyValue(isPageHeadingProperty, true);
+			}
+
+			var interceptor = new TprDividerViewInterceptor();
+
+			// Act
+			interceptor.InterceptBlockView(blockViewModel);
+
+			// Assert
+			Assert.That(blockViewModel.CloseGridRowAndColumn, Is.EqualTo(contentAlias == GovUkElementTypeAliases.PageHeading || hasLegendAsHeading || hasLabelAsHeading));
```
ZebSadiq: It would be better if the Asserts did not contain logic e.g. contentAlias == GovUkElementTypeAliases.PageHeading || hasLegendAsHeading || hasLabelAsHeading , but rather a pre-calculated expected value fed in as a parameter to the test. This will help a) write the test cases like a blackbox, separating your knowledge of the code from the business requirements b) encourage you to write more negative tests c) can help test and find bugs in the unit test itself.

ZebSadiq (dismissed):
Generally the code all seems fine. I've made a suggestion on the unit tests but am still approving the changes.

ThePensionsRegulator.Frontend.Umbraco.Tests/Services/TprDividerViewInterceptorTests.cs
sussexrick: @ZebSadiq I've already got some negative tests, but I agree with your reasons so I've updated the test code.

I've also updated how a layout block is identified to a more accurate definition. It was previously limited to well-known aliases but now looks for the presence of a grid area.

ZebSadiq (approved):

PR #375, opened 2025-02-04
207457 - Added Section Boxes feature

ZebSadiq:
Added Section Boxes named as Boxed Section. 
Incremented the version to 7.2.0.


villarin (changes requested):
- Please add component documentation (MD file) and refer to it from main [Readme.md](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/README.md)
- Include link to boxed section page under the [home page](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/home.config)
- Some document types or data types may be missing in the commits - see frontend and backoffice view of local setup using this branch after usync execution.

**Frontend**
 
image

**Backoffice** 

image

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprboxedsectionsettings.config
```diff
@@ -0,0 +1,68 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<ContentType Key="0544d56d-431c-4ede-91b5-bf194b47c6e1" Alias="tprBoxedSectionSettings" Level="2">
+  <Info>
+    <Name>Boxed section settings</Name>
+    <Icon>icon-thumbnails color-black</Icon>
+    <Thumbnail>folder.png</Thumbnail>
+    <Description>Settings for a collapsible section within an accordion component.</Description>
```
sussexrick: This needs to be updated

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/TPRBoxedSectionCardTextarea.config
```diff
@@ -0,0 +1,12 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<DataType Key="62ffbe9f-18aa-4b9b-9de5-c0ccc6b1c751" Alias="TPR - Boxed Section Card  - Textarea" Level="1">
+  <Info>
+    <Name>TPR - Boxed Section Card  - Textarea</Name>
```
sussexrick: I would suggest using rich text for this. You could add a rich text data type with just a few relevant options. I think it's highly likely they'll want to be able to use bold, for example, but a bullet or numbered list would be too much.

ThePensionsRegulator.Frontend.Umbraco/ThePensionsRegulator.Frontend.Umbraco.csproj
```diff
@@ -150,6 +150,10 @@
 		<Content Update="uSync\v9\DataTypes\TPRRichTextEditorHeaderAndFooter.config">
 		  <CopyToOutputDirectory>Never</CopyToOutputDirectory>
 		</Content>
+		<Content Update="Views\Shared\TPR\TPRBoxedSection.cshtml">
```
sussexrick: This is the only cshtml mentioned in the csproj file - probably not needed

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRBoxedSection.cshtml
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    if (string.IsNullOrWhiteSpace(descFieldName))
+    {
+        descFieldName = "description";
+    }
+
+    var cssClasses = (" " + Model.Settings.Value<string>(GovUkPropertyAliases.CssClasses)).TrimEnd();
+
+    <ul class="tpr-boxedsection-container govuk-body @cssClasses">
```
sussexrick: Lists should use `govuk-list`, not `govuk-body`: https://design-system.service.gov.uk/styles/lists/

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRBoxedSection.cshtml
```diff
@@ -0,0 +1,97 @@
...
+                        }
+                        var childName = child.Name;
+                        if(nameFieldName != "name")
+                        {
+                            childName = child.Value<string>(nameFieldName);
+                        }
+                        var childUrl = child.Url();
+                        var childDesc = child.Value<string>(descFieldName);
+                        <li class="tpr-boxedsection">
+                            <div class="tpr-boxedsection-body">
```
sussexrick: These class names should follow the BEM convention, to be in line with the GOV.UK Design System classes, so `tpr-boxedsection__body` for example.

ThePensionsRegulator.Frontend.Umbraco/uSync/v9/Content/boxed-section.config
```diff
@@ -0,0 +1,443 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Content Key="ba8e8ade-c4f2-4c58-9087-dcc5d79f168d" Alias="Boxed Section" Level="2">
+  <Info>
+    <Parent Key="5e682cbe-b867-491a-955f-3382446d5663">Home</Parent>
+    <Path>/Home/BoxedSection</Path>
+    <Trashed>false</Trashed>
+    <ContentType>boxedSection</ContentType>
+    <CreateDate>2025-01-15T11:39:29</CreateDate>
+    <NodeName Default="Boxed Section" />
```
sussexrick: This should be "Boxed section" to be in line with TPR house style (which is to use sentence case)

ThePensionsRegulator.Frontend.Umbraco/uSync/v9/ContentTypes/tprboxedsectionsettings.config
```diff
@@ -0,0 +1,68 @@
...
+  <GenericProperties>
+    <GenericProperty>
+      <Key>264192f7-0aaa-4210-bd94-def271ca3554</Key>
+      <Name>Description field name</Name>
+      <Alias>descriptionFieldName</Alias>
+      <Definition>0cc0eba1-9960-42c9-bf9b-60e150b429ae</Definition>
+      <Type>Umbraco.TextBox</Type>
+      <Mandatory>false</Mandatory>
+      <Validation></Validation>
+      <Description><![CDATA[Is used to determine which field to pull the text for the description of the card from.]]></Description>
```
sussexrick: This should explain the default using this phrase, to be consistent: "Defaults to 'name' if left blank." The one below would be "Defaults to 'description' if left blank."

I think it would make sense for the component to use the node name by default though, rather than requiring an extra name field. It's generally going to be what's needed, and you want it to stay up to date if the page title changes.

ThePensionsRegulator.Frontend/Styles/_tpr-boxed-section.scss
```diff
@@ -0,0 +1,61 @@
...
+    padding: 0;
+    box-sizing: border-box;
+    display: flex;
+
+    @include govuk-media-query($until:desktop) {
+        max-width: calc(100% / 2);
+    }
+
+    @include govuk-media-query($until:tablet) {
+        /*width: 100%;*/
```
sussexrick: comment can be removed

ThePensionsRegulator.Frontend/Styles/_tpr-boxed-section.scss
```diff
@@ -0,0 +1,61 @@
...
+        p {
+            margin-bottom: 0;
+        }
+    }
+}
+
+h2.tpr-boxedsection-title {
+    padding: 0 30px 30px;
+    margin: 0 -30px;
+    border-bottom: solid 1px #d6d6d6;
```
sussexrick: Other dividing lines are `$tpr-colour-very-light-grey` so that's probably the correct value here?

ThePensionsRegulator.Frontend/Styles/_tpr-boxed-section.scss
```diff
@@ -0,0 +1,61 @@
...
+
+        p {
+            margin-bottom: 0;
+        }
+    }
+}
+
+h2.tpr-boxedsection-title {
+    padding: 0 30px 30px;
+    margin: 0 -30px;
```
sussexrick: Most of these px widths should probably be using the `govuk-responsive-padding` and `govuk-responsive-margin` mixins to set larger values on big screens than mobiles, or `govuk-px-to-rem` so that the spacing adjusts with the base font size of the browser.

ThePensionsRegulator.Frontend/Styles/tpr.scss
```diff
@@ -76,4 +76,5 @@
 @import '_tpr-back-to-top.scss';
 @import '_tpr-add-link.scss';
 @import '_tpr-header.scss';
-@import '_tpr-footer.scss';
\ No newline at end of file
+@import '_tpr-footer.scss';
+@import '_tpr-boxed-section.scss';
```
sussexrick: I can see it hasn't been done in this last group so not surprised you didn't notice, but these imports are generally in alphabetical order to make lines easier to find.

sussexrick (dismissed):
I've reviewed as best I can without the example content.

I do wonder about the name. I remember there was a conversation but I don't remember the outcome. I remember though that "card" was the preferred term for the smaller boxes (which I recognise you are using), because it's called a card in other gov design systems. 

But that leaves us with a "Boxed section" which may not involve any boxes. 
- it contains cards, not boxes
- it may not be boxed because it's not responsible for the box outside it either, which is a separate 'Box' component that we can't guarantee will be used.

ThePensionsRegulator.Frontend.Umbraco/uSync/v9/Content/boxed-section-child-page-1.config
sussexrick, line 1: You don't want this content in the package, this can stay with the example app only.

ZebSadiq:
> I've reviewed as best I can without the example content.
> 
> I do wonder about the name. I remember there was a conversation but I don't remember the outcome. I remember though that "card" was the preferred term for the smaller boxes (which I recognise you are using), because it's called a card in other gov design systems.
> 
> But that leaves us with a "Boxed section" which may not involve any boxes.
> 
> * it contains cards, not boxes
> * it may not be boxed because it's not responsible for the box outside it either, which is a separate 'Box' component that we can't guarantee will be used.

@sussexrick 
I agree the the term Boxed would get confusing with Box (as there is already a component with that name). However, I didn't get a definitive name being decided upon in our recent discussion.

Using the reference to 'Card', could we call it:

- Carded section
- Carded navigation
- Section cards
- Card listing

The above are focused on naming the component how it looks. We could take a different approach and name it what this component is most likely going to be used for.  This type of components usually lists the major parts of a section, that's why we could call it:

- Section navigation
- Section cards
- Local navigation
- Key pages
- Section guide

Please let me know which one you are leaning towards.

Also, please let me know if you want just the name visible to editors to updated or do you want me to update the internal names and file names too...

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRBoxedSection.cshtml
ZebSadiq: Without the govuk-body being applied, the fonts and sizing is completely off brand. It seems that the the basic branding from the GDS are not being applied to the page as a whole, but rather to each component. If we can apply govuk-body to the actual body of the page, then we wouldn't need govuk-body on this component. Which would you prefer?

ThePensionsRegulator.Frontend.Umbraco/uSync/v9/ContentTypes/tprboxedsectionsettings.config
ZebSadiq: Sure, I will update the descriptions.

Right now it does default to Node name. The override functionality is so that we are able to solve the following problems:

- Long node name (I've seen this happen lots of times). The long name could increase the card size beyond reasonable. This would also push the boxes next to it to resize to stay symmetrical. Therefore, we could point the name to a shorter version such as a social media title field.
- A different title from the original that invites the user to click on the link without compromising on the full page title found on the actual page.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprboxedsectionsettings.config
ZebSadiq: Note to self: Update the description field

ThePensionsRegulator.Frontend.Umbraco/uSync/v9/Content/boxed-section.config
ZebSadiq: Note to self: Update the NodeName field

ThePensionsRegulator.Frontend.Umbraco/ThePensionsRegulator.Frontend.Umbraco.csproj
ZebSadiq: I believe this is required by the Grid when it renders it's blocks.  It renders the component as a block.

sussexrick: The same is true for several other files in the same folder, and they don't need this config in the project file. Better to be consistent to avoid any unforeseen side-effects.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRBoxedSection.cshtml
sussexrick: I'd prefer to follow the usage examples on the GOV.UK Design System website. In this case swapping `govuk-body` for `govuk-list` produces an almost-identical result.

sussexrick:
@ZebSadiq Let's go with "Section cards". I ran it past Jane in the content team and she's OK with that. And yes, it would need to be reflected in the code.

ThePensionsRegulator.Frontend.Umbraco/ThePensionsRegulator.Frontend.Umbraco.csproj
ZebSadiq: Now I understand what you mean. Will delete it.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRBoxedSection.cshtml
ZebSadiq: Ok that makes sense, have changed it to use govuk-list.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprboxedsectionsettings.config
ZebSadiq: Amended.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/TPRBoxedSectionCardTextarea.config
ZebSadiq: Amended.

ThePensionsRegulator.Frontend.Umbraco/ThePensionsRegulator.Frontend.Umbraco.csproj
ZebSadiq: Amended.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRBoxedSection.cshtml
ZebSadiq: Amended.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRBoxedSection.cshtml
ZebSadiq: Amended.

ThePensionsRegulator.Frontend.Umbraco/uSync/v9/ContentTypes/tprboxedsectionsettings.config
ZebSadiq: Amended.

ThePensionsRegulator.Frontend/Styles/_tpr-boxed-section.scss
ZebSadiq: Amended.

ThePensionsRegulator.Frontend/Styles/_tpr-boxed-section.scss
ZebSadiq: Amended.

ThePensionsRegulator.Frontend/Styles/_tpr-boxed-section.scss
ZebSadiq: Amended.

ThePensionsRegulator.Frontend/Styles/tpr.scss
ZebSadiq: Amended.

ThePensionsRegulator.Frontend.Umbraco/uSync/v9/Content/boxed-section-child-page-1.config
ZebSadiq: Deleted.

ZebSadiq:
> @ZebSadiq Let's go with "Section cards". I ran it past Jane in the content team and she's OK with that. And yes, it would need to be reflected in the code.

Amended.

ZebSadiq:
> * Please add component documentation (MD file) and refer to it from main [Readme.md](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/README.md)
> * Include link to boxed section page under the [home page](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/home.config)
> * Some document types or data types may be missing in the commits - see frontend and backoffice view of local setup using this branch after usync execution.
> 
> **Frontend**
> 
> image
> 
> **Backoffice**
> 
> image

Fixed and amended.

Directory.Build.props
```diff
@@ -1,5 +1,5 @@
 <Project>
 	<PropertyGroup>
-        <Version>7.1.0</Version>
+        <Version>7.2.0</Version>
```
villarin: Please increase to 7.3.0

villarin (dismissed):
All feedback actioned now. Happy to approve but please increase the package version

ZebSadiq:
> All feedback actioned now. Happy to approve but please increase the package version

done.

villarin (approved):

PR #376, opened 2025-02-07
Eliminate duplicate uSync files for easier packaging

sussexrick:
When maintaining Umbraco support in this repo our workflow has been:

1. make document and data type changes in the Umbraco example app
2. uSync export in the example app
3. copy the exported document types and data types to the package project, being careful to separate GOV.UK and TPR files
4. set the files not to copy to the output directory in package project
5. commit two copies of the file

In addition the example app would copy uSync files from the package at build time, sometimes overwriting the files you're working on with older versions.

This PR eases that workflow by including files in the packages directly from the example app. This approach:
- eliminates manual steps
- makes PR review easier without duplicate files
- speeds up the build of the example app by removing some file copy operations

It also updates documentation with a recommendation for how to test a change like this one that affects package contents.

Finally, it also removes uSync data for `GOVUKNoFormsBlockList.config` and `TPRNoFormsBlockList.config` from the example app. These are [obsolete and were removed in v6.0.0](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/releases/tag/v6.0.0), but have crept back in at some point.


villarin (approved):
@sussexrick 

- Conflicts resolved 
- Tested usync complete from empty database -passed
- Backoffice / Frontend components working as expected

PR #382, opened 2025-02-25
YouTube player block added with disabled AblePlayer

ZebSadiq:
Added a YouTube Player that can be added as a grid block or using the tag helper. The AblePlayer implementations have been commented out for now to the AblePlayer not working with some Content Security Policy and Cross Origin directives.


villarin (changes requested):
- Home page should link to youtube page - see highlighted area below

image

- There seems to be some missing data/doc types assets in the PR, because after usync I get the following error 
image

ThePensionsRegulator.Frontend/TagHelpers/TprYoutubeVideoTagHelper.cs
```diff
@@ -0,0 +1,139 @@
...
+       [HtmlAttributeName(UseAblePlayerAttributeName)]
+        public bool? UseAblePlayer 
+        {
+            get => _useAblePlayer;
+            set => _useAblePlayer = value; 
+        }
+
+
+        /// <inheritdoc/>
+        public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
```
villarin: Method is async but no operation is awaited in its body. Review if you need to await SomeMethod(params) or use  return Task.FromResult at the end

ZebSadiq:
> * Home page should link to youtube page - see highlighted area below
> 
> image
> 
> * There seems to be some missing data/doc types assets in the PR, because after usync I get the following error
>   image

Youtube component is listed under additional components. The top components are coming from GDS.

image


I've fixed the missing doc type changes. I've found that under some scenarios, uSync files can get overwritten and changes get lost. This is what's happened here.

ThePensionsRegulator.Frontend/TagHelpers/TprYoutubeVideoTagHelper.cs
ZebSadiq: Changed to non async method as it wasn't required.

Directory.Build.props
```diff
@@ -1,5 +1,5 @@
 <Project>
 	<PropertyGroup>
-        <Version>7.1.0</Version>
+        <Version>7.2.0</Version>
```
villarin: Increase to 7.2.1 or 7.3.0

ZebSadiq: Incremented.

ZebSadiq:
Hi @villarin 
I've made my amendments to the feedback. Please review again.

villarin (dismissed):
Player working as expected locally on both backoffice and frontend

ZebSadiq:
Version updated

villarin (dismissed):

villarin (approved):

PR #391, opened 2025-03-19
Bug/handle empty umbraco translations in date binding

BenHughesTPR:
I'd observed that the version of umbraco we're running with will return an empty string from the culture dictionary, rather than null, for keys that are not found.  This has a knock-on effect in the UmbracoDateInputModelBinder where we read from this dictionary and use the null coalesce operator to fall back onto a default error message if no configured message is found.  As a result we always end up getting an empty string as the error message.

This PR does the following:
- extends the test coverage around this area to cover off both default and configured error messages
- adds a new extension method for reading from the culture dictionary, and converting empty strings / whitespace into nulls.


GovUk.Frontend.Umbraco/ModelBinding/CultureDictionaryExtensions.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Umbraco.Cms.Core.Dictionary;
+
+namespace GovUk.Frontend.Umbraco.ModelBinding
+{
+    internal static class CultureDictionaryExtensions
+    {
+        public static string? ReadValue(this ICultureDictionary cultureDictionary, string key)
```
sussexrick: There is a built-in class that does this. If you inject `UmbracoHelper` you get access to `Umbraco.GetDictionaryValueOrDefault()` where you can pass the key as the first parameter and the default value `null` as the second parameter.

BenHughesTPR: Good shout, done - have wired in the UmbracoHelper and added an extra test to cover if it can't be accessed.  I've tested running the example app locally and it's working as expected.

sussexrick (approved):

PR #392, opened 2025-03-21
Section cards fix for opening links in new tab

ZebSadiq:
Now allows for the links to respect the 'Opens  the linked document in a new window or tab'

### DIFF
--- a/ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRSectionCards.cshtml
+++ b/ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRSectionCards.cshtml
@@ -51,9 +51,10 @@
                         continue;
                     var name = link.Name;
                     var url = link.Url;
+                    var target = link.Target;
                     var desc = box.Content.Value<string>(descFieldName);
                     <tpr-section-cards-card>                       
-                            <tpr-section-cards-card-title href="@url">
+                            <tpr-section-cards-card-title href="@url" target="@target">
                                  @name 
                             </tpr-section-cards-card-title>
                             @if (!string.IsNullOrWhiteSpace(desc))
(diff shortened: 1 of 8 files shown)


ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRSectionCards.cshtml
```diff
@@ -51,9 +51,10 @@
                         continue;
                     var name = link.Name;
                     var url = link.Url;
+                    var target = link.Target;
                     var desc = box.Content.Value<string>(descFieldName);
                     <tpr-section-cards-card>                       
-                            <tpr-section-cards-card-title href="@url">
+                            <tpr-section-cards-card-title href="@url" target="@target">
```
villarin, line 57: As target is nullable you probably need to set a default i.e. _self here to prevent an empty target attribute

ZebSadiq: Already thought of this and is addressed in ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSectionCards.cs line 33. We make sure that we check for nulls or empty strings. Tested in dev and works as expected.

villarin (approved):

sussexrick:
@ZebSadiq This support was deliberately excluded due to [GOV.UK advice on opening links in a new tab](https://design-system.service.gov.uk/styles/links/#opening-links-in-a-new-tab). Without enforcing it we are likely to get links set to open in new windows contrary to accessibility recommendations. Do we have a strong use case that overrides that?

## Opening links in a new tab
Avoid opening links in a new tab or window. It can be disorienting - and [can cause accessibility problems for people who cannot visually perceive that the new tab has opened](https://www.w3.org/TR/WCAG20-TECHS/G200.html).

If you need a link to open in a new tab, include the words ‘opens in new tab’ as part of the link. A use case example is to stop the user losing information they’ve entered into a form. There’s no need to tell the user the new tab opens in a new window as this is the default behaviour for most browsers.

PR #405, opened 2025-04-09
Feature/204473/Create-Tpr-Header-Search-Component

Max-Emery:
In this PR:
-Created TPR header search component, a sub component of TPR header bar. 
-Implemented the accessible-autocomplete component to support result generation into dropdown box
-Added examples of header search bar to 'TPRHeaderLockUp' partial view and TPR header and footer page


GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/headersearch.config
sussexrick, line 1: This file is from where you changed your mind and it doesn't need to be committed. In the Umbraco backoffice go to "Settings > uSync > Everything > Export > Clean export" to remove it.

GovUk.Frontend.Umbraco.ExampleApp/wwwroot/js/autocomplete.js
```diff
@@ -0,0 +1,28 @@
...
+document.addEventListener("DOMContentLoaded", function () {
+
+    const searchBars = document.querySelectorAll('.tpr-autocomplete-container');
+   
+    searchBars.forEach((searchBar) => {
+
+        accessibleAutocomplete({
+
+            element: searchBar,
+            id: 'easy-autocomplete',
```
sussexrick: I think an id like `tpr-header-search-autocomplete` would be more appropriate.

GovUk.Frontend.Umbraco.ExampleApp/wwwroot/js/autocomplete.js
```diff
@@ -0,0 +1,28 @@
...
+            source: function (query, populateResults) {
+                fetch('/data.json')
```
sussexrick: Hard-coding `fetch('/data.json')` means this file is only ever useful in the example app, but I think it could be useful to include it in the package along with the header search box component. For that to work your tag helper would need to support an attribute like `autocomplete-url` which is rendered in the HTML as `data-autocomplete-url`. This code could then look for that attribute on `searchBar` to get the URL.

GovUk.Frontend.Umbraco.ExampleApp/wwwroot/js/autocomplete.js
```diff
@@ -0,0 +1,28 @@
...
+                    .then(response => response.json())
+                    .then(data => {
+                        const results = data.filter(item => item.title.toLowerCase().includes(query.toLowerCase())).map(item => item.title);
+                        populateResults(results.slice(0,5))                 
+                    })
+                    .catch(error => {
+                        console.error('An error has occured with your fetch operation:', error);
```
sussexrick: It would be useful to say it's the fetch in the `tpr-header-search` component.

GovUk.Frontend.Umbraco.ExampleApp/wwwroot/js/autocomplete.js
```diff
@@ -0,0 +1,28 @@
...
+                    });
+        },
+            minLength: 2,
+            placeholder: 'Search'
```
sussexrick: This needs to be localisable (ie translatable into Welsh). You should be able to pick this value up from the `placeholder` attribute on the `input[type=search]` element.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -83,6 +85,76 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
...
+                    form.Attributes.Add("method", "get");
+
+                    var searchField = new TagBuilder("div");
+                    searchField.AddCssClass("searchFieldWithButton");
+
+                    var autoComplete = new TagBuilder("div");
+                    autoComplete.AddCssClass("tpr-autocomplete");
+                    searchField.InnerHtml.AppendHtml(autoComplete);
+
+                    var autoCompleteContainer = new TagBuilder("div");
```
sussexrick: This needs to be an `<input type="search" class="govuk-input" />` element (or there needs to be one in here). Currently your component is entirely reliant upon JavaScript, which isn't allowed in the Government Service Standard. Because it starts as a `div` the search box isn't even visible with JavaScript off. Autocomplete is always going to need JavaScript, but without JavaScript I would expect to see a fully-functional search box _without_ autocomplete.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -83,6 +85,76 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
                     contentElement.InnerHtml.Append(tprHeaderBar.Content.ToString()!);
                 }
             }
+            else
+            {
+                if (tprHeaderBar.DisplaySearchBar)
```
sussexrick: It would be neater to extract all of this search code into a separate method.

GovUk.Frontend.Umbraco.ExampleApp/Views/TPRHeader.cshtml
```diff
@@ -31,6 +34,8 @@
     }
     <title>@Model.PageHeadingOrName()</title>
     <link rel="stylesheet" href="~/css/site.css" />
+    <script src="~/js/accessible-autocomplete.min.js"></script>
```
sussexrick: These files are in the example app, but your component will be published in a NuGet package and consumed by other applications, and these files won't be there. You should be able to put them under `ThePensionsRegulator.Frontend/wwwroot/tpr` (`accessible-autocomplete.min.js` would go in a `wwwroot/tpr/lib/accessible-autocomplete` folder) - see https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/docs/contributing/include-client-side-files-in-packages.md

ThePensionsRegulator.Frontend/Models/TprHeaderLockupModel.cs
```diff
@@ -26,5 +26,6 @@ public class TprHeaderLockupModel
         public virtual string? Context1 { get; set; }
         public virtual string? Context2 { get; set; }
         public virtual string? Context3 { get; set; }
+        public virtual string? ActionPath { get; set; } = "en/search-result";
```
sussexrick: The default value `en/search-result` is specific to our public website but you're building a component that shouldn't know which site it'll be used on, so just remove this default value.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -83,6 +85,76 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
...
+                    searchField.AddCssClass("searchFieldWithButton");
+
+                    var autoComplete = new TagBuilder("div");
+                    autoComplete.AddCssClass("tpr-autocomplete");
+                    searchField.InnerHtml.AppendHtml(autoComplete);
+
+                    var autoCompleteContainer = new TagBuilder("div");
+                    autoCompleteContainer.AddCssClass("tpr-autocomplete-container");
+                    autoCompleteContainer.Attributes.Add("id", "eac-container-GlobalSearchInputBox");
+                    autoComplete.Attributes.Add("placeholder", "Search");
```
sussexrick: The placeholder text "Search" needs to be localisable (ie translatable) into Welsh. To do that, make the value come from a property on the tag helper. Then it's up to the consuming application to handle the translation and configure the desired text.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -83,6 +85,76 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
...
+                    autoComplete.InnerHtml.AppendHtml(autoCompleteContainer);
+
+                    var button = new TagBuilder("button");
+                    button.Attributes.Add("type", "submit");
+                    button.Attributes.Add("class", "searchButton");
+                    button.Attributes.Add("aria-label", "Search");
```
sussexrick: The ARIA label text "Search" needs to be localisable (ie translatable) into Welsh. To do that, make the value come from a property on the tag helper. Then it's up to the consuming application to handle the translation and configure the desired text.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -83,6 +85,76 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
...
+                    if (tprHeaderBar.SearchAttributes != null) { divTag.MergeAttributes(tprHeaderBar.SearchAttributes); }
+                    divTag.AddCssClass("search");
+
+                    var form = new TagBuilder("form");
+                    form.Attributes.Add("action", $"{tprHeaderBar.ActionPath}");
+                    form.Attributes.Add("id", "form-globalsearch");
+                    form.Attributes.Add("method", "get");
+
+                    var searchField = new TagBuilder("div");
+                    searchField.AddCssClass("searchFieldWithButton");
```
sussexrick: Your classes should use the [BEM](https://getbem.com/introduction/) naming pattern, eg `tpr-header-search`, `tpr-header-search__button` and so on. IDs should all be in the form `tpr-header-search-*`. 

This makes it clear where they apply to, and they shouldn't clash with anything else on the page, like another search in the middle of the page that uses autocomplete with a different data source.

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
```diff
@@ -0,0 +1,226 @@
+@import '_tpr-variables.scss';
+@import 'govuk/base';
+
+
+.search {
+    width: 30%;
+    padding: 0.5em 0.5em;
+    text-align: center;
+    display: flex;
+    flex: 0 0, auto;
```
sussexrick: There shouldn't be a comma in `flex: 0 0 auto`. You're specifying `text-align` and `flex-direction` twice.

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
```diff
@@ -0,0 +1,226 @@
...
+    align-items: center;
+}
+
+.searchFieldWithButton {
+    display: flex;
+    align-items: center;
+}
+
+.tpr-autocomplete {
+    width: 313.948px
```
sussexrick, line 21: This is very precise! What's the reason?

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
```diff
@@ -0,0 +1,226 @@
...
+    padding-left: 1.2rem;
+    padding-right: 2rem;
+    height: 3.1em;
+    border: none;
+    font-size: 0.9em;
+    margin: 0.3em;
+    vertical-align: middle;
+}
+
+.searchbox[type=text]:focus {
```
sussexrick: The search input has a black border that the current TPR website doesn't have.

When you tab to the search input and search button they should get a GOV.UK-style focus ring. Try tabbing to form controls on the other example pages to see what I mean.  You shouldn't need to define a custom focus style like this.

On the search input you can apply class `govuk-input` to get it. On the button I think you need the `govuk-focused-box` SASS mixin. See https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-focused-box

Also when the button is focussed it has an odd shape:

image

I found that if you set the `height` and `width` to `2rem` and change the `viewBox` attribute to `0 0 21 6` it makes it better.

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
```diff
@@ -0,0 +1,226 @@
...
+    outline-width: 0.19em;
+    outline-color: orange;
+    outline-style: solid;
+}
+
+svg {
```
sussexrick: This `svg` style is going to apply to any SVG image included in the page in the same way. You need to make it specific to the SVG in the header search component, eg by using a class like `tpr-header-search__button-image`.

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
```diff
@@ -0,0 +1,226 @@
...
+    box-sizing: border-box;
+    height: 2.5rem;
+    line-height: 1.25;
+    margin-bottom: 0;
+    width: 85%;
+    vertical-align: middle;
+}
+
+.autocomplete__input {
+    background-color: #fff;
```
sussexrick: All colours should use variables from the `_tpr-variables.scss` file. For example this one will be `$tpr-colour-white`. You may find available TPR styles are slightly different to what you have, eg a different shade of black. That's OK - go with TPR colours.

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
```diff
@@ -0,0 +1,226 @@
...
+    border: 1px solid #b1b4b6;
+    border-top: 0;
+    color: #0b0c0c;
+    margin: 0;
+    max-height: 342px;
+    overflow-x: hidden;
+    padding: 0;
+    width: 85%;
+    width: calc(100% - 4px);
+    font-family: "Segoe UI Symbol";
```
sussexrick: Use `font-family: $govuk-font-family` to fit in with fonts elsewhere.

GovUk.Frontend.Umbraco.ExampleApp/Views/TPRHeader.cshtml
```diff
@@ -181,6 +186,26 @@
             <tpr-context-bar>
                 <tpr-context-bar-context-1>@(settings?.Value<IHtmlEncodedString>("tprContext1"))</tpr-context-bar-context-1>
             </tpr-context-bar>
+            <div class="govuk-width-container govuk-!-margin-top-5">
+                <partial name="GOVUK/BlockList" model="Model.HeaderWithSearch" />
+            </div>
+            <tpr-header-bar>
+                <tpr-header-bar-logo alt="@headerLogoAlt" href="@(settings?.Value<Link>("tprHeaderLogoHref")?.Url)" />
+                <tpr-header-search action="/@languageTwoLetterCode/@actionpath"></tpr-header-search>
```
sussexrick: I like the examples you've added, but the search boxes do go to a 404 page. In this example app we usually send everything to `/success/`.

ThePensionsRegulator.Frontend/TagHelpers/TprHeaderSearchTagHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+        [HtmlAttributeName(ActionAttributeName)]
+        public string? ActionPath { get; set; }
+
+        public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
+        {
+            var headerSearchContext = context.GetContextItem<TprHeaderBarContext>();
+
+            var content = await output.GetChildContentAsync();
+
+            headerSearchContext.SetSearch(output.Attributes.ToAttributeDictionary(), true, ActionPath);
```
sussexrick: `ActionPath` may be null here, so you need to change the definition of `SetSearch` to accept `string?`

ThePensionsRegulator.Frontend/TagHelpers/TprHeaderBarContext.cs
```diff
@@ -19,6 +20,10 @@ internal class TprHeaderBarContext
         public AttributeDictionary? ContentAttributes => _content?.Attributes;
         public IHtmlContent? Content => _content?.Content;
         public bool ContentAllowHtml => _content?.AllowHtml ?? false;
+        public bool DisplaySearchBar => _search?.DispaySearchBar ?? false;
```
sussexrick: I don't think this is a "bar". It appears within the "header bar". I would call this `DisplaySearch`.

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
```diff
@@ -0,0 +1,226 @@
...
+    width: 85%;
+    vertical-align: middle;
+}
+
+.autocomplete__input {
+    background-color: #fff;
+    position: relative;
+}
+
+::placeholder {
```
sussexrick: This is going to affect every piece of placeholder text in every form element. This isn't the place to do something like that, so it should be specific to this `autocomplete_input::placeholder`. Ideally it wouldn't be `0.8em` which is smaller than the user's chosen font size, but see how `1rem` looks.

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
```diff
@@ -0,0 +1,226 @@
...
+svg {
+    border: none;
+    height: 3.87em;
+    width: 2em;
+    display: inline-block;
+    cursor: pointer;
+    vertical-align: bottom;
+}
+
+.autocomplete__wrapper {
```
sussexrick: All these `.autocomplete-*` styles should be in their own file because they don't only apply to the header search box. They will apply to any autocomplete used on the page.

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
```diff
@@ -0,0 +1,226 @@
...
+    .searchButton {
+        margin-left: -52px;
+    }
+
+    .autocomplete__menu--visible {
+        margin-left: 16px;
+    }
+}
+
+@media(max-width: 864px) {
```
sussexrick: I think this needs work. On the current public website the search is visible down to around 765px. Somewhere around 990px the header bar label "Making workplace pensions work" disappears to make way for the search to continue to show.

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
```diff
@@ -11,7 +11,13 @@
     var headerLogoAlt = Model.LogoAlternativeText;
     if (string.IsNullOrEmpty(headerLogoAlt) && !string.IsNullOrEmpty(Model.LogoHref)) { headerLogoAlt = ComponentGenerator.HeaderLogoDefaultAlt; }
     var showPhaseBanner = Model.ShowPhaseBanner && !string.IsNullOrEmpty(phase) && phase != "Live";
+    var headerContent = Model.HeaderBarContent;  
+        
 }
+<head>
```
sussexrick: You can't include the JavaScript files this way. There can only be one `<head>` in a page and this partial view will be used inside `<body>`. Instead you can put a separate partial view in https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/tree/develop/ThePensionsRegulator.Frontend/Views/Shared/TPR and document that it needs to be included by any sites wanting to use the search with autocomplete.

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
```diff
@@ -11,7 +11,13 @@
     var headerLogoAlt = Model.LogoAlternativeText;
     if (string.IsNullOrEmpty(headerLogoAlt) && !string.IsNullOrEmpty(Model.LogoHref)) { headerLogoAlt = ComponentGenerator.HeaderLogoDefaultAlt; }
     var showPhaseBanner = Model.ShowPhaseBanner && !string.IsNullOrEmpty(phase) && phase != "Live";
+    var headerContent = Model.HeaderBarContent;
```
sussexrick: This new variable is never used.

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
```diff
@@ -30,14 +36,17 @@
         {
             <tpr-header-bar class="@Model.HeaderBarClass">
                 <tpr-header-bar-logo alt="@headerLogoAlt" href="@Model.LogoHref" />
-                <tpr-header-bar-label>@Model.HeaderBarLabel</tpr-header-bar-label>
-                <tpr-header-bar-content allow-html="@Model.HeaderBarContentAllowHtml">@Html.Raw(Model.HeaderBarContent)</tpr-header-bar-content>
+                <tpr-header-bar-label>@Model.HeaderBarLabel</tpr-header-bar-label> 
+              <tpr-header-search action="@Model.ActionPath"></tpr-header-search>
```
sussexrick: You've _replaced_ the existing support for content with support for search. You need an if statement that supports a choice of one or the other.

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/HeaderFooterTpr.generated.cs
sussexrick, line 1: In Umbraco you could look at adding fields to `Document types > TPR > TPR header` that allow a site to configure text strings like the placeholder text and aria-label text.

sussexrick (changes requested):
I like what you've done @Max-Emery, this is a great first effort. It works well within the context of the example app. Now you need to make some changes so that it is:
- more consistent with how other components are built (mostly in the names and styles)
- able to be packaged up for re-use outside the example app
- localisable, so that Welsh is fully supported
- not reliant on JavaScript for the basic search

You also need to add details of your component to https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/docs/components/tpr-header-bar.md

GovUk.Frontend.Umbraco.ExampleApp/wwwroot/js/autocomplete.js
Max-Emery: Updated as suggested

GovUk.Frontend.Umbraco.ExampleApp/wwwroot/js/autocomplete.js
Max-Emery: Amended

GovUk.Frontend.Umbraco.ExampleApp/wwwroot/js/autocomplete.js
Max-Emery: Amended

GovUk.Frontend.Umbraco.ExampleApp/wwwroot/js/autocomplete.js
Max-Emery: Updated

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/headersearch.config
Max-Emery: Amended

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Updated as suggested

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
Max-Emery: Amended, mirrors current site better

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
Max-Emery: Removed variable

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
Max-Emery: Added back in support for content and added choice for search

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/HeaderFooterTpr.generated.cs
Max-Emery: Added fields to Umbraco

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Updated as suggested

GovUk.Frontend.Umbraco.ExampleApp/Views/TPRHeader.cshtml
Max-Emery: Amended

ThePensionsRegulator.Frontend/Models/TprHeaderLockupModel.cs
Max-Emery: Default value removed

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Updated as suggested

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Updated as suggested

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Updated as suggested

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
Max-Emery: Amended

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
Max-Emery: Button updated

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
Max-Emery: class added

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
Max-Emery: Updated

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
Max-Emery: Updated

GovUk.Frontend.Umbraco.ExampleApp/Views/TPRHeader.cshtml
Max-Emery: Updated

ThePensionsRegulator.Frontend/TagHelpers/TprHeaderSearchTagHelper.cs
Max-Emery: Updated as suggested

ThePensionsRegulator.Frontend/TagHelpers/TprHeaderBarContext.cs
Max-Emery: Amended

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
Max-Emery: Updated as suggested

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
Max-Emery: Updated, moved to new file

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
Max-Emery: There isn't a specific reason for this value. It was the default size from the original CSS file packaged with accessible-autocomplete.

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
Max-Emery: Separate view created and documentation updated

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
sussexrick: You're still specifying text-align and flex-direction twice.

ThePensionsRegulator.Frontend/TagHelpers/TprHeaderBarContext.cs
sussexrick: You've updated `TprHeaderBarContext` but not `TprHeaderBar.DisplaySearchBar`.

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
sussexrick: Your new file still has `header-search` in the name, so it's still part of the specific component. It could just be `_tpr-autocomplete.scss`.

Also I noticed you could be using `$govuk-focus-width` instead of `3px`.

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/HeaderFooterTpr.generated.cs
sussexrick: The aliases for these should start with `tprHeader` like the other properties on the same document type. That will prevent a clash with any other placeholder and aria label settings on other document types.

Also it would be useful to say here, in either the name or the description, that this is the aria label for the search _button_:

BackOffice view of aria label property

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Languages/en-us.config
sussexrick, line 1: Not your fault - in Umbraco en-US is a zombie language that keeps coming back to life no matter how many times you kill it. 🧟🧟🧟

Please delete it from your installation and then you can delete this file.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/text-input.config
```diff
@@ -735,10 +735,12 @@
...
   ]
-}]]></Value>
-    </blocks>
-    <grid>
-      <Value Culture="cy"><![CDATA[{
+}]]>
+			</Value>
+		</blocks>
+		<grid>
+			<Value Culture="de">
```
sussexrick: German was changed to Welsh and you're reverting it. This just needs not to change.

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
sussexrick: This view is called `TPRHeaderSearchAutocomplete` but, similar to the SCSS file, it's not unique to the header. It could actually be used by other implementations of autocomplete on the page.

Also, you could be using this on the demo page `TPRHeader.cshtml`. Instead you have the script tags repeated in there except that the paths are different. The paths in this partial view are wrong.

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/HeaderSearch.generated.cs
sussexrick, line 1: I suspect you need to regenerate Models Builder models because I don't see any reference to your new fields `arialabelText` and `placeholderText`.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -83,9 +85,96 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
...
         }
+
+        public virtual TagBuilder GenerateTprHeaderSearch(TprHeaderBar tprHeaderBar)
+        {
+            var divTag = new TagBuilder("div");
+            if (tprHeaderBar.SearchAttributes != null) { divTag.MergeAttributes(tprHeaderBar.SearchAttributes); }
+            divTag.AddCssClass("tpr-header-search");
+
+            var form = new TagBuilder("form");
+            form.Attributes.Add("action", $"{tprHeaderBar.ActionPath}");
```
sussexrick: `ActionPath` is a nullable property so this could generate `action=""`. Either `ActionPath` should be `required`, or you should avoid generating an attribute if it's blank. A form will no action will send the response the current page by default.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -83,9 +85,96 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
...
+            searchField.AddCssClass("tpr-header-search__field");
+
+            var autoComplete = new TagBuilder("div");
+            autoComplete.AddCssClass("tpr-header-search-autocomplete");
+            searchField.InnerHtml.AppendHtml(autoComplete);
+
+            var autoCompleteContainer = new TagBuilder("div");
+            autoCompleteContainer.AddCssClass("tpr-autocomplete-container");
+            autoCompleteContainer.Attributes.Add("autocomplete-url", tprHeaderBar.AutoCompleteUrl);
+            autoCompleteContainer.Attributes.Add("placeholder", tprHeaderBar.SearchPlaceholderText);
```
sussexrick: `placeholder` isn't a valid attribute on a `div`

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -83,9 +85,96 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
...
+            var searchField = new TagBuilder("div");
+            searchField.AddCssClass("tpr-header-search__field");
+
+            var autoComplete = new TagBuilder("div");
+            autoComplete.AddCssClass("tpr-header-search-autocomplete");
+            searchField.InnerHtml.AppendHtml(autoComplete);
+
+            var autoCompleteContainer = new TagBuilder("div");
+            autoCompleteContainer.AddCssClass("tpr-autocomplete-container");
+            autoCompleteContainer.Attributes.Add("autocomplete-url", tprHeaderBar.AutoCompleteUrl);
```
sussexrick: `AutoCompleteUrl` is a nullable property so this could generate `autocomplete-url=""`. Either `AutoCompleteUrl` should be `required`, or you should avoid generating an attribute if it's blank.

Also `autocomplete-url` is not a valid attribute in HTML. `data-autocomplete-url` would be valid.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -83,9 +85,96 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
...
+            autoCompleteContainer.Attributes.Add("placeholder", tprHeaderBar.SearchPlaceholderText);
+            autoComplete.InnerHtml.AppendHtml(autoCompleteContainer);
+
+            var searchInput = new TagBuilder("input");
+            searchInput.Attributes.Add("aria-label", tprHeaderBar.SearchAriaLabel);
```
sussexrick: This ARIA label is for the `button`, so it shouldn't be on the `input` as well.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -83,9 +85,96 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
...
+            searchInput.Attributes.Add("placeholder", tprHeaderBar.SearchPlaceholderText);
```
sussexrick: `SearchPlaceholderText` is a nullable property so this could generate `placeholder=""`. Either `SearchPlaceholderText` should be `required`, or you should avoid generating an attribute if it's blank.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -83,9 +85,96 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
...
+            searchInput.Attributes.Add("type", "search");
+            searchInput.Attributes.Add("name", "query");
```
sussexrick: It would be good to make `"query"` configurable, so that the destination page can choose what query string parameter it wants to handle. It's fine to have `"query"` as a default.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -83,9 +85,96 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
...
+            searchInput.AddCssClass("govuk-input");
+            searchInput.AddCssClass("tpr-header-search__input");
+
+            autoCompleteContainer.InnerHtml.AppendHtml(searchInput);
+
+            var button = new TagBuilder("button");
+            button.Attributes.Add("type", "submit");
+            button.AddCssClass("tpr-header-search__button");
+            button.AddCssClass("govuk-input");
+            button.Attributes.Add("aria-label", tprHeaderBar.SearchAriaLabel);
```
sussexrick: `SearchAriaLabel` is a nullable property so this could generate `aria-label=""`. Either `SearchAriaLabel` should be `required`, or you should avoid generating an attribute if it's blank.

ThePensionsRegulator.Frontend/Models/TprHeaderLockupModel.cs
```diff
@@ -26,5 +26,10 @@ public class TprHeaderLockupModel
         public virtual string? Context1 { get; set; }
         public virtual string? Context2 { get; set; }
         public virtual string? Context3 { get; set; }
+        public virtual bool DisplaySearch { get; init; } = true;
```
sussexrick: Should this really default to `true`? Most TPR websites don't have this search box, so I think you should have no default value which means it defaults to `false`.

docs/components/tpr-header-bar.md
```diff
@@ -14,6 +14,7 @@ The Pensions Regulator (TPR) uses the TPR header bar as a consistent part of the
         <a class="govuk-link" href="#">A link</a>
         <a class="govuk-link" href="#">Another link</a>
     </tpr-header-bar-content>
+    <tpr-header-search>
```
sussexrick: This now has attributes you could be showing.

docs/components/tpr-header-bar.md
```diff
@@ -60,6 +61,21 @@ Typically used for a menu of relevant links.
...
 Must be inside a `<govuk-header-bar>` element.
 
+
+### <tpr-header-search>
+|    Attribute      | Type   | Description                                                           |
+|-------------------| ------ | ----------------------------------------------------------------------|
+|     `action`      |`string`| Sets action path of search form submission.                           |
+|`autocomplete-url` |`string`| Sets endpoint for autocomplete.js fetch operation.                    |
+|   `placeholder`   |`string`| Supports setting custom placeholder for generated input elements.     |
+|   `aria-label`    |`string`| Programmatically setting aria-label value of input and button elements|
```
sussexrick: Just the button. They're separate things so they shouldn't have the same label. Testing with NVDA the placeholder text is sufficient for the input.

docs/components/tpr-header-bar.md
```diff
@@ -60,6 +61,21 @@ Typically used for a menu of relevant links.
...
+
+TPR Header Search will display when there is no header content and DisplayHeaderSearch property is set to true. On smaller screen sizes the tpr mobile menu component will take over, and the header search will display as part of the mobile menu when expanded.
```
sussexrick: TPR should be in caps, and you should also explain that the mobile behaviour is not yet implemented and therefore this is currently a desktop-only component.

docs/components/tpr-header-bar.md
```diff
@@ -60,6 +61,21 @@ Typically used for a menu of relevant links.
...
+
+TPR Header Search implements the alphagov/accessible-autocomplete component, rendering an input box and drop-down box for search resuts.
```
sussexrick: It would be useful to link the text `alphagov/accessible-autocomplete` to the repo.

docs/components/tpr-header-bar.md
```diff
@@ -60,6 +61,21 @@ Typically used for a menu of relevant links.
...
+
+Views which require the tpr-header-search should also include the 'TPRHeaderSearchAutocomplete' partial view, in order to use the autocomplete functionality.
```
sussexrick: `<tpr-header-search>` rather than `tpr-header-search`

sussexrick (dismissed):

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
Max-Emery: Duplicate code now removed

ThePensionsRegulator.Frontend/TagHelpers/TprHeaderBarContext.cs
Max-Emery: TprHeaderBar.DisplaySearchBar also updated

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
Max-Emery: Amended

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
Max-Emery: This has been amended

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/HeaderFooterTpr.generated.cs
Max-Emery: Properties renamed

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Languages/en-us.config
Max-Emery: Actioned

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/text-input.config
Max-Emery: I've changed this back

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/HeaderSearch.generated.cs
Max-Emery: Models regenerated

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Amended

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Removed from div

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Amended

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Amended

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Amended

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: This is now configurable

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Amended

ThePensionsRegulator.Frontend/Models/TprHeaderLockupModel.cs
Max-Emery: Removed default

docs/components/tpr-header-bar.md
Max-Emery: Attributes added to docs

docs/components/tpr-header-bar.md
Max-Emery: Amended

docs/components/tpr-header-bar.md
Max-Emery: Amended

docs/components/tpr-header-bar.md
Max-Emery: Amended

docs/components/tpr-header-bar.md
Max-Emery: Amended

ThePensionsRegulator.Frontend/Models/TprHeaderLockupModel.cs
sussexrick: For consistency I think `DisplaySearch` needs to be renamed to `ShowSearch`. All the other toggles for elements of the header are called `Show<thing>`.

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/HeaderSearch.generated.cs
sussexrick: I think you need to regenerate again. On downloading your branch I'm getting a build error on a ModelsBuilder file:

image

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
sussexrick: This isn't quite working. If you try your example in the 'header and footer' example page you see that the query string now includes two versions of the parameter:

image

Also, I think the attribute on the tag helper needs to be `input-name` not `input-query` because it's the `name` attribute that's being set. That would be consistent with other tag helpers. Follow that through the code, eg `SearchInputQuery` becomes `SearchInputName`.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
sussexrick: This looks good, but VS is highlighting a stray semi-colon:

image

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/text-input.config
sussexrick: This file has a stray letter 's' in it.

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
```diff
@@ -0,0 +1,75 @@
...
+}
+
+@media(max-width: 960px) {
+
+    .tpr-header-search {
+        width: 20%;
+        margin-left: 50px;
+    }
+
+    .searchbox {
```
sussexrick: I don't think .searchbox is used anywhere in your code. If not, it can be removed.

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/HeaderFooterTpr.generated.cs
sussexrick: You need to use these properties in `UmbracoTprHeaderLockupModel.cs` too.

ThePensionsRegulator.Frontend/Models/TprHeaderLockupModel.cs
sussexrick: Also I think these four properties need to start with `Search` to clarify what they apply to. Notice how that's been done for other properties relating to `SkipLink` and `Logo` for example.

image

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/HeaderFooterTpr.generated.cs
Max-Emery: Added to file

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/text-input.config
Max-Emery: Removed

ThePensionsRegulator.Frontend/Styles/_tpr-header-search.scss
Max-Emery: Amended

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Removed

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Removed JS code that was adding extra query string value, name updated

ThePensionsRegulator.Frontend/Models/TprHeaderLockupModel.cs
Max-Emery: Updated

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/HeaderSearch.generated.cs
Max-Emery: I've re-ran the model builder

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
sussexrick: The JS isn't working. It no longer adds the duplicate query string parameter, but it also doesn't add the search term: 

image

It does work without JS.

Also, you didn't follow the name change through the code. It's starts out as `InputName` but soon becomes `InputQuery` as soon as you pass it into `TprHeaderBarContext.SetSearch()`

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
sussexrick: This is better now, but `autocomplete.js` still has several references to header search that don't need to be there. That still makes this partial view limited to the header search when it could be generic for any use of autocomplete on the page.

1. your variables `searchBars` and `searchBar` could be something like `containers` and `container` that don't make any assumptions about what the autocomplete is used for.
2. `document.querySelectorAll('.tpr-header-search__input')` could be `searchBar.querySelector('.govuk-input')` since you don't actually want multiple results from the whole page, you only want the first input within the container you've identified. With only one result you then don't need the `forEach` that follows it.
3. On line 21 the hard-coded id `tpr-header-search-autocomplete` could be found. `searchBar.querySelector(".autocomplete__input").id` should do it.
4. In the error message on line 30 you could just replace `tpr-header-search` with `autocomplete`

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Amended

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
Max-Emery: Updated 1,2,4.

**question:** id: `tpr-header-search-autocomplete` is setting the value for the id on the element, this is needed in order to generate the search box, is it better to leave it as `tpr-autocomplete__input`, or should this value for the id come from another element and be able to be set on the tag helper?

AlexBeesley (approved):
Approving this, good work @Max-Emery. Have tested this locally and Max has responded to @sussexrick suggestions with appropriate code changes.

PR #406, opened 2025-04-09
Upgrade Umbraco to 13.7.2 because everything below 13.7.1 has a vulnerability

sussexrick:
Also upgrades `uSync` and `AspNetCore.SassCompiler`.

Some previous NuGet direct references are removed because non-vulnerable versions are now brought in by the updated Umbraco package.


villarin (approved):
Tested locally off an empty database and working as expected.

PR #410, opened 2025-04-15
The Umbraco example app crashed when trying to render a TPR component

### LINKED ISSUE #335: Umbraco example app crashes when run with GOV.UK styles
By default the Umbraco example app runs with TPR styles, which works. If you disable TPR styles in appsettings.config, some pages (notably the home page) crash because they can't find the view for a TPR component. 

These should instead render the component as unstyled HTML.

sussexrick:
The Umbraco example app crashed when trying to render a TPR component but set to use GOV.UK styles. 

This solves that problem by:
- registering the path provider for TPR components regardless, so that it can find TPR views
- re-using an existing solution from the 'TPR header and footer' page to display a message for other TPR components that either crash or don't make sense without TPR services and styles loaded
- leaves TPR components active where they degrade gracefully (related links still looks fine)

Fixes #335.


BenHughesTPR (approved):
Ran up Example app locally and it worked without issues.

PR #414, opened 2025-04-29
Remove the dependency on Yarn Classic

sussexrick:
Swap out all usage of Yarn Classic for npm, so we have one less dependency when developing.

Also updates vulnerable dependencies used when running JavaScript tests.

Also updates the deprecated `PublishCodeCoverageResults@1` task to the current `PublishCodeCoverageResults@2`.


ZebSadiq (approved):
Tested both example apps locally and all seems to be working without doing any yarn install

PR #418, opened 2025-05-08
Enforce 4 digits on input years

BenHughesTPR:
Adds validation to enforce 4 digits on input dates in line with GDS


ancientyouthman (approved):

sussexrick (approved):

sussexrick:
@BenHughesTPR Approved, but this code is [copied and adapted from our source project](https://github.com/x-govuk/govuk-frontend-aspnetcore/blob/main/src/GovUk.Frontend.AspNetCore/ModelBinding/DateInputModelConverterModelBinder.cs) and that still affects our non-Umbraco uses of this component. I would suggest raising an equivalent PR, or at least logging an issue.

PR #421, opened 2025-05-14
Fix Error Summary not display issue

HarryAmmon:
Since version [5.4.0 of the govuk-frontend project](https://github.com/x-govuk/govuk-frontend-aspnetcore/commit/f0930e3d5ed59da985d747942b7d243c49e8b5d9#diff-a462afec99464d0eb0bb3027dc0787b24264f5411f363bc11faed9c29ab48033) the Error Summary component is no longer creating a UL by default. A UL is only created if errors are present when the tag helper code runs.

To restore the functionality of JS updating Error Summaries when there is a validation error, we now create the UL if one isn't already present.


sussexrick:
@HarryAmmon You need to update JS tests to match

sussexrick (approved):

PR #422, opened 2025-05-16
Section cards new tabs text changes

ZebSadiq:


ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSectionCards.cs
```diff
@@ -50,6 +47,25 @@ public virtual TagBuilder GenerateTprSectionCards(TprSectionCards tprSectionCard
                             anchorElement.InnerHtml.Append(card.Title.ToHtmlString());
                         }
 
+                        if (!string.IsNullOrWhiteSpace(card.TitleTarget))
+                        {
+                            anchorElement.Attributes.Add("target", card.TitleTarget);
```
sussexrick: I don't think we need to support targets other than `_blank` because the Umbraco UI doesn't support setting those and we don't have a use case.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSectionCards.cs
```diff
@@ -50,6 +47,25 @@ public virtual TagBuilder GenerateTprSectionCards(TprSectionCards tprSectionCard
...
+                            if (card.TitleTarget.ToLower() == "_blank")
```
sussexrick: When setting `target="blank"` we should also set `rel="noopener noreferrer"`

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSectionCards.cs
```diff
@@ -50,6 +47,25 @@ public virtual TagBuilder GenerateTprSectionCards(TprSectionCards tprSectionCard
...
+                            {
+
+                                var newTabText = new TagBuilder("span");
+                                switch (culture)
```
sussexrick: Rather than this switch you could use the dictionary entry we already have for this text - see https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/pull/447/files (you will need to merge develop into your branch to get the constant) - and section cards docs should note that this dictionary entry can be set.

sussexrick (changes requested):

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSectionCards.cs
ZebSadiq: The issue is that this project doesn't have access to the dictionary as it's not Umbraco aware. Should I pass in the translation to it using an attribute to the tag helper?

sussexrick: Yes, that's what I've recommended for other text strings inside tag helpers. Then the tag helper itself doesn't have to be multi-lingual, it can be handled by the consuming app - whatever that is.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSectionCards.cs
ZebSadiq: Updated

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSectionCards.cs
ZebSadiq: Added

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSectionCards.cs
ZebSadiq: Updated

sussexrick (approved):

PR #434, opened 2025-06-13
Create a Q&As search results component

HarryAmmon:
Changes:
- Creates a Q&A search block to display Q&A search results
- Provides default headers and links, default links intentionally do not do anything at the moment as we don't know exactly what they will do
- Adds a JS file that creates an accordion with the search results 

This component is not yet supported by Umbraco, this will be added as a seperate PR


GovUk.Frontend.ExampleApp/Views/Home/Index.cshtml
```diff
@@ -43,6 +43,7 @@
     <li><a href="/dividerlinesandboxes" class="govuk-link">divider lines and boxes</a></li>
     <li><a class="govuk-link" href="/documents">documents</a></li>
     <li><a class="govuk-link" href="/relatedlinks">related links</a></li>
+    <li><a class="govuk-link" href="/searchresults">search results</a></li>
```
sussexrick, line 46: This is a different title to the page it leads to - we need to be clear whether this is a generic search results component or a specific one, because there will be other kinds of search results.

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
```diff
@@ -0,0 +1,42 @@
...
+    <div class="govuk-grid-column govuk-grid-column-full">
+        <h1 class="govuk-heading-l">@ViewData["Title"]</h1>
+    </div>
+</div>
+
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column govuk-grid-column-two-thirds-from-desktop">
+        <p class="govuk-body">
+            The Q&As search results block can be added to any page where you want the user to be able to search the Q&A content accessible on <a class="govuk-link" href="https://www.thepensionsregulator.gov.uk/faq-landing-page">The Pensions Regulator website</a>. Currently this is focused on Q&As for Automatic Enrolement.
```
sussexrick: Typo - "Enrolment" has an extra e

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
```diff
@@ -0,0 +1,42 @@
...
+        </p>
+
+        <p class="govuk-body">
+            The title and footer links can be overridden.
+        </p>
+
+        <p class="govuk-body">
+            This component relies on JavaScript to display content. If JavaScript is not available then this component will not display. The JavaScript file 'tpr-search.js' is provided as part of the 'ThePensionsRegulator.Frontend' nuget package. This will need to be included in any view that uses this component.
```
sussexrick: So if there's no JS what's the recommendation? Link to a page on the corp site?

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
```diff
@@ -0,0 +1,42 @@
...
+        </p>
+
+        <p class="govuk-body">
+            The URL for the required API can be configured in appsettings.json file under the key "FaqApiUrl". This should point to the API that provides the Q&A content. The API should return a JSON response that matches the expected structure for the search results.
+        </p>
```
sussexrick, line 86: Where can I find out what this expected structure is?

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
```diff
@@ -0,0 +1,42 @@
...
+        <tpr-search faq-api-url=@config["FaqApiUrl"]>
```
sussexrick: You've got an alignment issue:

Search box misaligned relative to page content

DevTools highlighting misalignment on left and right

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
```diff
@@ -0,0 +1,42 @@
...
+        <p class="govuk-body">
+            The title and footer links can be overridden.
+        </p>
+
+        <p class="govuk-body">
+            This component relies on JavaScript to display content. If JavaScript is not available then this component will not display. The JavaScript file 'tpr-search.js' is provided as part of the 'ThePensionsRegulator.Frontend' nuget package. This will need to be included in any view that uses this component.
+        </p>
+
+        <p class="govuk-body">
+            The URL for the required API can be configured in appsettings.json file under the key "FaqApiUrl". This should point to the API that provides the Q&A content. The API should return a JSON response that matches the expected structure for the search results.
```
sussexrick: Looks like the URL can be configured any way you like (which is good), and this example app uses the setting you mention. The key info is it's set on the tag helper.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearch.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public partial class ComponentGenerator
+    {
+        internal const string SearchElement = "aside";
+
+        public virtual TagBuilder GenerateTprSearch(string faqApiUrl)
+        {
+            var section = new TagBuilder(SearchElement);
+            section.AddCssClass("tpr-search");
+            section.Attributes.Add("id", "tpr-search");
```
sussexrick: So long as there's only ever one of these on a page this works, but two would make a hard-coded id invalid.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchFooterLinks.cs
```diff
@@ -0,0 +1,76 @@
...
+                foreach(var link in footerLinks.Links)
+                {
+                    var li = CreateLink(link.Attributes, link.Content);
+                    list.InnerHtml.AppendHtml(li);
+                }
+            }
+            else
+            {
+                var builder1 = new HtmlContentBuilder();
+                builder1.Append("Show more questions");
```
sussexrick: Should this use your default constant?

I know you said a further update is still to come, but this text needs to be localisable to Welsh, as does "Visit automatic enrolment Q&As" below.

At this point in my review it feels like `<tpr-search-footer-links></tpr-search-footer-links>` should just allow HTML content, or links as content...

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchInput.cs
```diff
@@ -0,0 +1,57 @@
...
+                searchLabel = DefaultSearchLabel;
+            }
+
+            var outer = new TagBuilder("div");
+            outer.AddCssClass("govuk-grid-row");
+
+            var button = new TagBuilder("button");
+            button.AddCssClass("govuk-button");
+            button.AddCssClass("tpr-button--no-next-step");
+            button.Attributes.Add("id", "ask-button");
```
sussexrick: The comments about hard-coded ids and non-localisable text apply here too.

If you're going to hard-code ids, consider prefixing with the component `tpr-search-results-` like you do with classes, so that other things could have an `ask-button` without conflicting.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchInput.cs
```diff
@@ -0,0 +1,57 @@
...
+            var input = new TagBuilder("input");
+            input.Attributes.Add("type", "text");
+            input.Attributes.Add("id", "ask-input");
+            input.AddCssClass("govuk-input");
+
+            formGroup.InnerHtml.AppendHtml(label);
+            formGroup.InnerHtml.AppendHtml(input);
+
+            var heading = new TagBuilder("h2");
+            heading.AddCssClass("govuk-heading-m");
```
sussexrick: Is it always `govuk-heading-m`? The starting point for heading sizes varies depending on the type of site, and we support that, so we should support setting this class.

GOV.UK advice on heading sizes

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchFooterLinks.cs
```diff
@@ -0,0 +1,76 @@
...
+    public partial class ComponentGenerator
+    {
+        internal const string DefaultShowMoreQuestionsContent = "Show more questions";
+        public virtual TagBuilder GenerateTprSearchFooterLinks(TprSearchFooterLinks footerLinks)
+        {
+            var outer = new TagBuilder("footer");
+
+            var list = new TagBuilder("ul");
+            list.AddCssClass("govuk-list");
+            list.AddCssClass("tpr-search-results__links");
```
sussexrick: Is this component 'TPR search' or 'TPR search results' or 'Q&As search results'? It seems unclear. CSS classes vary.

ThePensionsRegulator.Frontend/Styles/_tpr-search-results.scss
```diff
@@ -0,0 +1,43 @@
+﻿@import '_tpr-variables.scss';
+@import 'govuk/base';
+
+.tpr-search-results__heading {
+    background-color: $tpr-colour-white-smoke;
```
sussexrick, line 5: Why set a background colour?

ThePensionsRegulator.Frontend/Styles/tpr.scss
```diff
@@ -81,4 +81,5 @@
 @import "_tpr-summary-card.scss";
 @import "_tpr-task-list.scss";
 @import '_tpr-timeline.scss';
-@import '_tpr-youtube-video.scss';
\ No newline at end of file
+@import '_tpr-youtube-video.scss';
+@import '_tpr-search-results.scss';
```
sussexrick: We normally include these in alphabetical order...

ThePensionsRegulator.Frontend/TagHelpers/TprSearchTagHelper.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    [HtmlTargetElement(TagName)]
+    public class TprSearchTagHelper : TagHelper
+    {
+        internal const string TagName = "tpr-search";
+
+        private readonly ITprHtmlGenerator _htmlGenerator;
+
+        [HtmlAttributeName("faq-api-url")]
+        public string FaqApiUrl { get; set; } = string.Empty;
```
sussexrick: `FaqApiUrl` is different to the page title `Q&As search results`. It also locks this component into a FAQ-like content, when all your other naming in code suggests it's more generic than that.

ThePensionsRegulator.Frontend/wwwroot/tpr/tpr-search.js
```diff
@@ -0,0 +1,124 @@
...
+document.addEventListener("DOMContentLoaded", function () {
+
+    const searchButton = document.getElementById("ask-button");
+    searchButton.addEventListener("click", buttonOnClick);
+
+    const searchAside = document.getElementById("tpr-search");
+    const apiUrl = searchAside.getAttribute("data-api-url");
+
+    async function initaliseAccordion() {
+        const response = await fetch(`${apiUrl}/api/faq/content/popular`, { method: 'GET', headers: { 'Content-Type': 'application/json' } });
```
sussexrick: So API URL is actually API base URL...

ThePensionsRegulator.Frontend/wwwroot/tpr/tpr-search.js
```diff
@@ -0,0 +1,124 @@
...
+        const element = document.createElement(elementName);
+        element.className = className;
+
+        return element;
+    }
+
+    function createNoResultsFoundHeading() {
+        const noResultsHeading = document.createElement("h3");
+        noResultsHeading.className = "govuk-heading-m";
+        noResultsHeading.textContent = "No Results Found";
```
sussexrick: Text should be sentence case... and you probably need it to be localisable too.

This shows how GOV.UK components handle localisation in JS: https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/GovUk.Frontend.AspNetCore.Extensions/wwwroot/govuk/govuk-js-init.js

ThePensionsRegulator.Frontend/wwwroot/tpr/tpr-search.js
```diff
@@ -0,0 +1,124 @@
...
+        accordionContent.innerHTML = content;
+
+        accordionSection.appendChild(accordionHeader);
+        accordionSection.appendChild(accordionContent);
+
+        return accordionSection;
+    }
+
+    async function fetchContentById(contentId) {
+        const response = await fetch(`${apiUrl}/api/faq/content/${contentId}`, { method: 'GET', headers: { 'Content-Type': 'application/json' } });
```
sussexrick: This component seems very tightly coupled to the API, so much that it's impossible to see the results view in the example app without the API running and configured. The example app could use a hard-coded JSON file to make the results visible, and implementing this would naturally make you implement a more decoupled solution that could support different URLs. This is how Max's autocomplete in the header search works.

ThePensionsRegulator.Frontend/wwwroot/tpr/tpr-search.js
```diff
@@ -0,0 +1,124 @@
+﻿import { Accordion } from '/govuk/all.min.js';
+document.addEventListener("DOMContentLoaded", function () {
```
sussexrick: There are no tests for this JS. Our test coverage for other JS in this library is pretty good.

There are no tests for the C# either. That's true for all our components and not a problem I expect you to solve. FYI I want to find time to get our own version of conformance tests working - we have it for the task list so it should be similar, and then we could much more easily add tests.

sussexrick (changes requested):

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchInput.cs
```diff
@@ -0,0 +1,57 @@
...
+            var button = new TagBuilder("button");
+            button.AddCssClass("govuk-button");
+            button.AddCssClass("tpr-button--no-next-step");
+            button.Attributes.Add("id", "ask-button");
+            button.InnerHtml.Append("Ask");
+
+            var formGroup = new TagBuilder("div");
+            formGroup.AddCssClass("govuk-form-group");
+
+            var label = new TagBuilder("label");
```
sussexrick: This label needs a `for` attribute to be linked to its input. Testing using NVDA it is not aware of the label.

Feedback for the content designers you're working with - "Q&As" is announced as "Q and uhs" not "Q and Ays", but "FAQs" is announced exactly as you'd expect.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchFooterLinks.cs
HarryAmmon: We are not required to provide Welsh translations for the search results componet when used on the AE pages. This is one of the few exceptions. In the intrest of time, I'd suggest that if localisation is required, this should be added later.

ThePensionsRegulator.Frontend/Styles/_tpr-search-results.scss
HarryAmmon: The design I am working off of has this background.

ThePensionsRegulator.Frontend/TagHelpers/TprSearchTagHelper.cs
HarryAmmon: I have added and renamed attributes to make them more generic for a questions and answers component

ThePensionsRegulator.Frontend/wwwroot/tpr/tpr-search.js
HarryAmmon: I have added attributes so each URL that is used can be configured. The example app now works without the actual API configured and running.

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
```diff
@@ -0,0 +1,54 @@
...
+        <p class="govuk-body">
+            The search results component can be added to any page where you want the user to be able to search the Q&A content accessible on <a class="govuk-link" href="https://www.thepensionsregulator.gov.uk/faq-landing-page">The Pensions Regulator website</a>.
+            Currently this is focused on Q&As for Automatic Enrolment.
+            This component should be accompanied with a link, styled as a button, that takes the user to the search results component on the page.
+        </p>
+
+        <p class="govuk-body">
+            The title and footer links can be overridden. 
+            Footer links will always contain a 'Show more questions' link, that will show the remaining three questions, two are shown when the page loads. 
+            The heading level of the title and it's CSS class can be configured via the 'tpr-search-results-input' tag helper.
```
sussexrick: typo - should be "its CSS class"

GovUk.Frontend.ExampleApp/wwwroot/QuestionsAndAnswersData/popularContentExample.json
```diff
@@ -0,0 +1,22 @@
...
+    "name": "What order should I watch the Star Wars movies in?",
+    "pageContent": "<p class=\"govuk-body\">You can watch Star Wars movies in release order, chronological order, or the 'Machete Order.' Release order starts with Episode IV, while chronological order starts with Episode I.</p>"
+  },
+  {
+    "name": "Who are the Sith and the Jedi?",
+    "pageContent": "<p class=\"govuk-body\">The Jedi are peacekeepers who use the light side of the Force, while the Sith are their adversaries who use the dark side for power and domination.</p>"
+  },
+  {
+    "name": "What is the difference between the Death Star and Starkiller Base?",
+    "pageContent": "<p class=\"govuk-body\">The Death Star is a massive space station capable of destroying planets, while Starkiller Base is a larger weapon built into a planet that can destroy entire star systems.</p>"
```
sussexrick: This encoding is different to that returned by the real API, which encodes things like < and > as Unicode references. It doesn't seem to matter as it handles both. Just FYI so feel free to resolve this comment.

GovUk.Frontend.ExampleApp/wwwroot/QuestionsAndAnswersData/searchResults.json
```diff
@@ -0,0 +1,24 @@
+{
+  "results": [
+    {
+      "key": "example1.json",
```
sussexrick: This is confusing because, having read your page, I don't think these keys are filenames. If they matched the `key` of `example1`, `example2` etc in your other JSON it would be clearer.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchResults.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        internal const string SearchElement = "aside";
+
+        public virtual TagBuilder GenerateTprSearchResults(string popularContentApiUrl, string searchContentApiUrl, string contentByIdUrl)
+        {
+            var section = new TagBuilder(SearchElement);
+            section.AddCssClass("tpr-search");
+            section.Attributes.Add("data-popular-content-url", popularContentApiUrl);
+            section.Attributes.Add("data-search-content-url", searchContentApiUrl);
+            section.Attributes.Add("content-by-id-url", contentByIdUrl);
```
sussexrick: The third one should be a `data-` attribute too, otherwise it's invalid HTML

ThePensionsRegulator.Frontend/wwwroot/tpr/tpr-search.js
```diff
@@ -0,0 +1,162 @@
...
+
+function createElementWithClassName(elementName, className) {
+    const element = document.createElement(elementName);
+    element.className = className;
+
+    return element;
+}
+
+function createNoResultsFoundHeading() {
+    const noResultsHeading = document.createElement("h3");
```
sussexrick: You've made the heading level and class configurable in the C#, but this code doesn't check and reflect that choice.

ThePensionsRegulator.Frontend/wwwroot/tpr/tpr-search.js
```diff
@@ -0,0 +1,162 @@
...
+    noResultsHeading.className = "govuk-heading-m search-results-no-results-found";
+    noResultsHeading.textContent = "No results found";
```
sussexrick: I think this needs to be localisable to Welsh on the sites you're targeting.

This shows how GOV.UK components handle localisation in JS: https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/GovUk.Frontend.AspNetCore.Extensions/wwwroot/govuk/govuk-js-init.js

azure-pipelines.yml
```diff
@@ -64,7 +64,7 @@ steps:
 
   - script: |
       cd $(Build.SourcesDirectory)\govuk-frontend-aspnetcore-extensions
-      npx jest --ci --watchAll=false --reporters=jest-junit --reporters=default --coverage --coverageReporters=cobertura
+      npm test
```
sussexrick, line 67: Do we still have coverage reporting?

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchFooterLinks.cs
sussexrick: You've settled on 'TPR search results' but your root element still uses `tpr-search`.

docs/contributing/run-tests.md
```diff
@@ -6,7 +6,7 @@ To run unit tests on the client-side validation JavaScript:
 
 ```cmd
 npm install
```
sussexrick, line 8: Having done this I've ended up with a different package-lock.json, so I think that needs to be part of the PR

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
sussexrick: I'm still finding the no-JS experience confusing. Your guidance says "This component should be accompanied with a link, styled as a button, that takes the user to the search results component on the page... If JavaScript is not available then this component will not display and the accompanying button should link the user to the Q&A landing page.". There is no such link on the example page, so it's not clear what the placement is or whether this link should still appear with JS present. Currently the entire functionality is gone without JS.

A better no-JS experience might be to display the search box anyway, but put it in a `<form method="get">`. Now without JS the form still works and generates a querystring that could be used to create a functional search. Add a parameter for the URL `<form method="get" action="/some-url">` and that functional search can live somewhere else.

You need the form in any case for accessibility and usability. If you press Enter in the search box it should trigger the button, but it doesn't. To do this your form fields should be within a `<form>` element and your button should have `type="submit"`. Then you probably also need to make sure your JS intercepts and cancels the `onsubmit` event rather than handling `onclick` of the button.

sussexrick:
In the example app, even though you now have test data and display results, if I enter a search term and click "Ask" nothing happens.

azure-pipelines.yml
HarryAmmon: I can't see any coverage reporting for the JS tests in the pipeline run for this branch _but_ I can't see it in any of the runs for recent PRs either. I may just not know where to look but if it's not working, I don't think that was introduced by this PR.

GovUk.Frontend.ExampleApp/wwwroot/QuestionsAndAnswersData/searchResults.json
HarryAmmon: I'm a bit confused by your suggestion here. I have made all of the keys end in `.json` in `example1` etc and in `searchResults.json`. For the example app, the keys need to end in `.json` as (annoyingly) the browser returns a 404 when you request `https://localhost:44326/SearchResultsData/example1`

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchFooterLinks.cs
sussexrick: You still refer to `tpr-search.js` in the ASP.NET example app page

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchResults.cs
sussexrick: Two instances in the ASP.NET example app are unchanged (one is in the page text and may be intentional since all three names are missing the `data-` prefix)

azure-pipelines.yml
sussexrick: Yeah, fair point. Looks like this was a victim of Azure DevOps limitation. It only lets you visualise one set of coverage data and we've chosen C#.

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
```diff
@@ -0,0 +1,61 @@
...
+        <p class="govuk-body">
+            This component relies on JavaScript to display content. 
+            If JavaScript is not available then this component will not display and the accompanying button should link the user to the Q&A landing page.
+            The JavaScript file 'tpr-search-results.js' is provided as part of the 'ThePensionsRegulator.Frontend' nuget package. This will need to be included in any view that uses this component.
+        </p>
+
+        <p class="govuk-body">
+            Three URLs are required to be configured to allow this component to show content. These are configured via attributes on the tag helper.
+            <ul class="govuk-list govuk-list--bullet">
+                <li>popular-content-url - This should return a list of content to show when the page loads. The example app provides <a href="/SearchResults/popularContentExample.json" target="_blank">an example of the expected structure</a></li>
```
sussexrick: All 3 of these links are broken

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
sussexrick: Is there a good reason this link opens a new tab with JS disabled, since in most cases this is bad for accessibility? Using NVDA it gives no indication it will open a new tab, and if I try to go back it says "can't go back".

HarryAmmon: Feedback from users on the previous 'widget' was they liked that they could access the Q&A content without losing their place on the current page. Opening in a new tab is a similar experiance

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
```diff
@@ -0,0 +1,61 @@
...
+
+        <tpr-search-results popular-content-url="/SearchResultsData/popularContentExample.json" content-by-id-url="/SearchResultsData" search-content-url="/SearchResultsData/searchResults.json">
+            <tpr-search-results-input heading-level=2 govuk-heading-class="govuk-heading-m"></tpr-search-results-input>
+            <tpr-search-results-footer-links><a href="/">Home</a></tpr-search-results-footer-links>
+        </tpr-search-results>
+    </div>
+
+    <div class="govuk-grid-column govuk-grid-column-one-third-from-desktop">
+        <div class="govuk-grid-row">
+            <govuk-button-link is-start-button="false" class="tpr-button--no-next-step tpr-search-results-nav-button" href="https://www.thepensionsregulator.gov.uk/faq-landing-page" target="_blank">Search Q&A's</govuk-button-link>
```
sussexrick: There's no apostrophe in Q&As

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
sussexrick: You should probably add that to the guidance in that case...

ThePensionsRegulator.Frontend/wwwroot/tpr/tpr-search.js
HarryAmmon: Incase you missed this comment, we think we have addressed this.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchResults.cs
HarryAmmon: The only other instance of the HTML missing the `data-` prefix was in the jest tests, which we've updated. All other instances are intentionally missing the prefix as that is name of the attribute on the tag helper, not the HTML attribute.

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
sussexrick: A `title` attribute is not the best way to make the behaviour clear because screen reader users will often have these set not to be read out. These are the most important audience for the hint since they're the audience that has no visual cue about what happened. Better to have "(opens in new tab)" as part of the link, using the `govuk-visually-hidden` class so that it's just for screen readers. Also, the example doesn't follow your suggestion.

https://www.bbc.co.uk/accessibility/forproducts/guides/html/title-attributes/

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchResults.cs
sussexrick: My mistake, sorry.

ThePensionsRegulator.Frontend/wwwroot/tpr/tpr-search.js
sussexrick: The JS code hard-codes the heading level as h3. That means that if the tag helper is set to use h2, the document outline correctly identifies the results as part of the 'Search Q&As' section:

h2 then h3

If the tag helper is set to use h3, the document outline suggests the results are siblings of the 'Search Q&As' section, not part of it:

all h3

If the tag helper is set to use h4, the document outline suggests the results are some completely new sub-section of whatever the parent of 'Search Q&As` was:
h4 then h3

GovUk.Frontend.ExampleApp/Views/SearchResults/Index.cshtml
```diff
@@ -0,0 +1,61 @@
...
+
+        <p class="govuk-body">
+            This component relies on JavaScript to display content. 
+            If JavaScript is not available then this component will not display and the accompanying button should take the user to the Q&A landing page in a new tab. Feedback from users is that they like not losing where they were on original page. Opening in a new tab provides a similar user experience. A title attribute can be added to the button to tell users that the link will open in a new tab.
+            The JavaScript file 'tpr-search-results.js' is provided as part of the 'ThePensionsRegulator.Frontend' nuget package. This will need to be included in any view that uses this component.
+        </p>
+
+        <p class="govuk-body">
+            Three URLs are required to be configured to allow this component to show content. These are configured via attributes on the tag helper.
+            <ul class="govuk-list govuk-list--bullet">
```
sussexrick: `ul` inside `p` is invalid

ThePensionsRegulator.Frontend/TagHelpers/TprSearchResultsInputTagHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+        public TprSearchResultsInputTagHelper() : this(null) { }
+
+        public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
+        {
+            var inner = await output.GetChildContentAsync();
+            var content = inner.GetContent();
+
+            var result = _htmlGenerator.GenerateTprSearchResultsInput(_headingLevel, _headingClass, content);
+
+            output.TagName = "div";
```
sussexrick: Setting this to `null` instead would remove a redundant `div`

sussexrick (approved):

PR #435, opened 2025-06-16
Move Phase banner to below Tpr header layout

PauliePension:
[AB#228599](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/228599)
Moved phasebanner in TPRHeaderLockup to be below the TPR header control.

### DIFF
--- a/ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
+++ b/ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
@@ -5,7 +5,7 @@
 @using ThePensionsRegulator.Frontend.HtmlGeneration;
 @using ThePensionsRegulator.Frontend.Models;
 @{
-    var phase = Model.Phase ?? "(None)";
+    var phase = Model.Phase ?? "Live";
     var skipLink = Model.SkipLinkText;
     if (string.IsNullOrEmpty(skipLink)) { skipLink = "Skip to main content"; }
     var headerLogoAlt = Model.LogoAlternativeText;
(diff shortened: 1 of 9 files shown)


ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
```diff
@@ -5,7 +5,7 @@
 @using ThePensionsRegulator.Frontend.HtmlGeneration;
 @using ThePensionsRegulator.Frontend.Models;
 @{
-    var phase = Model.Phase;
```
sussexrick, line 8: I guess this was silencing a null reference warning, but it does create the possibility that "(None)" would show up in a banner. It would be better to default to "Live" which is hidden (see line 13).

sussexrick (commented):
Looks good aside from the one minor point I've raised, but your build is failing due to a new CVE so that will also need to be dealt with, either here or in a separate PR. The CVE is a Node.js one, which means it only affects the Jest tests for our JavaScript. Running those verifies any fix.

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
PauliePension: done

PauliePension: @sussexrick I've added a fix for the CVE into the pipeline, and that ran fine.

sussexrick: Fixing that way in the pipeline means it still runs with the vulnerable version locally, so it would be better to actually update the vulnerable package in our source.

package.json
```diff
@@ -27,6 +27,7 @@
     }
   },
   "dependencies": {
+    "brace-expansion":"1.1.12",
```
PauliePension, line 30: @sussexrick Updated the troublesome package

sussexrick: So no need for npm audit fix in the pipeline then, which would hide vulnerabilities from us  that would still be active when running on laptops...

PauliePension: Done and run.

sussexrick (approved):

PR #442, opened 2025-06-23
Bullet and numeric styles were not displayed correctly in backoffice previews

sussexrick:
Umbraco includes the option to select the numbering style for an ordered list:

Ordered list styles

and similar style options for unordered lists:

Unordered list styles

The selection of styles was not being reflected in backoffice previews, which always showed the default option. 

The problem was that the `safe_html` filter applied in the [out-of-the-box backoffice preview](https://github.com/umbraco/Umbraco-CMS/blob/v13/main/src/Umbraco.Cms.StaticAssets/wwwroot/App_Plugins/Umbraco.BlockGridEditor.DefaultCustomViews/umbBlockGridDemoRichTextBlock.html) for a rich text block was missing. This PR adds it wherever we render rich text in all backoffice previews.

Backoffice previews now work for ordered lists:

Backoffice preview of an ordered list

and unordered lists:

Backoffice preview of an unordered list

I have added example lists to the 'Typography' example page for Umbraco to demonstrate/test these options. 

There is another separate issue with the rendering of these styles in the final markup for unordered lists only, but it is out-of-scope for this PR.

[AB#231921](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/231921) [AB#231924](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/231924)

### DIFF
--- a/GovUk.Frontend.Umbraco/App_Plugins/GOVUK/blocks/views/GovUkCheckboxesPreview.html
+++ b/GovUk.Frontend.Umbraco/App_Plugins/GOVUK/blocks/views/GovUkCheckboxesPreview.html
@@ -6,12 +6,12 @@
             <p class="backoffice-additional-blocks" ng-if="!block.data.fieldsetBlocks.contentData">No blocks.</p>
             <p class="backoffice-additional-blocks" ng-if="block.data.fieldsetBlocks.contentData.length===1">1 block.</p>
             <p class="backoffice-additional-blocks" ng-if="block.data.fieldsetBlocks.contentData.length>1">{{checkbox.fieldsetBlocks.contentData.length}} blocks.</p>
-            <div class="govuk-hint" ng-bind-html="block.data.hint.markup" ng-if="block.data.hint.markup"></div>
+            <div class="govuk-hint" ng-bind-html="block.data.hint.markup | safe_html" ng-if="block.data.hint.markup"></div>
             <div class="govuk-checkboxes">
                 <div class="govuk-checkboxes__item" ng-repeat="checkbox in block.data.checkboxes.contentData">
                     <input class="govuk-checkboxes__input" type="checkbox" value="{{checkbox.value}}">
                     <label class="govuk-checkboxes__label govuk-label">{{checkbox.label}}</label>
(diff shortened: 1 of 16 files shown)


GovUk.Frontend.Umbraco/App_Plugins/GOVUK/blocks/views/GovUkCheckboxesPreview.html
```diff
@@ -6,12 +6,12 @@
             <p class="backoffice-additional-blocks" ng-if="!block.data.fieldsetBlocks.contentData">No blocks.</p>
             <p class="backoffice-additional-blocks" ng-if="block.data.fieldsetBlocks.contentData.length===1">1 block.</p>
             <p class="backoffice-additional-blocks" ng-if="block.data.fieldsetBlocks.contentData.length>1">{{checkbox.fieldsetBlocks.contentData.length}} blocks.</p>
-            <div class="govuk-hint" ng-bind-html="block.data.hint.markup" ng-if="block.data.hint.markup"></div>
+            <div class="govuk-hint" ng-bind-html="block.data.hint.markup | safe_html" ng-if="block.data.hint.markup"></div>
```
AlexBeesley, line 9: **question** Why have you added this to the checkboxes and not the buttons?

sussexrick: Which buttons are you referring to?

AlexBeesley: Its more a question to just check you are adding this ability to all the datatypes that need it, I noticed there are views here for some but not all, so just wanted to make sure that you are adding it to all those backoffice views that need it.

Happy to resolve this if you are confident.

sussexrick: Yes, it only applies to properties using the rich text editor. Since these all have a `.markup` property I searched for and updated all instances of that.

AlexBeesley (approved):

PR #443, opened 2025-06-23
Custom bullet styles were not rendered when a strict CSP was applied

sussexrick:
Selecting custom bullet styles for an unordered list in the backoffice did not change the bullet style in the rendered page when a strict CSP was applied. This was because Umbraco applied the bullet style using a `style` attribute, which is often blocked with a strict CSP.

This was already solved for ordered lists in https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/pull/182 and the same solution is applied here. This removes all style attributes from unordered lists, but where a attribute value is recognised as representing a bullet style choice in the backoffice, a class is applied instead, which is allowed.

This PR also applies a CSP to both example apps, both as a test for this issue and to better represent how all of these components are used in practice.

[AB#231921](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/231921)

### DIFF
--- a/GovUk.Frontend.ExampleApp/Middleware/SecurityHeadersMiddleware.cs
+++ b/GovUk.Frontend.ExampleApp/Middleware/SecurityHeadersMiddleware.cs
@@ -1,0 +1,37 @@
+﻿using GovUk.Frontend.AspNetCore.Extensions.Security;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Hosting;
+using System;
+using System.Threading.Tasks;
+
+namespace GovUk.Frontend.ExampleApp.Middleware
+{
+    public class SecurityHeadersMiddleware
+    {
+        private readonly RequestDelegate _next;
+        private readonly IWebHostEnvironment _webHostEnvironment;
+        public SecurityHeadersMiddleware(RequestDelegate next, IWebHostEnvironment webHostEnvironment)
+        {
+            _next = next ?? throw new ArgumentNullException(nameof(next));
+            _webHostEnvironment = webHostEnvironment;
+        }
+
+        public async Task InvokeAsync(HttpContext context, INonceProvider nonceProvider)
+        {
+            var connectSrcForHotReload = _webHostEnvironment.IsDevelopment() ? "'self' ws://localhost:* http://localhost:64300" : string.Empty;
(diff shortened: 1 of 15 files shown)


GovUk.Frontend.ExampleApp/Middleware/SecurityHeadersMiddleware.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace GovUk.Frontend.ExampleApp.Middleware
+{
+    public class SecurityHeadersMiddleware
+    {
+        private readonly RequestDelegate _next;
+        private readonly IWebHostEnvironment _webHostEnvironment;
+        public SecurityHeadersMiddleware(RequestDelegate next, IWebHostEnvironment webHostEnvironment)
+        {
+            _next = next ?? throw new ArgumentNullException(nameof(next));
```
AlexBeesley, line 16: **question** Is this the correct exception type? I seem to remember you once telling me to never throw `ArgumentNullException`...

GovUk.Frontend.Umbraco.ExampleApp/Middleware/SecurityHeadersMiddleware.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace GovUk.Frontend.Umbraco.ExampleApp.Middleware
+{
+    public class SecurityHeadersMiddleware
+    {
+        private readonly RequestDelegate _next;
+        private readonly IWebHostEnvironment _webHostEnvironment;
+        public SecurityHeadersMiddleware(RequestDelegate next, IWebHostEnvironment webHostEnvironment)
+        {
+            _next = next ?? throw new ArgumentNullException(nameof(next));
```
AlexBeesley, line 17: **note** refer to previous question on `GovUk.Frontend.ExampleApp/Middleware/SecurityHeadersMiddleware.cs`

GovUk.Frontend.ExampleApp/Middleware/SecurityHeadersMiddleware.cs
sussexrick: Yes it is. `ArgumentNullException` is the correct type if the argument is null and not allowed to be, which is the case here. If there's any other issue with an incoming argument then `ArgumentException` is the appropriate type.

See https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types#argumentexception-argumentnullexception-and-argumentoutofrangeexception

AlexBeesley (approved):

PR #445, opened 2025-06-25
Support custom classes and alignment 

sussexrick:
## Support for central and right alignment
Content designers have requested the ability to centre or right-align text, and the GOV.UK design system already provides classes for this. Alignment buttons are added to rich text editors where this could be appropriate. However, when they're used Umbraco by default adds the alignment as a `style` attribute, which is often blocked by a strict CSP. 

This was already solved for ordered lists in https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/pull/182 and the same solution is applied here. This removes all `style` attributes from paragraphs, but where a attribute value is recognised as representing an alignment choice in the backoffice, a GOV.UK class is applied instead, which is allowed.

Alignment options in rich text editor

Alignment supported

This led to the discovery and fixing of some further bugs, and extra example content to demonstrate that it's working.

## Alignment for single paragraphs
If alignment is supported, then it should be supported in the Details component since that is used at TPR to make long, detailed help text available. However, this component, along with Inset text, Warning text and hints within form components, is defined by the GOV.UK design system not to have a surrounding paragraph, so there was no paragraph to apply a class to. So this PR updates those scenarios to support rendering a paragraph element when an alignment class is applied.

## Support for custom classes
Similarly, we already had support for defining a custom class in the rich text editor and applying it, but if applied to a single paragraph in one of the scenarios listed above it didn't work, because the paragraph and the class would be removed. The PR supports applying custom classes in those scenarios.

How to add custom classes

Inset text with custom classes

## Support for multiple paragraphs
While Inset text, Warning text and form element hints are typically just one paragraph, and shown that way in the GOV.UK design system, there could be more. It's not appropriate to add a lot of content but a couple of paragraphs might be. When doing this the paragraphs have `.govuk-body` applied automatically and the styling differed to the same component without a paragraph tag, so the styling is updated to allow those to render as expected.

Details with complex formatting

## Notification banner
It's not appropriate to support alignment in a Notification banner, which is centred by default in TPR styles. Because left is the default in most places it doesn't apply a class, but this meant that left alignment within a Notification banner didn't override the default centred alignment.

Notification banner now uses dedicated data types so that it can define the rich text options that are appropriate. The 'GOV.UK Rich text editor' data type, which was only used by Notification banner and not clearly named, is removed in favour of two new ones with clearer naming.

Notification banner element type using new data types

[AB#231917](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/231917)


villarin (dismissed):
New rich text tools available and working as expected locally. 
@sussexrick  - Only indents missing now as far as I can tell?

sussexrick:
> New rich text tools available and working as expected locally. @sussexrick - Only indents missing now as far as I can tell?

@villarin Yes, and the PR is up for that. If you're happy with this one please approve.

villarin (approved):

PR #447, opened 2025-06-25
Pass through target attribute to links, styled as a button 

HarryAmmon:
Changes:
- Pass through target attribute if available
- If target attribute is `_blank` set `rel="noreferrer noopener"` and add `<span class="govuk-visually-hidden">(opens in a new tab)</span> as per [govuk standard](https://design-system.service.gov.uk/styles/links/#opening-links-in-a-new-tab)

### DIFF
--- a/GovUk.Frontend.Umbraco/Views/Shared/GOVUK/GovUkLinkAsButton.cshtml
+++ b/GovUk.Frontend.Umbraco/Views/Shared/GOVUK/GovUkLinkAsButton.cshtml
@@ -24,14 +24,23 @@
 
 @if(!string.IsNullOrEmpty(target))
 {
-    <govuk-button-link is-start-button="@isStartButton" href="@link?.Url" id="@Model.Content.Key" class="@string.Join(' ', buttonStyle)@cssClasses" target="@target">
-        @text
+    @if (openInNewTab)
+    {
+        <govuk-button-link is-start-button="@isStartButton" href="@link?.Url" id="@Model.Content.Key" class="@string.Join(' ', buttonStyle)@cssClasses" target="@target" rel="noreferrer noopener">
+            @text
 
-        @if (openInNewTab)
-        {
-            <span class="govuk-visually-hidden">(opens in new tab)</span>
-        }
-    </govuk-button-link>
+            @if (openInNewTab)
+            {
+                <span class="govuk-visually-hidden">(opens in new tab)</span>
+            }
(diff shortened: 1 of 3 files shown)


GovUk.Frontend.Umbraco/Views/Shared/GOVUK/GovUkLinkAsButton.cshtml
```diff
@@ -9,12 +9,42 @@
...
\ No newline at end of file
+
+@if(!string.IsNullOrEmpty(target))
+{
+    @if (openInNewTab)
+    {
+        <govuk-button-link is-start-button="@isStartButton" href="@link?.Url" id="@Model.Content.Key" class="@string.Join(' ', buttonStyle)@cssClasses" target="@target" rel="noreferrer noopener">
+            @text
+
+            @if (openInNewTab)
```
sussexrick: This is always true

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/GovUkLinkAsButton.cshtml
```diff
@@ -9,12 +9,42 @@
...
+            {
+                <span class="govuk-visually-hidden">(opens in new tab)</span>
```
sussexrick: This text should be localisable. You could add a property to `DictionaryConstants.cs` and use `@Umbraco.GetDictionaryValueOrDefault()` to get a localised dictionary value, or this default text if the dictionary value is absent.

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/GovUkLinkAsButton.cshtml
```diff
@@ -9,12 +9,42 @@
...
+            }
+        </govuk-button-link>
+    }
+    else
+    {
+        <govuk-button-link is-start-button="@isStartButton" href="@link?.Url" id="@Model.Content.Key" class="@string.Join(' ', buttonStyle)@cssClasses" target="@target">
```
sussexrick: I wouldn't bother supporting this scenario. Umbraco only lets you select `_blank` or no target, so you should never get here anyway. I don't know of a good use case for supporting any other targets even if we could.

sussexrick (changes requested):

sussexrick (approved):

PR #448, opened 2025-06-25
Support indenting text up to 5 levels

sussexrick:
Content designers have requested the ability to indent text in the rich text editor. This PR adds indent buttons to rich text editors for the Details, Text, Inset text and Warning text components and Hint fields within form components. 

Indent options added to rich text toolbar

Examples are updated showing this for Details and Text, where it is most likely to be used.

Indenting example

When the indent buttons are used Umbraco by default adds the alignment as a `style` attribute, which is often blocked by a strict CSP. This was already solved for ordered lists in https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/pull/182 and the same solution is applied here. This removes all `style` attributes from paragraphs, but where a attribute value is recognised as representing a indent applied in the backoffice, a GOV.UK class is applied instead, which is allowed.

The GOV.UK design system provides[ enough spacing classes to indent by one level](https://design-system.service.gov.uk/styles/spacing/) (assuming we use the same spacing as Umbraco's default of 40px per indent). This PR uses a GOV.UK class where we can and adds the classes we need to support up to five levels of indent, following the naming and spacing patterns established by the GOV.UK design system.

The code which replaces `style` attributes with classes is refactored to avoid repeatedly converting a string of HTML to an HtmlAgilityPack `HtmlDocument` and back again.

[AB#231916](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/231916)


villarin (approved):
Tested from vanilla setup and indents / outdent working as expected.

PR #454, opened 2025-06-30
#229347 Amend breadcrumb component to use different fields

Ethan-Lambourne:
In this PR:
- Created an 'IBreadcrumbLinksService' interface that can be implemented by other projects to have greater control over how their breadcrumbs name and url fields are collected;
- Created a 'BreadcrumbViewModel' that can be filled in by the class that implements 'IBreadcrumbLinksService';
- Implemented the new breadcrumb component within the ExampleApp;
- Created documentation showcasing how the new breadcrumb component implementation can be used.

Once this has been merged and a new version of the extensions project has been released I will update the corporate websites implementation of the breadcrumb to include additional fields to conclude the work item: https://dev.azure.com/thepensionsregulator/TPR/_workitems/edit/229347.


GovUk.Frontend.Umbraco.ExampleApp/Components/RenderBreadcrumbComponent.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public RenderBreadcrumbComponent(IBreadcrumbLinksService breadcrumbLinksService, IUmbracoPublishedContentAccessor publishedContext)
+        {
+            if (breadcrumbLinksService is not null && publishedContext is not null)
+            {
+                ViewModel = breadcrumbLinksService.GetLinks(publishedContext.PublishedContent);
+            }
+            else if (breadcrumbLinksService is null)
+            {
+                ViewModel.Error = "Breadcrumb links service is not initialised.";
```
HarryAmmon, line 22: **suggestion:** Whilst this is just an example, I'm not sure if handing errors in this way is the best idea. If either of those servies are null, I don't think we would want to display that to a user so it may  be better to throw a `ArgumentNullException` here instead. 

Though I don't think those services would ever be null, unless they haven't been registered, in which case .NET would throw an exception about that, which would be logged

Ethan-Lambourne: I've added a thrown `ArgumentNullException` and removed where it would display the error on the view in my latest commit:

https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/pull/454/commits/0f7948e34e4a62ec3d808d4f1296b2e38df930bf

HarryAmmon (approved):

GovUk.Frontend.Umbraco/Views/Shared/GOVUK/Breadcrumbs.cshtml
villarin, line 1: This one got deleted - so to prevent backward compatibility issues we will include on the next release

PR #459, opened 2025-07-07
TprHostUpdater was incorrectly matching and updating TPR automatic enrolment websites

sussexrick:
This PR fixes the bug and also moves the unit tests to the correct project.

[AB#233726](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/233726)


villarin (approved):
Tested locally an external urls within the same tpr domain are working as expected now.

PR #466, opened 2025-07-15
Fixes for the Breadcrumb component

ZebSadiq:
Fixed issues with Breadcrumb not working in consumed apps and moved breadcrumb to GovUk package + tidy up.


ZebSadiq:
Updating one more Property name

villarin (approved):

PR #478, opened 2025-07-24
#220663 Left hand side nav control

Ethan-Lambourne:
In this PR:
- Created a left hand side navigation component;
- Added example page(s) to the umbraco example app;
- Added javascript functionality for extending / retracting the menu on mobile if scripting is detected. Displays just the first level of navigation list items expanded if not.


Ethan-Lambourne:
Currently the side nav component can be implemented both within an Umbraco block grid and as a View Component. From today until the 14th of August I am on annual leave so if anyone needs changes made to this branch before then, you'll have to remote into it and make them yourselves.

ThePensionsRegulator.Frontend.Umbraco/Views/Components/TprSideNavigation/SideNavigation.cshtml
```diff
@@ -0,0 +1,69 @@
...
+@using ThePensionsRegulator.Frontend.Umbraco
+@using ThePensionsRegulator.Umbraco
+@using Umbraco.Cms.Core.Models
+@using Umbraco.Cms.Core.Models.PublishedContent
+@using Umbraco.Cms.Core.Routing
+@using Umbraco.Extensions
+
+@inject IPublishedUrlProvider publishedUrlProvider
+@{
+    IPublishedContent currentPage = Model.CurrentPage;
```
villarin: We should encapsulate the logic in this view in a service contract - so we can control the nav items that get displayed based on project requirements.
For that we could provide a view model i.e. NavigationItem with:
1. The label and url 
2. A list of children with the same properties: label and url.
3. Every children item can contain a nested list of children defined in step 2 above

ThePensionsRegulator.Frontend.Umbraco/Components/TprSideNavigationComponent.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.AspNetCore.Mvc;
+using ThePensionsRegulator.Frontend.Umbraco.Models;
+using Umbraco.Cms.Core.Models;
+using Umbraco.Cms.Core.Models.PublishedContent;
+
+namespace ThePensionsRegulator.Frontend.Umbraco.Components
```
villarin, line 8: Worth injecting a service dependency here to calculate the NavigationItems that will be rendered in the component view

villarin (changes requested):
Hi @Ethan-Lambourne 

As per single comments left on the view and component files. Let's introduce some level of abstraction to ensure we can extend this component to use across subscribing projetcs.

ThePensionsRegulator.Frontend.Umbraco/Views/Components/TprSideNavigation/SideNavigation.cshtml
Ethan-Lambourne: Myself and @ZebSadiq believe we have fulfilled these requirements and that this PR is ready for a re-review.

ThePensionsRegulator.Frontend.Umbraco/Components/TprSideNavigationComponent.cs
Ethan-Lambourne: Myself and @ZebSadiq believe we have fulfilled this requirement and that this PR is ready for a re-review.

ThePensionsRegulator.Frontend.Umbraco/Views/Components/TprSideNavigation/SideNavigation.cshtml
villarin: I can see the view model is populated via a service now and the view rendering the nav items as expected

ThePensionsRegulator.Frontend.Umbraco/Components/TprSideNavigationComponent.cs
villarin: Side navigation link service injected into view component now as expected

villarin:
@Ethan-Lambourne  - A performance issue has been identified on deeper levels causing the pages to timeout -https://localhost:44350/side-navigation/side-navigation-child-2/side-navigation-grandchild-2/. We need to identify the root cause of this issue. If the problem lies in the service layer we can progress this now and deal with the issue later as tech debt. Alternatively if the problem is present in the view layer we need to fix before approval and release.

villarin (dismissed):
Moving on with this one. A performance issue still remains which has been captured on https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/issues/493 for follow up

villarin (approved):

PR #482, opened 2025-08-04
Add a plugin that removes proprietary Word formatting on paste

sussexrick:
This PR:
- adds a plugin from https://github.com/pangaeatech/tinymce-paste-from-word-plugin that removes unwanted Word formatting on paste into TinyMCE
- enables the built-in Remove formatting button, which doesn't handle the extra formatting added by Word but is a quick way to remove bold, links, lists etc applied in Umbraco.

[AB#235666](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/235666)


villarin (approved):
Needs appsettings update

PR #492, opened 2025-08-21
Feature/204431 Create TPR header bar

Max-Emery:
In this PR
- Note: I've marked this PR as draft as there is still some work to do to fully complete the full implementation of the navigation service but could use a review of current work. 

- Created TPR Header Menu Component which supports setting link text and URL. Aria label for generated nav menu and a custom destination for a navigation page if JS is disabled.
- Component is used in both desktop and mobile views.
- Created tag helpers and added Header Menu to component generator interface.
- Can be implemented using a custom block list data type.
- Created `TPRGlobalNavigationService`  to dynamically populate navigation from content tree, respecting the the `umbracoNaviHide` property.
- Styling for desktop menu bar and mobile menu implemented.
- Docs Updated.


docs/components/tpr-header-bar.md
```diff
@@ -80,6 +87,34 @@ TPR Header Search implements the [alphagov/accessible-autocomplete](https://gith
 
 Views which require the `<tpr-header-search>` should also include the 'TPRHeaderSearchAutocomplete' partial view, in order to use the autocomplete functionality.
 
+### `<tpr-mobile-menu>`
+
+| Attribute         | Type   | Description                                                                                                                                                |
+| ----------------- | ------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
+|   `aria-label`    |`string`| Sets aria-label for `<ul>` element whithin navigation menu.                                                                                                |
```
sussexrick: Typo for "within"

docs/components/tpr-header-bar.md
```diff
@@ -80,6 +87,34 @@ TPR Header Search implements the [alphagov/accessible-autocomplete](https://gith
...
+
+Using the `<tpr-mobile-menu>` tag will generate the toggle as part of the header bar and the associated nav which will dispay underneath the header.
+Adding  `<tpr-mobile-menu-parent-item>` will create items to populate the the menu and `<tpr-mobile-menu-child-item>` can be nested inside these parent items to populate each sub menu.
+ 
+### `<tpr-mobile-menu-parent-item>`
+
+| Attribute    | Type   | Description                                                       |
+| ------------ | ------ | ----------------------------------------------------------------- |
+| `href`       |`string`| Sets destination for the menu item.                               |     
+| `link-text`  |`string`| Inner HTML value for menu item title.                             |
```
sussexrick: Does `link-text` need to support HTML, or could it just be text? It's good to allow just text if HTML support isn't required, as it removes any possibility of an [XSS vulnerability](https://owasp.org/www-community/attacks/xss/).

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprheadermenu.config
```diff
@@ -0,0 +1,66 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<ContentType Key="9fdef12a-0587-4817-8add-10464c32999d" Alias="tPRHeaderMenu" Level="2">
```
sussexrick: Casing of alias should be `tprHeaderMenu`

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprheadermenuchilditem.config
```diff
@@ -0,0 +1,66 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<ContentType Key="90600a82-4925-4287-bb81-45247a91a514" Alias="tPRHeaderMenuChildItem" Level="2">
```
sussexrick: Casing of alias should be `tprHeaderMenuChildItem`

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprheadermenuparentitem.config
sussexrick, line 1: This file reflects one of your earlier iterations and doesn't need to be in your PR.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprheadermenuparentitems.config
```diff
@@ -0,0 +1,82 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<ContentType Key="7cddfcdc-5921-4dcf-a203-2ce06c447ebf" Alias="tprHeaderMenuParentItems" Level="2">
```
sussexrick: Your alias is "items" but the name of the document type is "item"

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/TPRHeaderMenuChildItemBlockList.config
sussexrick, line 1: This file reflects one of your earlier iterations and doesn't need to be in your PR.

ThePensionsRegulator.Frontend/HtmlGeneration/TprHeaderMenuChildItem.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public class TprHeaderMenuChildItem
+    {
+        public AttributeDictionary? Attriubutes { get; set; }
+        public string? LinkText { get; set; }
+        public string? LinkDestination { get; set; }
+
+        public TprHeaderMenuChildItem(){}
+
+        public TprHeaderMenuChildItem(AttributeDictionary? attriubutes, string? linkText, string? linkDestination)
```
sussexrick: Typo "attributes"

ThePensionsRegulator.Frontend/HtmlGeneration/TprHeaderMenuChildItem.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Microsoft.AspNetCore.Mvc.ViewFeatures;
+
+namespace ThePensionsRegulator.Frontend.HtmlGeneration
+{
+    public class TprHeaderMenuChildItem
+    {
+        public AttributeDictionary? Attriubutes { get; set; }
+        public string? LinkText { get; set; }
```
sussexrick: Are the text and destination of a menu item really nullable? A menu item isn't much good without them. You could make them non-null and `required` instead.

ThePensionsRegulator.Frontend/HtmlGeneration/TprHeaderMenuParentItem.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Collections.Generic;
+
+namespace ThePensionsRegulator.Frontend.HtmlGeneration
+{
+    public class TprHeaderMenuParentItem
+    {
+        public AttributeDictionary? Attributes { get; set; }
+        public string? LinkText { get; set; }
+        public string? LinkDestination { get; set; }
+        public List<TprHeaderMenuChildItem>? HeaderMenuChildItems { get; set; } = new();
```
sussexrick: Are the text and destination of a menu item really nullable? A menu item isn't much good without them. You could make them non-null and `required` instead. Also for `HeaderMenuChildItems`, since you provide an empty list as a default value this could be non-nullable too unless you intend for people to set it to null.

ThePensionsRegulator.Frontend/Styles/_tpr-autocomplete.scss
```diff
@@ -141,3 +141,23 @@
...
+        display: block;
+        max-width: 100%;
+        margin-left: 19.5px;
+    }
+
+    .autocomplete__menu {
+        width: calc(99.5% - 4px);
+    }
+
+    .autocomplete__menu--visible {
```
sussexrick: No need for multiple uses of the `.autocomplete__menu--visible` - you could move properties from the earlier block into this one.

ThePensionsRegulator.Frontend/Styles/_tpr-header.scss
```diff
@@ -50,6 +52,13 @@ $tpr-header-left-width-xl: 200px;
     padding: $tpr-header-vertical-space 0;
 }
 
+.tpr-mobile-menu__container {
+    display: flex;
+    flex-direction: row;
```
sussexrick: You don't need to specify `flex-direction` twice.

ThePensionsRegulator.Frontend/TagHelpers/TprHeaderMenuParentItemTagHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        internal const string TagName = "tpr-header-menu-parent-item";
+
+        private const string UrlAttributeName = "href";
+        private const string LinkTextAttriubuteName = "link-text";
+
+        [HtmlAttributeName(UrlAttributeName)]
+        public string? Url {  get; set; }
+
+        [HtmlAttributeName(LinkTextAttriubuteName)]
```
sussexrick: Typo "Attribute"

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
sussexrick, line 1: On the mega-menu you've got some items aligned to the centre and some to the left.

Image

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
sussexrick, line 1: On the mega-menu when you tab to a top-level item like 'Employers' the dropdown opens automatically. The existing behaviour is preferable, where you press down arrow to open the dropdown and up arrow if you want to close it. 

The existing implementation also lets you use the left and right arrow keys to navigate through the top-level items, but that's not yet supported here.

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
sussexrick, line 1: The line below each child item on the mega menu is thinner on your version than on the existing implementation.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderMenu.cs
```diff
@@ -0,0 +1,206 @@
...
+
+            var navContainer = new TagBuilder("div");
+            navContainer.AddCssClass("navigation");
+            govContainer.InnerHtml.AppendHtml(navContainer);
+
+            var headerMenuList = new TagBuilder("ul");
+
+            if (!string.IsNullOrWhiteSpace(tprHeaderBar.HeaderMenuAriaLabel))
+            {
+                headerMenuList.Attributes.Add("aria-label", tprHeaderBar.HeaderMenuAriaLabel);
```
sussexrick: Put the `aria-label` on the `<nav>` element rather than the `<ul>` because if there is more than one `<nav>` element on the page [we should add an aria-label attribute to differentiate them](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/navigation_role).

Also in your example app use a realistic aria label like "Sitewide menu" rather than "Test label content. "

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderMenu.cs
```diff
@@ -0,0 +1,206 @@
...
+                    anchorTag.Attributes.Add("role", "button");
+
+                    mobileMenuItem.InnerHtml.AppendHtml(anchorTag);
+
+                    if (!string.IsNullOrWhiteSpace(item.LinkText))
+                    {
+                        anchorTag.InnerHtml.Append(item.LinkText);
+                    }
+
+                    var headerMenuSubMenu = new TagBuilder("div");
```
sussexrick: This sub-menu ought to be a nested list `<ul>` rather than a `<div>`. Using assistive technology a list tells you what you're dealing with, but a div tells you nothing.

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
```diff
@@ -0,0 +1,105 @@
...
+    });
+
+    const mediaQuery = window.matchMedia("(max-width: 993px)");
+
+    function expandMobileMenu(e) {
+        const arrows = document.querySelectorAll(".arrow");
+
+        arrows.forEach((arrow) => {
+            if (e.matches) {
+                arrow.setAttribute("tabindex", "0");
```
sussexrick: When you tab through the mobile menu it stops on each arrow because of the `tabindex=0` setting (which I think is this one...). The existing implementation doesn't do that.

When you have focus on the arrow NVDA just says "Clickable". If you press space it switches an `aria-expanded` attribute between `true` and `false` so it's clearly meant to expand and contract a menu... but since the existing implementation doesn't do that, I wonder if that's behaviour only meant for the mega menu that's leaking into the mobile one? This is probably also the reason that NVDA says "Employers button expanded" when tabbing to a mobile menu item. "Expanded" doesn't make sense here.

ThePensionsRegulator.Frontend/TagHelpers/TprHeaderBarTagHelper.cs
sussexrick, line 1: This is a warning in DevTools for the mobile menu toggle button:

Image

Also that `style="color: white"` shouldn't be there. If the colour setting is needed it should be part of a CSS class.

sussexrick, line 1: When I use NVDA and tab to the mobile menu toggle it says "Menu button collapsed button" which is good. If I press space to expand and collapse the menu it says nothing, which is bad. I suspect this action should be setting an `aria-expanded` attribute to `true` or `false` somewhere. 

The existing implementation is only slightly better. When you expand the menu the first time it reads out "Times close" (where times is the x symbol), so you get a sense that the menu opened. But there's no announcement for further closing and opening.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderMenu.cs
```diff
@@ -0,0 +1,206 @@
...
+            var mobileMenuToggle = new TagBuilder("a");
+            mobileMenuToggle.AddCssClass("tpr-mobile-menu-toggle");
+            mobileMenuToggle.Attributes.Add("href", tprHeaderBar.MobileMenuNoJsNavPage);
+
+            if (tprMobileMenu.Attributes != null)
+            {
+                mobileMenuToggle.MergeAttributes(mobileMenuToggle.Attributes);
+            }
+
+            var icon = new TagBuilder("div");
```
sussexrick, line 21: The search icon on the mobile menu has no alt text, so when I use NVDA it just announces "Graphic button". The same is true for the search icon on the desktop search.

Image

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
```diff
@@ -0,0 +1,105 @@
...
+            menu.classList.toggle("open");
+
+            const svg = document.querySelector(".tpr-mobile-menu__svg");
+            svg?.classList.toggle("hide");
+
+            const buttonText = document.querySelector(".closed");
+            if (buttonText) {
+                const isMenu = buttonText.textContent === "Menu";
+                buttonText.textContent = isMenu ? "Close" : "Menu";
+                buttonText.style.color = isMenu ? "#006ebc" : "white";
```
sussexrick: Set a class rather than a style, so that all styling is maintained together in the scss file.

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
```diff
@@ -0,0 +1,105 @@
...
+        menu.addEventListener("click", function () {
+            menu.classList.toggle("open");
+
+            const svg = document.querySelector(".tpr-mobile-menu__svg");
+            svg?.classList.toggle("hide");
+
+            const buttonText = document.querySelector(".closed");
+            if (buttonText) {
+                const isMenu = buttonText.textContent === "Menu";
+                buttonText.textContent = isMenu ? "Close" : "Menu";
```
sussexrick: These words need to be localisable, ie to be Welsh on Welsh pages and English on English pages.

Two possible approaches:

* use [i18n](http://i18njs.com/) which [we already have](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/GovUk.Frontend.AspNetCore.Extensions/wwwroot/govuk/govuk-js-init.js) because it's used by GOV.UK components
* set the text in an attribute on the server-side in an HTML attribute like `data-close-label="Close"`. Then it's up to the consuming app to set the right localised value in the app, and from Umbraco that's easy because it handles it.

ThePensionsRegulator.Frontend.Umbraco/Services/ITprGlobalNavigationSerivce.cs
sussexrick, line 1: You've got a typo in the filename of this file.

ThePensionsRegulator.Frontend.Umbraco/Services/ITprGlobalNavigationSerivce.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections.Generic;
+using ThePensionsRegulator.Frontend.HtmlGeneration;
+
+namespace ThePensionsRegulator.Frontend.Umbraco.Services
+{
+    public interface ITprGlobalNavigationSerivce
+    {
+        public List<TprHeaderMenuParentItem> GetMenuItems(int rootId);
```
sussexrick: These methods should accept a `Guid rootKey` rather than `int rootId`. 

Integer ids in Umbraco are a legacy thing, and everything also now has a GUID in a `.Key` property that is the preferred one to use.

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    .right {
+        transform: rotate(-45deg);
+        -webkit-transform: rotate(-45deg);
+        top: 27%;
+    }
+
+    .down {
+        transform: rotate(45degg);
```
sussexrick: Typo "45deg"

GovUk.Frontend.Umbraco.ExampleApp/Views/TPRHeader.cshtml
```diff
@@ -214,6 +218,35 @@
...
+                        {
+                            var linkText = item?.Content.Value<string>("linkText");
+                            var linkUrl = item?.Content.Value<Link>("linkUrl");
+                            var headerMenuChildItems = item?.Content.Value<OverridableBlockListModel>("tprHeaderMenuChildItems");
+                            <tpr-header-menu-parent-item href="@linkUrl?.Url" link-text="@linkText">
+                                @if (headerMenuChildItems != null)
+                                {
+                                    @foreach (var child in headerMenuChildItems)
+                                    {
+                                        var childLinkText = child?.Content.Value<string>("linkText");
```
sussexrick: Any new properties you add should have a constant in `TprPropertyAliases.cs` and you should use the content here. The same is true for any new document types, which should have the alias in `TprElementTypeAliases.cs`

ThePensionsRegulator.Frontend.Umbraco/Services/ITprGlobalNavigationSerivce.cs
```diff
@@ -0,0 +1,13 @@
...
+        public List<TprHeaderMenuParentItem> AddParentMenuItem(int rootId, int placement, TprHeaderMenuParentItem parentMenuItem);
```
sussexrick: You could return `IList<T>` rather than `List<T>` from these interface methods to add a little more flexibility.

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
sussexrick, line 1: If there are no child items it still creates and expands a box for them on the mega-menu

Image

ThePensionsRegulator.Frontend.Umbraco/Services/TprGlobalNavigationService.cs
```diff
@@ -0,0 +1,81 @@
...
+                        .ToList()
+                })
+                .ToList();
+
+            return menuItems;
+        }
+
+        public List<TprHeaderMenuParentItem> AddParentMenuItem(int rootId, int placement, TprHeaderMenuParentItem parentItem)
+        {
+            var navigation = GetMenuItems(rootId);
```
sussexrick: Visual Studio indicates that `AddParentMenuItem` and `AddChildMenuItem` are not used. Perhaps they're leftovers from an earlier iteration and can be removed? I can see that they might be provided as an API for consuming applications to call, but if so then that needs to be included in your docs.

If needed then you're potentially calling `GetMenuItems()` many times here, with each call to `AddParentMenuItem` or `AddChildMenuItem`. You only need to call it once, and calling it many times would be inefficient especially if it's hitting the database.

Internally you could have a method something like `EnsureMenuItems()` that checks whether you already have them, and gets them if not. That's the one you'd call from `AddParentMenuItem` or `AddChildMenuItem`. It could use an instance of `Lazy<TprHeaderMenuParentItem>`, but it doesn't have to.

ThePensionsRegulator.Frontend.Umbraco/Services/TprGlobalNavigationService.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            if (parentItem != null)
+            {
+                navigation.Insert(placement, parentItem);
+            }
+
+            return navigation;
+        }
+
+        public List<TprHeaderMenuParentItem> AddChildMenuItem(int rootId, int placement, int hierarchy, TprHeaderMenuChildItem tprMobileMenuChildItem)
```
sussexrick: Visual Studio indicates that `AddParentMenuItem` and `AddChildMenuItem` are not used. Perhaps they're leftovers from an earlier iteration and can be removed? I can see that they might be provided as an API for consuming applications to call, but if so then that needs to be included in your docs.

If needed then your parameter naming is inconsistent here. `AddParentMenuItem` has the last parameter named `parentItem`, so this should be `childItem`.

I'd suggest also renaming `placement` to `index` just to be more consistent with .NET in general.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/settings.config
```diff
@@ -28,6 +28,79 @@
...
+      <ValidationRegExpMessage></ValidationRegExpMessage>
+      <LabelOnTop>false</LabelOnTop>
+    </GenericProperty>
+  </GenericProperties>
+  <Tabs>
+    <Tab>
+      <Key>b84aa149-2f37-415b-a6d5-5cec81ac766b</Key>
+      <Caption>TPR header menu</Caption>
+      <Alias>tprHeaderMenu</Alias>
+      <Type>Group</Type>
```
sussexrick: Rather than putting these properties directly on the `Settings` document type, create a composition and put them there. Then add that composition to the 'Settings' document type. Note that this is how others have been done.

The reason for this is that this 'Settings' document type belongs to the example app. Consuming apps have their own, and a composition allows them to add all your properties easily (if they need them) rather than one-by-one.

Also you should change the sort order so that your group of fields comes after the 'TPR header' group (or becomes part of it, since the mobile menu is inseparable from the header bar).

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/settings.config
```diff
@@ -28,6 +28,79 @@
...
+      <SortOrder>0</SortOrder>
+      <Tab Alias="tprHeaderMenu">TPR header menu</Tab>
+      <Variations>Nothing</Variations>
+      <MandatoryMessage></MandatoryMessage>
+      <ValidationRegExpMessage></ValidationRegExpMessage>
+      <LabelOnTop>false</LabelOnTop>
+    </GenericProperty>
+    <GenericProperty>
+      <Key>3b2ace9e-edf0-42c9-b62f-8d87bea998bc</Key>
+      <Name>Header Menu Aria Label</Name>
```
sussexrick: These property names should be sentence case.

Also, although not part of this PR, the property for `ShowSearch` should be `Show search`.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/settings.config
```diff
@@ -28,6 +28,79 @@
...
+      <SortOrder>3</SortOrder>
+      <Tab Alias="tprHeaderMenu">TPR header menu</Tab>
+      <Variations>Nothing</Variations>
+      <MandatoryMessage></MandatoryMessage>
+      <ValidationRegExpMessage></ValidationRegExpMessage>
+      <LabelOnTop>false</LabelOnTop>
+    </GenericProperty>
+    <GenericProperty>
+      <Key>093d52d7-b535-4b1b-97a8-84c899fb2408</Key>
+      <Name>TPR header menu</Name>
```
sussexrick: A clearer name for this property would be "Menu items", since the whole group of properties is for the `TPR header menu`.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprheadermenu.config
sussexrick, line 1: I'm not sure of the function of this element type. Is it leftover from an earlier iteration?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprheadermenuparentitems.config
```diff
@@ -0,0 +1,82 @@
...
+    <Folder>TPR</Folder>
+    <Compositions />
+    <DefaultTemplate></DefaultTemplate>
+    <AllowedTemplates />
+  </Info>
+  <Structure />
+  <GenericProperties>
+    <GenericProperty>
+      <Key>7f8cf40d-a1ad-4658-88d6-2e8de1f18d57</Key>
+      <Name>Link Text</Name>
```
sussexrick: Property names should be in sentence case.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprheadermenuparentitems.config
```diff
@@ -0,0 +1,82 @@
...
+      <Alias>linkText</Alias>
```
sussexrick, line 43: Neither link text nor link URL are mandatory properties. I don't think this one should be mandatory, but if the custom link text is not supplied and an Umbraco content node is selected, it should display the `node.Name` property.

If you don't fill in link text you get a link with no `href`. I think this one needs to be a mandatory property.

docs/components/tpr-header-bar.md
sussexrick, line 1: This is good documentation - nice work. But it does also need to mentioned the requirement to include the JavaScript (via your partial view) in the page.

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderMenu.cshtml
sussexrick, line 1: You also need to include this script in `ThePensionsRegulator.Frontend.Umbraco/ApplicationBuilderExtensions.cs`

ThePensionsRegulator.Frontend.Umbraco/Services/ContentUrlProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+using Umbraco.Extensions;
+
+namespace ThePensionsRegulator.Frontend.Umbraco.Services
+{
+    public interface IContentUrlProvider
+    {
+        string GetUrl(IPublishedContent content);
+    }
+    internal class ContentUrlProvider : IContentUrlProvider
+    {
```
sussexrick: This shouldn't need to exist as Umbraco provides `IPublishedUrlProvider` - see PR 25139 in our corporate website repo.

ThePensionsRegulator.Frontend.Umbraco/Services/ContentVisibilityChecker.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Umbraco.Cms.Core.Models.PublishedContent;
+using Umbraco.Extensions;
+
+namespace ThePensionsRegulator.Frontend.Umbraco.Services
```
sussexrick: I can see you've created this so that you can avoid testing Umbraco's internal implementation of `.IsVisible()`, but you should be able to make that internal implementation work with a better mock, and remove this one.

I haven't tried it out but something like this ought to work:

```csharp
var content = UmbracoContentFactory.CreateContent<IPublishedContent>();
content.SetupUmbracoBooleanPropertyValue("umbracoNaviHide", true);
```

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -91,8 +91,31 @@ public virtual TagBuilder GenerateTprHeaderBar(TprHeaderBar tprHeaderBar)
                 var headerSearch = GenerateTprHeaderSearch(tprHeaderBar);
                 headerContent.InnerHtml.AppendHtml(headerSearch);
             }
+            if (tprHeaderBar.DisplayHeaderMenu)
+            {
+                var navDiv = new TagBuilder("div");
+                navDiv.AddCssClass("tpr-mobile-menu__container");
+                headerContent.InnerHtml.AppendHtml(navDiv);
+
+                var mobileMenu = new TprHeaderMenu
```
sussexrick, line 100: You've got a warning in DevTools about the HTML structure being not accessible:

Image

See if you can update it so that either it uses two `<button>` elements that are siblings, or one `<button>` element that changes with the open/closed state.

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
```diff
@@ -0,0 +1,339 @@
...
+
+.tpr-mobile-menu-toggle .si-glyph-fill {
+    fill: $tpr-colour-white;
+}
+
+.tpr-header-menu-nav {
+    display: none;
+}
+
+%mobile-menu-visible {
```
sussexrick: Using the % sign like this is a neat trick in SASS that I wasn't aware of, but since it's only used once you could just move this implementation to where it's used and simplify it a bit.

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
```diff
@@ -0,0 +1,339 @@
...
+        padding: 1rem;
+        display: flex;
+
+        &:hover {
+            text-decoration: underline;
+        }
+
+        &:focus {
+            background: $tpr-colour-tangerine-yellow;
+            border-bottom: 3px solid $tpr-colour-black;
```
sussexrick: This could be `$govuk-focus-width` rather than hard-coded 3px

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    .tpr-header-search.mobile .autocomplete__wrapper {
+        position: relative;
+        width: 100%;
+    }
+
+    @media (max-width: 960px) {
+        .tpr-header-menu-nav.active .tpr-header-search.mobile .tpr-header-search__button {
+            margin-left: -37px;
```
sussexrick: I wanted to test this because values in px are always worth checking - do they still work when the base font size is changed? Sometimes yes, but not always.

However, in this case I couldn't find a situation in the browser where this style shows up, so I wonder whether it's required?

I also wonder why we've got all these `.tpr-header-search` selectors in the `_tpr-header-menu.scss` file rather than `_tpr-header-search.scss`?

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
```diff
@@ -0,0 +1,339 @@
...
+        width: 100%
+    }
+
+    .navigation__sub-menu-item-title {
+        border-bottom: solid $tpr-colour-dark-grey 1px;
+        display: flex;
+    }
+}
+
+.tpr-header-menu-nav.active {
```
sussexrick: You've got a lot of CSS classes in play that do not follow the [BEM convention](https://getbem.com/introduction/). Examples include `arrow`, `right`, `navigation__sub-menu-item`, `nav-overlay`, `active`, `mobile`, `open`, `closed`. 

These are classes that might easily be defined by some other element on the page and cause this to break the other element or vice versa. All classes for the mobile menu should start with `tpr-mobile-menu`. An element inside should look something like `tpr-mobile-menu__sub-item`. A state change should look something like `tpr_mobile-menu--open`.

docs/components/tpr-header-bar.md
```diff
@@ -80,6 +87,34 @@ TPR Header Search implements the [alphagov/accessible-autocomplete](https://gith
 
 Views which require the `<tpr-header-search>` should also include the 'TPRHeaderSearchAutocomplete' partial view, in order to use the autocomplete functionality.
 
+### `<tpr-mobile-menu>`
```
sussexrick: `TPRHeaderLockup.cshtml` uses `<tpr-header-menu>` so I think this documentation may be out-of-date.

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
```diff
@@ -29,6 +30,36 @@
                 {
                     <tpr-header-search action="@Model.HeaderSearchActionPath" autocomplete-url="@Model.HeaderSearchAutocompleteUrl" placeholder="@Model.HeaderSearchPlaceholderText" aria-label="@Model.HeaderSearchAriaLabel" input-name="@Model.HeaderSearchInputName"></tpr-header-search>
                 }
+                @if (Model.ShowHeaderMenu)
```
sussexrick: I think the menu option is meant to be mutually exclusive with the `HeaderBarContent` option (which normally contains some links), so you should add a check for that to be null, similar to the one shown for the search option a few lines earlier.

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
```diff
@@ -29,6 +30,36 @@
...
+                {
+                    <tpr-header-menu>
+                        @if (Model.HeaderMenuParentItems != null)
+                        {
+                            @foreach (var item in Model.HeaderMenuParentItems)
+                            {
+                                <tpr-header-menu-parent-item href="@item?.LinkDestination" link-text="@item?.LinkText">
```
sussexrick: We shouldn't render a `<tpr-header-menu-parent-item>` if `item` is null. Ideally `item` would never be null, so explore whether you can make `Model.HeaderMenuParentItems` a `List<TprHeaderMenuParentItem>` rather than `List<TprHeaderMenuParentItem>?`. If not then use `var item in Model.HeaderMenuParentItems.OfType<TprHeaderMenuParentItem>()` in the foreach statement to eliminate any nulls.

ThePensionsRegulator.Frontend.Umbraco/Services/ITprGlobalNavigationSerivce.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections.Generic;
+using ThePensionsRegulator.Frontend.HtmlGeneration;
+
+namespace ThePensionsRegulator.Frontend.Umbraco.Services
+{
+    public interface ITprGlobalNavigationSerivce
```
sussexrick: Typo "Service"

ThePensionsRegulator.Frontend.Umbraco.Tests/Services/TprGlobalNavigationTests.cs
```diff
@@ -0,0 +1,130 @@
...
+            };
+
+            var nestedContent = mockNestedContent.Select(m => m.Object).ToList();
+            _childMocks[0].Setup(c => c.Children).Returns(nestedContent);
+
+            var mockContentAccessor = new Mock<IUmbracoContextAccessor>();
+
+            var mockContext = new Mock<IUmbracoContext>();
+
+            mockContentAccessor.Setup(x => x.TryGetUmbracoContext(out It.Ref<IUmbracoContext>.IsAny)).Callback(new TryGetUmbracoContextCallback((out IUmbracoContext context) =>
```
sussexrick: You've got some build warnings in here to resolve.

ThePensionsRegulator.Frontend.Umbraco.Tests/Services/TprGlobalNavigationTests.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            //Arrange
+            SetUp();
+
+            //Act
+            var result = _sut.GetMenuItems(_rootId);
+
+            //Assert 
+            Assert.That(result?.Count, Is.EqualTo(_rootNode.Object.Children.Count()));
+            Assert.That(result[3].LinkText, Is.EqualTo(_expectedLinkText));
```
sussexrick: What about the other results?

ThePensionsRegulator.Frontend.Umbraco.Tests/Services/TprGlobalNavigationTests.cs
```diff
@@ -0,0 +1,130 @@
...
+            Assert.That(result[3].LinkDestination, Is.EqualTo(_expectedLinkDestination));
+
+            Assert.That(result[0]?.HeaderMenuChildItems?.Count, Is.EqualTo(_childMocks[0].Object.Children.Count()));
```
sussexrick: Can we also test that these are the _right_ child items?

GovUk.Frontend.Umbraco.ExampleApp/Views/Shared/_Layout.cshtml
```diff
@@ -3,10 +3,12 @@
 @using Microsoft.AspNetCore.Localization
 @using Microsoft.Extensions.Configuration;
 @using ThePensionsRegulator.Frontend.Umbraco.Models;
+@using ThePensionsRegulator.Frontend.Umbraco.Services
 @using Umbraco.Cms.Core.Models
 @using Umbraco.Cms.Web.Common
 @inject UmbracoHelper Umbraco
 @inject IConfiguration Configuration
+@inject ITprGlobalNavigationSerivce NavigationSerivce;
```
sussexrick: Typo "Service"

(As a variable this should really be camel case `navigationService`, although I can see you've just copied the examples above...)

ThePensionsRegulator.Frontend/Models/TprHeaderLockupModel.cs
```diff
@@ -31,6 +32,11 @@ public class TprHeaderLockupModel
         public virtual string? HeaderSearchAutocompleteUrl { get; set; }
         public virtual string? HeaderSearchPlaceholderText { get; set; } 
         public virtual string? HeaderSearchAriaLabel { get; set; } 
-        public virtual string? HeaderSearchInputName { get; set; } 
+        public virtual string? HeaderSearchInputName { get; set; }
+        public virtual bool ShowHeaderMenu { get; init; }
+        public virtual string? HeaderMenuAriaLabel { get; set; }
+        public virtual string? HeaderMenuNoJSNavPage { get; set; }
+        public virtual List<TprHeaderMenuParentItem>? HeaderMenuParentItems { get; set; }
+        public virtual List<TprHeaderMenuChildItem>? ExtraChildItem { get; set; }
```
sussexrick: `ExtraChildItem` has a few problems:
- it's actually `IList` so should be "Items"
- it appears to be an API for consuming applications to add custom menu items, but that would duplicate `TprGlobalNavigationService.AddChildMenuItem()`. I think the latter is a better implementation.
- if you add an item using `ExtraChildItem` it's added to _every_ menu

sussexrick (commented):
@Max-Emery I've added a lot of comments here but I recognise it is only a draft PR and you're probably aware of a lot of them anyway. 

Don't let the number of comments give you the wrong impression - this is impressive work. It's a complex thing to implement and it already works pretty well. You're taking a good approach and are clearly on top of a lot of the detail of how this repo works.

One general comment is that it feels a little odd to say "add a parent menu item" before it's actually the parent of anything. Your naming is perfectly logical and definitely not wrong, but I think it might be more intuitive to drop the references to "parent" everywhere and use `<tpr-header-menu-item>` and `<tpr-header-menu-child-item>` (and equivalent names in other places).

ThePensionsRegulator.Frontend/HtmlGeneration/TprHeaderMenuParentItem.cs
Max-Emery: Updated to required property

ThePensionsRegulator.Frontend/HtmlGeneration/TprHeaderMenuChildItem.cs
Max-Emery: I've ended up removing this property altogether from this constructor.

ThePensionsRegulator.Frontend/HtmlGeneration/TprHeaderMenuChildItem.cs
Max-Emery: Updated to required property

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
Max-Emery: With JS enabled  default is to not show child items until expanded. Navigating through top level menu items using arrow keys is also now supported.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderMenu.cs
Max-Emery: `aria-label` moved to `nav ` element and value updated

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
Max-Emery: Added html attributes to support localization

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
Max-Emery: I've aligned all items to the left to get a more uniform look

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprheadermenu.config
Max-Emery: Yes, removed duplicate

ThePensionsRegulator.Frontend.Umbraco/Services/ContentUrlProvider.cs
Max-Emery: Updated to use `IPublishedUrlProvider` and removed wrapper class.

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
Max-Emery: Updated CSS classes to BEM convention.

ThePensionsRegulator.Frontend.Umbraco.Tests/Services/TprGlobalNavigationTests.cs
Max-Emery: Updated to check all results

ThePensionsRegulator.Frontend.Umbraco.Tests/Services/TprGlobalNavigationTests.cs
Max-Emery: Updated to check child items

ThePensionsRegulator.Frontend/TagHelpers/TprHeaderBarTagHelper.cs
Max-Emery: Added type to button, styling now coming from CSS class

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
Max-Emery: Removing the role value from the `<a>` tag has resolved this

ThePensionsRegulator.Frontend/Models/TprHeaderLockupModel.cs
Max-Emery: I've removed this implementation and going to use the methods in `TPRGlobalNavigationService`.

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
Max-Emery: I've removed the tab index setting on the arrows to use the existing implementation, JS is now updating the aria-expanded value when the sub menu is open. Media queries added to avoid other functionality leaking into mobile menu.

ThePensionsRegulator.Frontend/Views/Shared/TPR/TPRHeaderLockup.cshtml
Max-Emery: Updated

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
Max-Emery: I've removed the unused styling rule and moved `.tpr-header-search` selectors to correct file

ThePensionsRegulator.Frontend/TagHelpers/TprHeaderBarTagHelper.cs
Max-Emery: I've made updates to ensure the correct aria-expanded value is displayed depending on state.

docs/components/tpr-header-bar.md
Max-Emery: Yes, I don't think HTML support is needed. Is my set-up ok as long as we don't have `@Html.Raw()` in any views to render it?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprheadermenuparentitem.config
sussexrick: The file is still present

ThePensionsRegulator.Frontend.Umbraco.Tests/Services/TprGlobalNavigationTests.cs
sussexrick: Still 2 unresolved, and more in `TprHeaderBarTagHelper.cs`

ThePensionsRegulator.Frontend/HtmlGeneration/TprHeaderMenuParentItem.cs
sussexrick: I still think `HeaderMenuChildItems` would be better as a non-null property. If a null value is passed into the constructor, just don't update it from the default empty list.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/settings.config
sussexrick: You've moved these to a composition, which is good. However your group of fields still needs to come after the 'TPR header' group. To do that go to the document type for your composition and click the Reorder button at the top right. Set the order for the group to 12, because the 'TPR header' group is 11. However the document types for 'TPR context 1', 'TPR context 2' and 'TPR context 3' are already using 12, so you'll need to update those to 13.

Also the display names of your properties should start with a capital letter, and ARIA should be in all caps because [ARIA is an acronym](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA). ARIA is a standard with many parts so say it's an ARIA label, not just ARIA.

Properties with names in all lowercase

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprheadermenuparentitems.config
sussexrick: image

URL is an exception - that should be all caps because [URL is an initialism](https://en.wikipedia.org/wiki/URL) (ie it's short for Uniform Resource Locator).

Also 'TPR header menu child' doesn't make sense as a display name here. 'Child menu items' would make sense.

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
sussexrick: All good apart from `si-glyph-fill`.

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
sussexrick, line 1: If I add one more item to the menu, all menu items are styled incorrectly with an indent on the first line:

Incorrect menu styling

Hover styles also go wrong in this scenario. The background of line 2 overlays line 1 making it unreadable:

Unreadable menu item on hover

GovUk.Frontend.Umbraco.ExampleApp/Models/ModelsBuilder/TPrheaderMenu.generated.cs
```diff
@@ -18,30 +18,55 @@
...
+		string TprHeaderMenuAriaLabel { get; }
+
+		/// <summary>TPR header menu</summary>
+		[global::System.CodeDom.Compiler.GeneratedCodeAttribute("Umbraco.ModelsBuilder.Embedded", "13.9.2+b414456")]
+		[global::System.Diagnostics.CodeAnalysis.MaybeNull]
+		global::ThePensionsRegulator.Umbraco.Blocks.OverridableBlockListModel TprHeaderMenuBlockList { get; }
+	}
+
 	/// <summary>TPR header menu</summary>
-	[PublishedModel("tPRHeaderMenu")]
```
sussexrick, line 22: You've lost some code in here that means it's causing build errors. As it's a generated file you just need to alter it in any way so that it builds, then go into Umbraco and re-generate the models to get it right.

docs/components/tpr-header-bar.md
sussexrick: Yes, this works correctly now. You just need to update the docs to match. Don't talk about "Inner HTML value", it's just the title. Also your tags are no longer called `<tpr-mobile-menu...>`

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
sussexrick: Much better, but there's still one behaviour from the current site that isn't working. If you press down arrow to open a menu, tab to move to a child item, then up arrow to close the menu, the current implementation puts the focus back on the main menu item you just closed. Your version isn't doing that yet.

ThePensionsRegulator.Frontend/TagHelpers/TprHeaderBarTagHelper.cs
sussexrick: It is setting aria-expanded but it's the wrong way around - the menu button says it's expanded when it says "Menu" and collapsed when it says "Close".

There's still no audio feedback when you press space to open or close the menu.

Also, I would expect to be able to open the menu with down arrow and close it with up arrow. The current implementation doesn't do that, but I would expect the behaviour to be the same as with the mega menu.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprheadermenuparentitems.config
sussexrick: You've made the URL properties mandatory now, but link text is non-mandatory and has no fallback to `node.Name`. It's easy to add a link with no text and it shows up as an empty link.

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
sussexrick, line 1: You've got some purple space above and below the mobile menu toggle that I don't think is meant to be there:

Image

This is the old Sitecore version where the toggle is flush with the top and bottom of the header bar:

Image

sussexrick, line 1: If you open the mobile menu and move your mouse cursor across a menu item, the cursor changes from a pointing hand to an arrow as you go over the chevron on the right. The chevron is clickable so it should be a pointing hand.

Also it's not behaving correctly when you click it. In the existing implementation it follows the link in the main menu item. I think that's the correct behaviour as it's what the arrow looks like it will do.

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
sussexrick, line 1: 1. Start with the window big enough to see the mega menu
2. With the keyboard, tab to the first menu item (Employers) and press down arrow to open the child menu
3. Resize the window to be small enough to see the mobile menu
4. Open the mobile menu, and the child menu is still visible

Image

sussexrick, line 1: 1. Start with the window small enough to show the mobile menu
2. Click to open and close the mobile menu
3. Resize your window to be large enough to see the mega menu, then back again to smaller
4. Now the mobile menu is not clickable

ThePensionsRegulator.Frontend/Styles/_tpr-header.scss
sussexrick, line 1: On the current implementation when I open the mobile menu and I'm already within a main section, that section is highlighted with a purple background:

Image

On the new implementation that isn't happening (I've renamed the header and footer example page to 'Employers' for this test):

Image

sussexrick (dismissed):
Keep going Max, it's getting there. This is a complicated component.

Also, be sure to keep merging the develop branch into yours so you keep up-to-date, as that's moving on as well. The more often you merge, the easier it is.

ThePensionsRegulator.Frontend.Umbraco.Tests/Services/TprGlobalNavigationTests.cs
Max-Emery: Addressed build warnings in the `TprHeaderBarTagHelper.cs`. I've updated the service so the tests are no longer using this set up.

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
Max-Emery: This seemed to be caused by the styling pushing the document library menu item to the side, when adding another menu item this styling is removed.

ThePensionsRegulator.Frontend/Styles/_tpr-header.scss
Max-Emery: This is now supported

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
Max-Emery: Menu is now clickable when screen is resized

Max-Emery: Menu fully collapses when screen is resized

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
Max-Emery: Pointer CSS styling added to chevron. 

 ah so the chevron should take the user to the menu item's destination? To expand the child menu would you click on the parent menu item instead?

ThePensionsRegulator.Frontend/TagHelpers/TprHeaderBarTagHelper.cs
Max-Emery: This should be set the correct way round. 

I'm not sure if there is a better way to set the aria-labels, I've added `aria-live ="polite"` attributes to the toggle and menu items. When menu items are expanded I'm getting the child menu announced. However not very consistent with the toggle, seems to only announce the new label. 

Toggle can be done with the arrow keys, do we also want to support navigating through the menu items using the arrow keys?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/settings.config
villarin: Still unchanged - please amend @Max-Emery 
image

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprheadermenuparentitems.config
villarin: As per previous comments labels for the properties under Tpr Search Header group can be simplified

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
villarin: not resolved, please review

villarin (commented):
Please bump Directory.Build.props version to 9.0.9

villarin (approved):

Issue #493, opened 2025-08-26
Pages with left hand side nav component timeout due to infinite loop or recursive routine


villarin:
While reviewing the left hand side nav - some deeper levels in the content tree are not accessible due to left nav being unable to render.

The root cause of the issue has been identified on the TprSideNavigationLinksService - which enters an infinite loop or blocking recursive routine.

<img width="1845" height="522" alt="Image" src="https://github.com/user-attachments/assets/b509f231-f4db-463c-a46f-0f6d7e2f66f3" />

Example urls as follows:

- https://localhost:44350/side-navigation/side-navigation-child-2/side-navigation-grandchild-2/side-navigation-great-grandchild-1/
- https://localhost:44350/side-navigation/side-navigation-child-2/side-navigation-grandchild-2/side-navigation-great-grandchild-2/
- https://localhost:44350/side-navigation/side-navigation-child-2/side-navigation-grandchild-1/

Ethan-Lambourne:
Bug resolved in version 8.7.0

PR #494, opened 2025-08-27
Resolving side nav service timeout issue and other improvements

Ethan-Lambourne:
In this PR:
- Changed a while loop which was causing the timeout issue into an if statement, resolving the issue raised here: 
https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/issues/493;
- Removing redundant parameter for the `Invoke()` method;
- Including 'expanded' CSS class and editing which html elements it is assigned to in order to achieve more consistent styling;
- changing links CSS class from `govuk-link` to `govuk-link--no-visited-state`.


villarin:
@Ethan-Lambourne Looking good!

I can see the infinite recursivity issue is sorted now.

However I was wondering whether we should preserve the :visited state for links in this auxiliary nav?

It looks a bit confusing when multiple pages have been visited in the scope of the nav? 

@sussexrick feedback will be appreciated here.

Looking like this so far 

image

sussexrick:
On the question about visited links the advice from W3C/WAI is "most menus are not expected to change based on the visited state" https://www.w3.org/WAI/tutorials/menus/styling/

You can apply the existing class `govuk-link--no-visited-state`

villarin:
Thank you @sussexrick 

@Ethan-Lambourne  please apply suggested fix for the visited state re: govuk-link--no-visited-state

ThePensionsRegulator.Frontend/Styles/_tpr-side-navigation.scss
sussexrick, line 1: This new component is a little confused as to what belongs to Umbraco and what doesn't. For most TPR components we build support for plain .NET applications in `ThePensionsRegulator.Frontend`, then add any configuration required to integrate them with Umbraco in `ThePensionsRegulator.Frontend.Umbraco`.

In this case we have:
* styles and js in `ThePensionsRegulator.Frontend`, but with no component to apply them to if referencing only that project. The js is included (correctly) in the bundle in `ApplicationBuilderExtensions` in `ThePensionsRegulator.Frontend.Umbraco`.
* a view component in `ThePensionsRegulator.Frontend.Umbraco` that has no references to Umbraco except for a service injected into `TprSideNavigationComponent` that looks like it could instead be injected into the Umbraco-specific `SideNavigationLinksServiceForExampleApp`
* an interface implementation in the example app `SideNavigationLinksServiceForExampleApp` that, with a little rework, could be a useful default implementation for consuming apps to be included in `ThePensionsRegulator.Frontend.Umbraco`. It just needs a way to configure which node is used as the root of the menu then it would be re-usable. In the example app it would make more sense to pass in the 'Side navigation' page as the root rather than render a menu for the whole site, which is much too long.

sussexrick, line 1: I don't see a compelling reason why this component is implemented as a view component rather than a tag helper, and it's not explained anywhere. There's nothing wrong with view components, but a code library should take a consistent approach and this one uses tag helpers.

sussexrick, line 1: Every TPR component should have a Markdown documentation page similar to [Section cards documentation](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/docs/components/tpr-section-cards.md) and linked from README.md. That is missing for this component.

ThePensionsRegulator.Frontend/Styles/_tpr-side-navigation.scss
```diff
@@ -61,20 +61,21 @@ ul.side-nav-list {
...
-            
-    > a.govuk-link
-    {
-        font-weight:bold;
-    }
 }
 
-.side-nav-list-item-active li a::before {
+.side-nav-list-item-expanded li a::before {
     content: '\2014' ' '; /* em dash */
```
sussexrick: Although the browser copes by discarding the second value, this `content` property is invalid. There should be one value enclosed in quotes, not two.

ThePensionsRegulator.Frontend/Styles/_tpr-side-navigation.scss
sussexrick, line 1: The CSS classes should all have tpr- as a prefix, ie `.tpr-side-nav` rather than `.side-nav`. This would make it consistent with other TPR components and reduces the possibility of any clash with other packages or consuming apps defining a `.side-nav` class.

Descendant elements within the parent element should use a double underscore to separate the parent and child, eg `.tpr-side-nav__list` rather than `.side-nav-list`. Those elements in certain states should use a double hyphen, eg `.tpr-side-nav__list--expanded`. This is the [BEM naming convention](https://en.bem.info/methodology/quick-start/) used by the GOV.UK design system and therefore adopted by us.

sussexrick, line 1: The SCSS file includes a commented style that should be removed if not required:

```sass
    // @include govuk-responsive-margin(4, "top");
```

sussexrick, line 1: All GOV.UK and TPR components have a standard bottom margin. It's missing from this one. It should be defined as follows:

```sass
.tpr-side-nav {
    @include govuk-responsive-margin(6, "bottom");
}
```

sussexrick, line 1: The implementation of this component shown in the example app fails [WCAG Success Criterion 2.4.1 Bypass blocks](https://www.w3.org/WAI/WCAG21/Understanding/bypass-blocks), which is a Level A requirement and must be met.

You're partway there, in that you've used a `<nav>` element which defines a landmark, and landmarks are an acceptable way to meet the criterion. You should be able to skip to the `<main>` landmark to bypass the navigation block. However the `<nav>` is inside `<main>`, meaning this doesn't work.

sussexrick, line 1: Your use of a media query to test for `scripting: enabled` is interesting - I haven't seen that before. However, it only handles the situation where JavaScript is disabled as a feature. There are [other more likely reasons that JavaScript can fail to load](https://www.gov.uk/service-manual/technology/using-progressive-enhancement#css-fail), and your media query wouldn't catch these.

A better way to handle a no-JS scenario is to render the no-JS version by default and have JS make the changes required to support JS. In this case you'd render a class appropriate for no-JS, then have JS apply a different class on load.

sussexrick, line 1: You've got two uSync content types left behind from a previous draft, `tprsidenavigation.config` and `tprsidenavigationsettings.config`. These should both be deleted. They're both instructions to delete a previous content type, but I've checked our environments and if they were ever there they're already gone.

sussexrick, line 1: The rendered HTML repeatedly includes `<li class=" ">` . If there's no class we shouldn't render the attribute.

sussexrick, line 1: In the example app page you render the block grid in a 2/3 layout. 

By default content expects to have the the full width of `.govuk-width-container` but only render at 2/3 of that width to keep line lengths readable. This is a GOV.UK recommendation. 

In your case this is resulting in content that is two-thirds of two-thirds, which is too small. The solution is just to change the settings on each block in your grid to be full width. Full width within two thirds is actually two thirds.

sussexrick, line 1: `TprSideNavigationViewModel.cs` contains two classes. We normally put one class per file to make things easier to find.

sussexrick, line 1: There are build warnings to resolve. I noticed some in `TprSideNavigationViewModel.cs` and `SideNavigationLinksServiceForExampleApp.cs` but go though the list of warnings and resolve any related to this work.

sussexrick, line 1: You've set a margin-left of `-1.3em` to account for the em-dash character. This makes sense as both the size of the character and the width of the margin that accommodates it will change together.

However, this indent also relates to the width of that character:

```css
.side-nav-list ul {
    padding-inline-start: 20px;
}
```

Because this is set in pixels it won't change with the font size. When the base font size is set to very large the levels start to overlap:

Image

If you re-use your value of 1.3em they remain aligned:

Image

sussexrick, line 1: Why is the hierarchy indicated visually with an em-dash for the current section, but without an em-dash for other sections?

Image

Image

This suggests a different meaning which I don't think exists.

sussexrick, line 1: Using a `<nav>` element for this is good as it creates a ARIA landmark role allowing assistive tech users to move around the page. However if there is more than one `<nav>` element on the page [we should add an `aria-label` attribute to differentiate them](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/navigation_role). 

On the GOV.UK design system website they do it this way:

aria-labelledby pointing to an h2

"Pages in this section" is a good name that we can use as default text, but we should ensure it can be passed in to the component so that it can be localised, and for Umbraco we should support overriding that default text with a dictionary entry.

sussexrick, line 1: The link to the current page is not properly marked up. It should either:
 
* be a link as it is now, but with an `aria-current="page"` attribute
* not be a link, but something else like `<em aria-current="page">Name of page</em>`

In the second case `aria-current` is optional, but does better indicate the intent.
See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current

sussexrick, line 1: The behaviour here is not the same as the original, and I don't see a changed requirement documented on the work item. The original is a bit broken but the intent is clear. It uses the [disclosure pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/) to reveal the next level of hierarchy when you click on a link.

Top level of hierarchy only

Second level revealed

Third level revealed

On larger screens the new implementation just shows the entire hierarchy all of the time. This is part of the reason I've asked for design team input, to determine how we want this component to work.

sussexrick, line 1: The heading at the top of the side nav is not using a GOV.UK heading class. Adding `class="govuk-heading-m"` makes the alignment with the main content better.

Not aligned

Aligned

I'm also concerned that this doesn't really work as a heading. "Home" doesn't explain to me what the content following it actually is. If you look at the GOV.UK design system website they use "Pages in this section" as a heading (hidden from sighted users) and this makes more sense. 

Pages in this section as a hidden heading

It's even less clear on mobile view where you have a random button floating in space. It doesn't look like a button so I would expect it to be a link, and it's labelled "Home". It's also bold when the larger-screen equivalent was not. What should I expect this interface element to do?

Home button

On the GOV.UK design system website on small screens their side nav becomes part of a dropdown menu which is labelled "Menu" and has an icon suggesting that it reveals a menu. Again this makes more sense, and is part of the reason I've suggested we get input from the design team.

Menu button

sussexrick, line 1: In mobile view when the button is pressed to expand and collapse the menu, there is no indication for assistive technology about what's happening. NVDA is silent. You need to set the attribute `aria-expanded="true"` or `aria-expanded="false"` to match the changed appearance.

sussexrick (commented):
I've reviewed the whole component rather than just this PR - sorry for not picking it up earlier. Otto asked me to put my review comments on this PR. 

Ignore the fact that most of the comments are attached the the SCSS file - that's just because the relevant parts of the code weren't part of the PR and I had to attach the comments to something.

ThePensionsRegulator.Frontend/Styles/_tpr-side-navigation.scss
Ethan-Lambourne: During development of the side nav I was told that the service implementation should be the same as how we implemented the breadcrumb, and that was done as a view component so I followed suit.

sussexrick: Looks like that one may need a review too. It is a bit different, in that there is already a `<govuk-breadcrumbs>` tag helper and the view component is a way to provide configuration to it, so I'm not sure it's breaking the convention I'm talking about. In the side nav case there is no tag helper and the view component is the actual UI element, so it does break the convention. 

But it also needs review because there is a view at `Views/Shared/GOVUK/Breadcrumbs.cshtml` that appears to be an unused duplicate of the view component. 

Neither one is documented in the README, which links only to the base documentation for the tag helper, so this is all very confusing for a developer looking at how to implement breadcrumbs using this library. I'll create an issue to review it.

sussexrick: Created #499

villarin: I can see the component has been moved to ThePensionsRegulator.Frontend project and there is no longer an umbraco dependency in ITprSideNavigationLinksService

villarin: This was a left over from the breadcrumbs rework - so let's remove on #499

villarin: @Ethan-Lambourne this one still missing - can you add to the readme please

villarin: I can see the nav is outside of the main tag now - however I think a skip may need to be implemented here...  Moving the nav outside of the main tag is also causing some layout issues with the expand/collapse operations - so in the interest of time I have logged  #502 for follow up.

villarin: @Ethan-Lambourne  please remove these 2

villarin: That is correct. On this version, the backend is responsible for rendering and keeping the expanded/active states for current page and its ancestors - while client side code only handles expand and collapse actions when available.

villarin: Markup has been updated accordingly now

villarin: done now

villarin: Looking at https://webarchive.nationalarchives.gov.uk/ukgwa/20241030121409/https://www.thepensionsregulator.gov.uk/en/document-library/code-of-practice - it appears the dash is displayed on the current page  and expanded ancestors

villarin: As per our design review section client side navigation functions are reinstated in this PR while preserving navigation states for the non javascript functionality

villarin: @Ethan-Lambourne can you check this one again please? I still can see a few warnings related to the left nav work

villarin: As per the design review the heading has been moved to the main content section now in  1/3 - 2/3 column grid structure.

villarin (approved):

PR #500, opened 2025-08-29
Enable the selection of heading levels to be overridden by a consuming application

sussexrick:
Our work on a guide template similar to that used by GOV.UK requires the size of headings to be altered. [AB#238709](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/238709)

The current implementation of how to customise the size of headings would require content designers to manually change the heading size on each page of a guide, whereas we should be able to automate this for them.

The use of a static method to work out the heading size makes it difficult to override. This PR:

* Keeps the static method but marks it as obsolete, so that this is a non-breaking change.
* Creates and documents and new interface with an implementation the same as the static method, but which is overridable by consuming projects.
* Replaces all uses of the static method with the new interface.


AlexBeesley (dismissed):
Happy with these changes.

AlexBeesley (approved):
Approved.

PR #501, opened 2025-09-01
A summary list item value should be able to have custom CSS classes applied by a content editor AB#231506

HarryAmmon:


sussexrick (changes requested):
Happy to approve the addition of the styles dropdown, but not the removal of the 'remove format' option that was requested by content designers.

sussexrick (approved):

PR #513, opened 2025-10-13
Bumped version to 8.8.6

tpr-martyn-hills:
Added the ability to override the section card title headings level so it can be set from a h1 to a h6 so that a page can follow the correct structure for semantic design.

Updated example and documentation for the new feature.


GovUk.Frontend.ExampleApp/Views/SectionCards/Index.cshtml
```diff
@@ -19,10 +19,14 @@
         </div>
     </div>
 
-
     <div class="govuk-grid-row">
         <div class="govuk-grid-column govuk-grid-column-two-thirds-from-desktop">
             <p class="govuk-body">The section cards component displays a list of cards for navigation, usually linking to child pages.</p>
+
+            <p class="govuk-body">
+                The heading level of all the titles can be set from 1 to 6 using the <strong>card-titles-heading-level</strong> parameter.
```
sussexrick: I don't think h1 will ever be appropriate in this scenario, and probably not the lower levels either. Rather than use the existing 'GOV.UK Heading level' data type I would suggest creating a custom data type for this scenario similar to 'GOV.UK Heading level (Accordion)'.

ThePensionsRegulator.Frontend/ITprHtmlGenerator.cs
```diff
@@ -15,7 +14,7 @@ public interface ITprHtmlGenerator
         TagBuilder GenerateTprFooterBar(TprFooterBar tprFooterBar);
         TagBuilder GenerateTprContextBar(TprContextBar tprContextBar);
         TagBuilder GenerateTprRelatedLinks(TprRelatedLinks tprRelatedLinks);
-        TagBuilder GenerateTprSectionCards(TprSectionCards tprSectionCards);
+        TagBuilder GenerateTprSectionCards(int sectionCardsTitleHeadingLevel, TprSectionCards tprSectionCards);
```
sussexrick: Add a property to `TprSectionCards` rather than a new parameter on the method.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRSectionCards.cshtml
```diff
@@ -17,17 +17,25 @@
...
         descFieldName = TprPropertyAliases.SectionCardDescriptionPropertyDefault;
     }
-    
+
     var cssClasses = Model.Settings.Value<string>(GovUkPropertyAliases.CssClasses);
-    <tpr-section-cards class="@cssClasses" new-tab-text="@Umbraco.GetDictionaryValueOrDefault(DictionaryConstants.OpensInNewTab, DictionaryConstants.OpensInNewTab)">
-    
+
+    string headingLevelProperty = Model.Settings?.Value<string>(TprPropertyAliases.SectionCardsTitleHeadingLevel) ?? string.Empty;
+    var headingLevel = headingLevelProperty.Replace("heading", "", System.StringComparison.OrdinalIgnoreCase);
```
sussexrick: Very minor point as it's clearly working, but the replacement should be for "heading " (with a space) rather than "heading".

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRSectionCards.cshtml
```diff
@@ -17,17 +17,25 @@
...
+
+    if (!int.TryParse(headingLevel, out int numericHeadingLevel))
+    {
+        numericHeadingLevel = 2;
+    }
+
+    <tpr-section-cards class="@cssClasses" new-tab-text="@Umbraco.GetDictionaryValueOrDefault(DictionaryConstants.OpensInNewTab, DictionaryConstants.OpensInNewTab)" card-titles-heading-level="@numericHeadingLevel">
```
sussexrick: The appearance is different for each heading level (h3 is appearing smaller than h2 on your example page). I think you need to update the `tpr-section-card__title` CSS class to specify the same font size as is used for h2 by default.

sussexrick (changes requested):

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRSectionCards.cshtml
tpr-martyn-hills: Added the ability to set the titles heading class using the same pattern as the Search Results feature.

ThePensionsRegulator.Frontend/HtmlGeneration/TprSectionCards.cs
```diff
@@ -8,5 +8,7 @@ public record TprSectionCards
         public AttributeDictionary? Attributes { get; set; }
         public List<TprSectionCard> Cards { get; set; } = new();
         public string? NewTabText { get; set; }
+        public int sectionCardsTitleHeadingLevel = 2;
```
sussexrick: We shouldn't ever use public fields, so these should be PascalCase properties.

sussexrick (changes requested):

ThePensionsRegulator.Frontend/HtmlGeneration/TprSectionCards.cs
tpr-martyn-hills: Apologies as a typo and all updated

sussexrick (approved):

PR #521, opened 2025-10-15
Upgrade to GOV.UK Frontend v5.9.0

sussexrick:
We are currently on GOV.UK Frontend v5.8.0 and this PR is to upgrade to v5.9.0. This is a breaking change version because it involves upgrading our upstream dependency on `govuk-frontend-aspnetcore` from v2.8.1 to v3.0.1 and [breaking changes are introduced by them](https://github.com/x-govuk/govuk-frontend-aspnetcore/releases/tag/v3.0.0).

Key changes to note:

- The API provided by `govuk-frontend-aspnetcore` for processing date input fields has changed significantly, so our code is updated to work with that. End user functionality should be unaltered. Some updates were made to example pages to reflect testing. Support for month and day only remains Umbraco-only as, although the upstream library theoretically supports it, I found that it threw errors. Further upgrades will follow that may fix this.
- The [file upload component gets an upgrade originating in GOV.UK Frontend](https://github.com/alphagov/govuk-frontend/releases/tag/v5.9.0). It's now a modern JavaScript-based upload component supporting drag and drop and multiple files. In Umbraco the multiple file option is a setting which defaults to false for backwards compatibility.
- The ASP.NET example page for file uploads is updated to match the Umbraco one. Both versions are updated to demonstrate how to handle multiple file uploads with a single instance.
- The new file upload component requires GOV.UK JavaScript. This was not loading because the URL had changed, so this URL is updated in a number of places.
- We have to specify `HtmlEncoded.Default` where it was previously hardcoded, affecting many locations. This is a change in the API for `govuk-frontend-aspnetcore`.
- Several utility classes and methods from `govuk-frontend-aspnetcore` are copied into this repo because they've been removed or made internal and we still need them.
- `GovUkFrontendAspNetCoreOptions` changes to `GovukFrontendOptions` affecting many locations. This is a change in the API for `govuk-frontend-aspnetcore`.
- Both Button component example pages required examples with notification banners using reversed buttons to be removed, because the appearance of notification banners was changed (before this PR) and reversed buttons are no longer appropriate.
- In `OverridableBlockModelExtensions` I noticed (when my tests failed) two methods that were still specific to block lists which should've been updated to the interface that is shared between block lists and block grids. I made this update.

Revised file upload component (normal, hover and focus states):

File upload
File upload hover state
File upload focus state

The file upload component falls back to the previous appearance and behaviour if JavaScript is not available.


tpr-martyn-hills:
If I set the file upload not to accept multiple, then I drag and drop multiple files, the component accepts this and displays that 3 files have been selected.

image

Then I click submit and there are no errors where I would of thought it might of errored with more than one file has been selected.

image

This might not be avoidable but thought I would flag it.

sussexrick:
@tpr-martyn-hills That's a useful observation. However because this code comes from GOV.UK Frontend it's out of our control. We can report it to them if it's still true on the latest version of GOV.UK Frontend - this upgrade only goes to the next version, not the latest. Realistically in our own apps we'd use an `IFormFile` rather than `IEnumerable<IFormFile>` property in this situation, and that works but gets only the first file.

tpr-martyn-hills (dismissed):

tpr-martyn-hills (dismissed):

tpr-martyn-hills (approved):

PR #547, opened 2025-10-28
Prevent long unbreakable words like URLs and filenames from overflowing

### LINKED ISSUE #374: Success banner does not align if text is really long
The filename in the example depicted below is very long and causes the green success box to expand but the box behind it does not (see the red markers).

Example text: `UploadAssessmentOfSupportableRiskOverTheReliabilityPeriod.pdf`

Image

sussexrick:
This PR allows a long unbreakable word (such as a filename without spaces) to wrap with a Panel component, which is important as the text is white and typically overflows onto a white background if it doesn't wrap.

This screenshot shows it working for both heading and body content within a panel.

Wrapping text

Fixes #374.

This PR also removes a duplicate style for select dropdowns. This is because `_govuk-panel.scss` was recently created by copying `_govuk-select.scss` and the existing style was not removed. It still exists in `_govuk-select.scss`.

### DIFF
--- a/GovUk.Frontend.AspNetCore.Extensions/Styles/_govuk-panel.scss
+++ b/GovUk.Frontend.AspNetCore.Extensions/Styles/_govuk-panel.scss
@@ -1,11 +1,13 @@
-﻿/* 
-    If there is a fieldset level error we do not want to display an error message for the field, 
-    but without an error message the .govuk-select--error class will not be applied.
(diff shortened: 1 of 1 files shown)


samuelhaa:
Looks good to me. Works as expected.

samuelhaa (approved):

PR #551, opened 2025-11-05
Fixed bug where No results text remains displayed after clearing the form.

tpr-martyn-hills:
When using the TPR search results component, if a search returned no results, then clear button clicked, it does not clear the 'No results found' message.

image

**Error validation messaging has also been updated as per the content team requirements**

image

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,5 +1,5 @@
 <Project>
   <PropertyGroup>
-    <Version>9.0.2</Version>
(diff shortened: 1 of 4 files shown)


AlexBeesley (dismissed):

HarryAmmon (approved):
Approving but a tech debt item should be raised to make error messages and inputs configurable, both in 'vanilla' MVC application and with Umbraco.

PR #568, opened 2025-12-09
All menu items being highlighted as current section when at root

Max-Emery:
Why:

- When at the root all menu items were being highlighted as current section.

In this PR:

- Updated to avoid setting current section styling if at the root.
- Minor formatting updates.


villarin (commented):
Please bump release version

villarin (approved):

PR #574, opened 2025-12-17
Date inputs can be configured to show the year

HarryAmmon:
Adds a tag helper and Umbraco configuration to hide the 'year' input box as part of a date input. This extends the work already done to hide the 'day' field. 

Includes some changes to the model binder, as there were a few bugs for when a year was not provided.

Defaults the year to 1900 after discussions with the team. I could provide a way to configure the default year, but that could come as an enhancement of this work.

### DIFF
--- a/GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukdateinputsettings.config
+++ b/GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukdateinputsettings.config
@@ -48,6 +48,22 @@
       <ValidationRegExpMessage></ValidationRegExpMessage>
       <LabelOnTop>false</LabelOnTop>
     </GenericProperty>
+    <GenericProperty>
+      <Key>886b5aae-aad0-493c-aba8-15eb548cb019</Key>
+      <Name>Show the year field</Name>
+      <Alias>showYear</Alias>
+      <Definition>189de753-cac2-4307-868b-114832c68a07</Definition>
+      <Type>Umbraco.TrueFalse</Type>
+      <Mandatory>false</Mandatory>
+      <Validation></Validation>
+      <Description><![CDATA[If the year field is hidden the defaults to 1900.]]></Description>
+      <SortOrder>5</SortOrder>
+      <Tab Alias="settings">Settings</Tab>
+      <Variations>Nothing</Variations>
+      <MandatoryMessage></MandatoryMessage>
+      <ValidationRegExpMessage></ValidationRegExpMessage>
+      <LabelOnTop>false</LabelOnTop>
+    </GenericProperty>
   </GenericProperties>
   <Tabs>
     <Tab>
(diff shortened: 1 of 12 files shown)


GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukdateinputsettings.config
```diff
@@ -48,6 +48,22 @@
...
     </GenericProperty>
+    <GenericProperty>
+      <Key>886b5aae-aad0-493c-aba8-15eb548cb019</Key>
+      <Name>Show the year field</Name>
+      <Alias>showYear</Alias>
+      <Definition>189de753-cac2-4307-868b-114832c68a07</Definition>
+      <Type>Umbraco.TrueFalse</Type>
+      <Mandatory>false</Mandatory>
+      <Validation></Validation>
+      <Description><![CDATA[If the year field is hidden the defaults to 1900.]]></Description>
```
sussexrick: "the year defaults to"?

GovUk.Frontend.Umbraco/ModelBinding/UmbracoDateInputModelBinder.cs
```diff
@@ -241,8 +255,15 @@ internal static DateInputParseErrors Parse(DateInputItemTypes itemTypes, string?
...
                 {
-                    var y = yearIsValid ? parsedYear : 2000;  // 2000 is a leap year
-                    maxDaysInMonth = DateTime.DaysInMonth(y, parsedMonth);
+                    var assumedYear = 2000; // 2000 is a leap year
+                    if (yearIsValid && parsedYear != 0)
+                    {
+                        assumedYear = parsedYear;
+                    }
+                    /*var assumedYear = yearIsValid ? parsedYear : 2000; */ 
+                    // Bug is here, the year could be valid because it has not been provided, which means parsedYear will be null
```
sussexrick: Looks like some unfinished work here?

sussexrick (commented):
Seems to work well and just minor comments added. However this is an important change and the ASP.NET example app should have an example too.

HarryAmmon:
> Seems to work well and just minor comments added. However this is an important change and the ASP.NET example app should have an example too.

I have addressed your comments. 

I'm unable to add an example to the ASP.NET example app as the changes are in the `GovUk.Frontend.Umbraco` namespace, which that project doesn't have access to. The ASP.NET example app does not contain an example for 'show-day', and justification for not doing that was included in the original PR #249.

sussexrick:
Things have changed since then, as the base [`DateInputModelBinder` supports both](https://github.com/x-govuk/govuk-frontend-aspnetcore/blob/main/src/GovUk.Frontend.AspNetCore/ModelBinding/DateInputModelBinder.cs) turning off the day field that we already supported and turning off the year field that you're implementing now.

That said, you're following our current approach well so I'm happy to approve and add aligning with the base project as a tech debt issue for the future.

sussexrick (approved):

PR #575, opened 2025-12-18
Header menu improvements

Max-Emery:
Why:

- When on a child page e.g `employers/new-employers` the current section was not being highlighted in the menu.
- When a sub menu was expanded the background did not update.
- Some previous functionality not supported.

In this PR:

- Update `highlightCurrentSection` function to retrieve correct part of URL.
- Added correct styling for when menu item is focused and hovered over.
- Added support to close the sub menu using the `Esc` key for keyboard navigation.
- Ensuring previous menu item state is returned to default when user tabs through entire sub menu.
- Able to close the overlay with click in desktop view.

### DIFF
--- a/ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
+++ b/ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
@@ -217,12 +217,6 @@
             text-decoration: none;
             padding: 1rem;
         }
-
-        a:hover {
-            background: $tpr-colour-white;
-            text-decoration: underline;
-            color: $tpr-colour-black;
-        }
     }
 
     .tpr-header-menu__nav-menu-item {
@@ -295,6 +289,12 @@
         background-color: rgba(0,0,0,0.5);
         z-index: -1;
         cursor: pointer;
+    }
+
+    .tpr-header-menu__nav-menu-item:focus-within > a, .tpr-header-menu__nav-menu-item:hover > a /*.tpr-header-menu__nav-container a[aria-expanded="true"]*/ { //this is displaying without js
+        background: $tpr-colour-white;
+        text-decoration: underline;
+        color: $tpr-colour-black;
     }
 }
 
--- a/ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
+++ b/ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
@@ -291,7 +291,7 @@
         cursor: pointer;
     }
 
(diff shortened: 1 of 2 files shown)


ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
```diff
@@ -296,6 +290,12 @@
         z-index: -1;
         cursor: pointer;
     }
+
+    .tpr-header-menu__nav-menu-item:focus-within > a, .tpr-header-menu__nav-menu-item:hover > a /*.tpr-header-menu__nav-container a[aria-expanded="true"]*/ { //this is displaying without js
```
sussexrick: Was this comment an old experiment? If so it should be removed.

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
```diff
@@ -134,6 +134,12 @@ function displayDesktopOverlayOnHover() {
     });
 }
 
+function removeOverlayOnClick() {
+
+    const overlay = document.querySelector(".tpr-header-menu__nav-overlay");
```
sussexrick: I'm not sure if this is what you meant by "Able to close the overlay with click in desktop view" but if complete the following steps the overlay remains:

1. load the example page https://localhost:44350/the-pensions-regulator-tpr-header/ and scroll to the bottom
2. position my mouse below the menu
3. use the keyboard to open a submenu, thereby displaying the overlay
4. click on the overlay

I think this is a pretty unlikely real-world scenario as a mouse user is unlikely to be navigating the menu with the keyboard, but I wasn't sure what other scenario you were trying to fix.

sussexrick (commented):
This looks like a good set of improvements @Max-Emery. Please take a look at my comments but once the one about the comment is resolved I'm happy to approve.

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
max-tpr: Yes, comment has been removed

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
max-tpr: The main scenario that I was trying to fix was if you would tab forward or back though to the end (then out) of the sub menu the overlay was still displaying. I've created a new function that now restores the menu to its default state when tabbing through the entire sub menu and have removed the now redundant `removeOverlayOnClick()` function.

sussexrick (approved):

PR #589, opened 2026-01-12
Upgrade first set of backoffice previews for Umbraco 17

sussexrick:
## Context for this PR

## Scope of this PR

Migrate the first set of backoffice previews for blocks to the v17 format. All existing backoffice previews are written for AngularJS, but the biggest change between Umbraco 13 and 17 is a total rewrite of the backoffice to get rid of AngularJS, so they no longer work.

All backoffice previews are migrated, but this PR only has about a third of them to keep the PR to a size that's easier to review. This PR includes:
- Accordion
- Accordion section
- Button group
- Button
- Caption
- Details
- Error message
- Error summary
- Fieldset
- Inset text
- Link, styled as a button
- Link
- Notification banner
- Page heading
- Pagination
- Panel
- Summary card
- Summary list
- Text
- Warning text

Other blocks still appear just as a rectangle with the name of the block.

- TypeScript

```cmd
npm install
npm run build
```

This will compile the TypeScript, and then use Vite to publish the output under the `wwwroot/App_Plugins/ThePensionsRegulator.GovUk.Frontend.Umbraco` folder. Umbraco's backoffice looks for files under `wwwroot/App_Plugins` so it must go there, and `<StaticWebAssetBasePath>/</StaticWebAssetBasePath>` is required to serve these files at `/App_Plugins` rather than the default of `/_content/GovUk.Frontend.Umbraco/App_Plugins`.

`manifests.ts` is the config that connects each view to the alias of the block that it represents.

Changes to note:
- uSync data has UDIs removed. This is a change coming from Umbraco/uSync, where that form of id has been deprecated in favour of a simple GUID.
- `label` settings have gone from Angular syntax like `{{...}}` to empty strings. This is just removing obsolete code, because the label is never visible when there's a backoffice preview.
- Views are not just migrated, they are improved. I didn't keep a list, but in many cases I added support for more properties so that they better reflect the way a block is configured than the old views. I also added new views where there were none before. For example when you edit a block that has a child block list, you'll see previews in the child block list where previously you got just a list of names.

## Out-of-scope for this PR

Everything else. 

- .NET code builds
- .NET tests pass

Otherwise you can expect:
- build warnings
- outdated documentation

Also coming later: 

- Migrating remaining Umbraco backoffice views for blocks from AngularJS to web components


ZebSadiq (approved):
Tested locally. Previews work, only one suggestion:
Minor point: Pagination shows {{page}} and {{total}}, it could be improved to maybe showing example numbers in preview e.g 0 of 0 instead.

PR #590, opened 2026-01-14
Upgrade remaining GOV.UK backoffice views for v17

sussexrick:
## Context for this PR

## Scope of this PR

Migrate the second set of backoffice previews for blocks to the v17 format, following on from the first set in #589. All existing backoffice previews are written for AngularJS, but the biggest change between Umbraco 13 and 17 is a total rewrite of the backoffice to get rid of AngularJS, so they no longer work.

All backoffice previews are migrated, but this PR only has about a third of them to keep the PR to a size that's easier to review. This PR includes:
- Checkbox
- Checkboxes
- Date input
- File upload
- Hidden field
- Radio buttons
- Select
- Summary list/Summary card (adds extra views when you click into child items and actions)
- Task list
- Textarea
- Text input

- TypeScript

```cmd
npm install
npm run build
```

This will compile the TypeScript, and then use Vite to publish the output under the `wwwroot/App_Plugins/ThePensionsRegulator.GovUk.Frontend.Umbraco` folder. Umbraco's backoffice looks for files under `wwwroot/App_Plugins` so it must go there, and `<StaticWebAssetBasePath>/</StaticWebAssetBasePath>` is required to serve these files at `/App_Plugins` rather than the default of `/_content/GovUk.Frontend.Umbraco/App_Plugins`.

`manifests.ts` is the config that connects each view to the alias of the block that it represents.

Changes to note:
- `label` settings have gone from Angular syntax like `{{...}}` to empty strings. This is just removing obsolete code, because the label is never visible when there's a backoffice preview.
- Views are not just migrated, they are improved. I didn't keep a list, but in many cases I added support for more properties so that they better reflect the way a block is configured than the old views. I also added new views where there were none before. For example when you edit a block that has a child block list, you'll see previews in the child block list where previously you got just a list of names.

## Out-of-scope for this PR

Everything else. 

- .NET code builds
- .NET tests pass

Otherwise you can expect:
- TPR blocks still appear just as a rectangle with the name of the block
- build warnings
- outdated documentation

Also coming later: 


ZebSadiq:
There is a contrast issue with the File upload component, when the user hovers over the button after clicking it.

image

ZebSadiq:
When I visit the (Text input) or (Textarea) pages, I don't see the example content added in the back office. It works fine once we add the block to the page. 

image
image

sussexrick:
> There is a contrast issue with the File upload component, when the user hovers over the button after clicking it.
> 
> image

Good spot. Fixed in #592 and #593.

sussexrick:
> When I visit the (Text input) or (Textarea) pages, I don't see the example content added in the back office. It works fine once we add the block to the page.
> 
> image image

I've seen this before and it's because a new install has US English, and those pages are multi-lingual and have no US English content. If you delete the US English language in the settings section it resolves the root cause, but the tree is cached. Running `npm run build` is one way to clear the cache. Restarting _may_ be another.

At the end of this series of PRs I update the docs for running the example app to include an instruction to delete US English: https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/feature/v17/docs/umbraco/run-example-application.md

ZebSadiq (approved):

ZebSadiq:
> > When I visit the (Text input) or (Textarea) pages, I don't see the example content added in the back office. It works fine once we add the block to the page.
> > image image
> 
> I've seen this before and it's because a new install has US English, and those pages are multi-lingual and have no US English content. If you delete the US English language in the settings section it resolves the root cause, but the tree is cached. Running `npm run build` is one way to clear the cache. Restarting _may_ be another.
> 
> At the end of this series of PRs I update the docs for running the example app to include an instruction to delete US English: https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/feature/v17/docs/umbraco/run-example-application.md

Yes, the US English was the problem. Tested locally.

ZebSadiq:
> > There is a contrast issue with the File upload component, when the user hovers over the button after clicking it.
> > image
> 
> Good spot. Fixed in #592 and #593.

Both of these have been approved.

ZebSadiq:
>

PR #594, opened 2026-01-14
Refactor GOV.UK backoffice views

sussexrick:
## Context for this PR

## Scope of this PR

Backoffice views for GOV.UK components have already been migrated to Umbraco 17, but this does another sweep over all of them to pick up a few issues:

- More `label` settings have gone from Angular syntax like `{{...}}` to empty strings. This is just removing obsolete code, because the label is never visible when there's a backoffice preview.
- Views use built-in Umbraco TypeScript types where possible rather than custom ones.
- Restoring the v13 feature where you can click anywhere in a preview to edit the block content, rather than just on the tiny edit icon.

- TypeScript

```cmd
npm install
npm run build
```

This will compile the TypeScript, and then use Vite to publish the output under the `wwwroot/App_Plugins/ThePensionsRegulator.GovUk.Frontend.Umbraco` folder.

## Out-of-scope for this PR

Everything else. 

- .NET code builds
- .NET tests pass

Otherwise you can expect:
- TPR blocks still appear just as a rectangle with the name of the block
- build warnings
- outdated documentation

Also coming later: 

### DIFF
--- a/GovUk.Frontend.Umbraco.ExampleApp/uSync/v17/DataTypes/GOVUKBlockGrid.config
+++ b/GovUk.Frontend.Umbraco.ExampleApp/uSync/v17/DataTypes/GOVUKBlockGrid.config
@@ -420,7 +420,7 @@
       "settingsElementTypeKey": "b4bca409-19d8-460d-9155-6b84fc90e820",
       "view": "~/App_Plugins/GOVUK/blocks/views/GovUkSelectPreview.html",
       "stylesheet": "~/css/govuk-umbraco-backoffice.css",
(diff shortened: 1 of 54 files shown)


ZebSadiq:
Things generally work fine. I have an observation / suggestion and wondering what you think.

The details component shows all of the sections minimised in edit mode. The user has to click on the component to see the text. Would it be more useful to the editors to see all of the sections expanded instead?
image

ZebSadiq (approved):

sussexrick:
Good suggestion. As you've approved I'll merge now, and include that change in the next PR.

PR #598, opened 2026-01-19
Add address lookup component for v13

HarryAmmon:
Adds a TPR address lookup component. This is the first of a series of PRs to build this component, with the aim of this component being delivered iterativley. 

This PR creates the component, which adds a fieldset with the CSS class `tpr-address-lookup` so it can be targetted by JavaScript in the next PR in this series. 

It adds Umbraco support, adding it as a block for TPR Block Grid and TPR Block List.

This PR also bumps the version of this library to `9.2.0--beta001` so that future PRs can increase the `beta` version, with the final PR being an offical release.

Once this component has been built for Umbraco 13, a version will be created to support Umbraco 17


GovUk.Frontend.ExampleApp/Models/AddressLookupViewModel.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace GovUk.Frontend.ExampleApp.Models
+{
+    public class AddressLookupViewModel
+    {
+        public string? AddressLine1 { get; set; }
```
sussexrick: You _could_ use naming that fits with [BS7666](https://www.agi.org.uk/bs-7666-guidelines/), which is the UK's address standard. Your properties would convert into:

AddressableObjectName
StreetReference
Town
AdministrativeArea
Country
Postcode

That said, if our OS datasource doesn't provide addresses in BS7666 then it's just playing lip service to the standard. Even the above is not all the BS7666 fields. Addresses are messy. You'll never consistently get the right data in the right fields, which is why people do "line 1", "line 2" etc.

GovUk.Frontend.Umbraco.ExampleApp/Models/AddressLookupViewModel.cs
```diff
@@ -0,0 +1,28 @@
...
+        [Required(ErrorMessage = nameof(AddressLine1))]
+        public string? AddressLine1 { get; set; }
+
+        public string? AddressLine2 { get; set; }
+
+        [Required(ErrorMessage = nameof(Town))]
+        public string? Town { get; set; }
+
+        [Required(ErrorMessage = nameof(County))]
+        public string? County { get; set; }
```
sussexrick: Why would we require county? The Post Office no longer use it and it's optional almost everywhere, including on the GOV.UK Design System example.

GOV.UK advice says "Remove the county field if you’re sure your users will not need it, and your service will not use it."

GovUk.Frontend.ExampleApp/Models/AddressLookupViewModel.cs
sussexrick: **Update:** I see now your naming reflects that on the GOV.UK example. I would agree you should probably follow that unless the OS data is actually returning BS7666 data. However TPR does have a data standards catalogue which will mandate a preferred address format, so check that. Whatever you do also needs to be easily translatable into the database you're writing to... addresses are messy.

GovUk.Frontend.Umbraco.ExampleApp/Models/AddressLookupViewModel.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public string? Country { get; set; }
+
+        [Required(ErrorMessage = nameof(Postcode))]
+        [UkPostcode(ErrorMessage = nameof(Postcode))]
+        public string? Postcode { get; set; }
```
sussexrick: You can have addresses without postcodes. Electricity substations for example. TPR are probably only ever going to need office addresses though, so it would only be the very rare edge case of a brand new building that hasn't yet been assigned a postcode. Just FYI if you want to ask the question. I think your implementation is fine.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tpraddresslookupsettings.config
```diff
@@ -0,0 +1,27 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<ContentType Key="f80f2e52-4a45-4dd8-8e73-824015296fd5" Alias="tprAddressLookupSettings" Level="2">
+  <Info>
+    <Name>Address lookup settings</Name>
+    <Icon>icon-circus color-black</Icon>
+    <Thumbnail>folder.png</Thumbnail>
+    <Description>Settings for a TPR component to lookup an address. Requires JavaSript.</Description>
```
sussexrick: typo - JavaScript. 

More importantly, what is the intended behavior for a non-JS user? Will they type in their address, or get a clunkier but functional lookup? This affects the naming of the component. Is it 'Address lookup' or just 'Address'?

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
```diff
@@ -0,0 +1,36 @@
...
+@using Umbraco.Cms.Core.Models
+@using Umbraco.Extensions
+@using Umbraco.Cms.Core.Models.PublishedContent
+@using Umbraco.Cms.Web.Common
+@using Microsoft.Extensions.Configuration
+@using PropertyAliases = GovUk.Frontend.Umbraco.PropertyAliases
+@addTagHelper *, ThePensionsRegulator.Frontend
+
+@{
+    var legend = Model.Content.Value<string>(PropertyAliases.Legend)?.Replace("{{name}}", Umbraco.AssignedContentItem.Name);
```
sussexrick, line 24: You should support "legend is page heading" as an option (which is already on other components). I imagine it would be common for this component to be the "one thing per page".

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
```diff
@@ -0,0 +1,36 @@
...
+	var cssClasses = (" " + Model.Settings.Value<string>(PropertyAliases.CssClasses)).TrimEnd();
+	var blocks = Model.Content.Value<OverridableBlockListModel>(TprPropertyAliases.AddressLookupBlocks);
+}
+
+<div class="@cssClasses">
+    <tpr-address-lookup>
```
sussexrick: You should be able to apply the CSS classes to `<tpr-address-lookup>` rather than adding an extra div.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
```diff
@@ -0,0 +1,36 @@
...
+        <tpr-address-lookup-legend>@legend</tpr-address-lookup-legend>
+        @if (blocks is not null && blocks.Any())
+        {
+            var childBlocks = new BlockListViewModel(blocks)
+            {
+                RenderGrid = false,
+                RenderWidthContainer = false
+            };
+            <div>
```
sussexrick: Why the extra div?

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
```diff
@@ -0,0 +1,56 @@
...
+                    <govuk-input-label>Address line 2</govuk-input-label>
+                </govuk-input>
+                <govuk-input for="TownOrCity">
+                    <govuk-input-label>Town or City</govuk-input-label>
+                </govuk-input>
+                <govuk-input for="Region">
+                    <govuk-input-label>Region</govuk-input-label>
+                </govuk-input>
+                <govuk-input for="Postcode">
+                    <govuk-input-label>Postal Code</govuk-input-label>
```
sussexrick: Postcode is the more recognisable term

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
```diff
@@ -0,0 +1,56 @@
...
+            <tpr-address-lookup>
+                <tpr-address-lookup-legend>Enter new address</tpr-address-lookup-legend>
+                <govuk-input for="AddressLine1">
+                    <govuk-input-label>Address line 1</govuk-input-label>
+                </govuk-input>
+                <govuk-input for="AddressLine2">
+                    <govuk-input-label>Address line 2</govuk-input-label>
+                </govuk-input>
+                <govuk-input for="TownOrCity">
+                    <govuk-input-label>Town or City</govuk-input-label>
```
sussexrick: TPR house style is sentence case, "Town or city"

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
```diff
@@ -0,0 +1,56 @@
...
+        </p>
+
+        <p class="govuk-body">
+            Any child content (such as <code>govuk-input</code> components) will be rendered inside the fieldset after the legend.
+        </p>
+
+        <form asp-controller="AddressLookup" asp-action="Post" method="post" novalidate>
+            <tpr-address-lookup>
+                <tpr-address-lookup-legend>Enter new address</tpr-address-lookup-legend>
+                <govuk-input for="AddressLine1">
```
sussexrick: You should set the widths of fields and `autocomplete` attributes to match the expected input - see the first example on https://design-system.service.gov.uk/patterns/addresses/

GovUk.Frontend.ExampleApp/Models/AddressLookupViewModel.cs
```diff
@@ -0,0 +1,12 @@
...
+        public string? AddressLine2 { get; set; }
+        public string? TownOrCity { get; set; }
+        public string? Region { get; set; }
+        public string? Country { get; set; }
+        public string? Postcode { get; set; }
+    }
```
sussexrick, line 48: Examples in both apps should be equivalent wherever possible. You have no validation here, but you do in the other app.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/TPRAddressLookupBlockList.config
sussexrick, line 1: I think using a block list for this is a bit too generic.  You're asking the user to rebuild the same set of fields consistently every time including width classes and autocomplete attributes... and the latter aren't even supported yet on the 'Text input' block. It also gives you help text that doesn't make sense in this context:

Postcode field offering to use the page name as the label

I think it would be better to hard-code fields with all the config you need. Your set of fields is always going to be the same. You can still re-use the `<govuk-input>` tag helper, but don't ask the user to do all the work to configure it.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/address-lookup.config
```diff
@@ -0,0 +1,361 @@
...
+          {
+            "contentTypeKey": "fe2dd0de-55a9-4249-be03-6896aa719974",
+            "hint": "",
+            "label": "Town or city",
+            "udi": "umb://element/4b2eff39f72745bdb6fa02dc0521b573"
+          },
+          {
+            "contentTypeKey": "fe2dd0de-55a9-4249-be03-6896aa719974",
+            "hint": "",
+            "label": "Province/region/state",
```
sussexrick: I think you mean county here. 

Be clear that that is a UK address lookup, not an international one. This affects the naming of the component - it's not really 'Address lookup' it's 'UK address lookup'.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/address-lookup.config
```diff
@@ -0,0 +1,361 @@
...
+            "udi": "umb://element/3383e44d8618424f914b1e756d8aa4db"
+          },
+          {
+            "contentTypeKey": "fe2dd0de-55a9-4249-be03-6896aa719974",
+            "hint": "",
+            "label": "Country",
```
sussexrick: What are we expecting people to put in for country? England, Scotland, Wales, UK, U.K., Britain, Great Britain, United Kingdom... a text field will give you messy data.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tpraddresslookupsettings.config
HarryAmmon: A non-JS user would be expected to enter their address line by line. There is discussion about the 'Country' field being a select box. We would want the last to come from a data source, as this changes surprsingly frequently.

As for the name of this component, I think it should be 'UK Address lookup' as that is what the majority of users will use it for. The JS version of this allows a user to enter an international address, or a UK address if they are unable to find theirs if they are yet to be assigned a postcode.

HarryAmmon:
# Changes since last PR
- Adds support of legend as page heading to follow the existing pattern for fieldsets.
- Address lookup umbraco block has pre-defined inputs, each with a model property picker, and relevant error messages
- Adds `AddressLookupStateHelper` to reduce code duplication and improve testability of the Address lookup block partial view
- Adds `AddressLookupValidationMetadataProvider` for the address lookup block, allowing us to retrieve the correct error message for view model property and it's validation attribute
- Adds examples of using two instances of the component on one page, as this is the pattern used in TPR prototypes so far

@sussexrick I belive I have addressed all your comments since your last review

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
sussexrick: I can see you've updated the Umbraco version but not the plain ASP.NET version, so that example is now not a best-practice one for consumers to copy.

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
sussexrick: I can see you've updated the Umbraco version but not the plain ASP.NET version, so that example is now not a best-practice one for consumers to copy.

ThePensionsRegulator.Frontend.Umbraco/App_Plugins/ThePensionsRegulator.Frontend.Umbraco/blocks/views/TPRAddressLookupPreview.html
```diff
@@ -0,0 +1,8 @@
+﻿<div class="govuk-form-group backoffice-block-view" ng-click="block.edit()" ng-focus="block.focus" aria-label="Edit address lookup component">
+    <fieldset class="govuk-fieldset tpr-address-lookup" aria-hidden="true">
+        <legend class="govuk-fieldset__legend govuk-fieldset__legend--for-field" ng-bind-html="block.data.legend"></legend>
```
sussexrick: This class should be `govuk-fieldset__legend--for-fieldset`

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
```diff
@@ -0,0 +1,167 @@
...
+@using Microsoft.Extensions.Configuration
+@using PropertyAliases = GovUk.Frontend.Umbraco.PropertyAliases
+@addTagHelper *, ThePensionsRegulator.Frontend
+@addTagHelper *, GovUk.Frontend.AspNetCore
+@addTagHelper *, GovUk.Frontend.AspNetCore.Extensions
+
+@{
+    var legend = Model.Content.Value<string>(PropertyAliases.Legend)?.Replace("{{name}}", Umbraco.AssignedContentItem.Name);
+    var legendIsPageHeading = Model.Settings.Value<bool>(PropertyAliases.FieldsetLegendIsPageHeading);
+    var legendClass = legendIsPageHeading ? headingClassProvider.HeadingClasses(Umbraco.AssignedContentItem).LegendAsHeading1 : "govuk-fieldset__legend--for-field";
```
sussexrick: This class should be `govuk-fieldset__legend--for-fieldset`

GovUk.Frontend.Umbraco.ExampleApp/Models/AddressLookupViewModel.cs
sussexrick: I'm resolving the comment as I can see county is now optional, but I would suggest adding a switch to hide the county field altogether for services that want to follow GOV.UK guidance.

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
```diff
@@ -0,0 +1,75 @@
...
+                    <govuk-input-label>Address line 1</govuk-input-label>
+                </govuk-input>
+                <govuk-input for="ShippingAddressLine2">
+                    <govuk-input-label>Address line 2</govuk-input-label>
+                </govuk-input>
+                <govuk-input for="ShippingTownOrCity">
+                    <govuk-input-label>Town or city</govuk-input-label>
+                </govuk-input>
+                <govuk-input for="ShippingCounty">
+                    <govuk-input-label>County</govuk-input-label>
```
sussexrick: All optional fields should have an ` (optional)` suffix as part of the label.

This will take some thought for the Umbraco implementation as it could potentially affect the backoffice view and the 'Defaults to ... if left blank' description as well as the main view, but as it stands the decision of whether a field is optional is up to the consuming application's view model. 

Just a suggestion - perhaps it would be helpful to include a view model with recommended validators in `ThePensionsRegulator.Frontend`? That wouldn't necessarily work for your use case with two instances on a page, but I think it would work for others.

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
```diff
@@ -0,0 +1,75 @@
...
+            The legend will be rendered as a proper <code>&lt;legend&gt;</code> element with the <code>govuk-fieldset__legend</code> class applied.
+        </p>
+
+        <p class="govuk-body">
+            Any child content (such as <code>govuk-input</code> components) will be rendered inside the fieldset after the legend.
+        </p>
+
+        <form asp-controller="AddressLookup" asp-action="Post" method="post" novalidate>
+            <tpr-address-lookup>
+                <tpr-address-lookup-legend>Enter shipping address</tpr-address-lookup-legend>
```
sussexrick, line 31: These legends should have `govuk-fieldset__legend--for-fieldset` applied. 

You could potentially apply this in the tag helper whenever 'legend is page heading' is false, so that it doesn't get missed. The pattern in other similar components it to populate it from outside, as you're doing in the Umbraco version, but those others are probably all GOV.UK component tag helpers that shouldn't embed a TPR class, whereas this is a TPR component.

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
```diff
@@ -0,0 +1,75 @@
...
+                <govuk-input for="ShippingAddressLine1">
+                    <govuk-input-label>Address line 1</govuk-input-label>
+                </govuk-input>
+                <govuk-input for="ShippingAddressLine2">
```
sussexrick: You're not demonstrating client-side validation in this example, but you are in the Umbraco one.

ThePensionsRegulator.Frontend.Umbraco/Validation/AddressLookupValidationMetadataProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+            _umbracoContextAccessor.TryGetUmbracoContext(out var umbracoContext);
+            if (umbracoContext is null) { return; }
+            if (umbracoContext.PublishedRequest?.PublishedContent is null) { return; }
+
+            var validationAttributes = context.ValidationMetadata.ValidatorMetadata.OfType<ValidationAttribute>().Where(x => !string.IsNullOrEmpty(x.ErrorMessage)).ToList();
+            if (!validationAttributes.Any()) { return; }
+
+            var addressLookupBlocks = umbracoContext.PublishedRequest.PublishedContent
+                .FindOverridableBlockModels(_publishedValueFallback)
+                .FindBlocksByContentTypeAlias("tprAddressLookup", _publishedValueFallback);
```
sussexrick: We put all content type aliases as constants in `TprElementTypeAliases.cs`, so you should do that and use the constant here.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
```diff
@@ -0,0 +1,167 @@
...
+    var hintText = Model.Content.Value<IHtmlEncodedString?>(PropertyAliases.Hint);
+    var hintTextBlocks = Model.Content.Value<OverridableBlockListModel>("fieldsetBlocks");
```
sussexrick: This alias should use a constant from `PropertyAliases` or `TprPropertyAliases` (depending on whether it's from a GOV.UK composition or defined here...)

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
HarryAmmon: I will update the example applications to show the best practice for this and align it with the [GDS example](https://design-system.service.gov.uk/patterns/addresses/).

I can update the umbraco description to 'Defaults to ... (optional) if left blank' for fields that should be optional. If a consuming service wants to change the label for these inputs it should be up to them to ensure they add '(optional)' suffix.

As for your suggestion, I think this would be a good idea if the [ModelPropertyController](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/v13/GovUk.Frontend.Umbraco/Validation/ModelPropertyController.cs) could find properties of a property where the type is a class. This library could provide the consuming service with the correct properties and attributes. This would be used straight away by the scheme return application. I do not think this should be done as part of this PR, if we want to do that it should be done as part of the PR that updates `ModelPropertyController`.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/address-lookup.config
HarryAmmon: For the first iteration of this component we are happy with a free text field which the consuming application will validate for actual countries using whatever list that service has. Longer term we want this to be a select component populated with countries from a source.

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
sussexrick: Only the first example on the page is updated for the ASP.NET version.

Thank you for the reminder on the use-case. I remembered it came from your team but not what it was. The first part of this is [implemented on `develop` for v17](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/pull/601) and there's no reason it couldn't be back-ported. However it's probably not much use without also updating the model binding code, which is still to-do. If you could write up your use case as a Github issue that would help.

ThePensionsRegulator.Frontend.Umbraco/Validation/AddressLookupValidationMetadataProvider.cs
sussexrick: There should be an entry in that class for the new settings element type as well.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
sussexrick: Wrong class - you missed the double hyphen in the tag helper

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
sussexrick: Wrong class - you missed the double hyphen in the tag helper

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/TPRAddressLookupBlockList.config
sussexrick, line 1: I think this file is now redundant. The backoffice shows the data type is not referenced.

sussexrick (changes requested):

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/TPRAddressLookupBlockList.config
HarryAmmon: I have deleted the file rather than have a uSync file with a delete/remove command, you will need to do a fresh install of Umbraco to see that is has been removed from the backoffice.

sussexrick (dismissed):

HarryAmmon:
Intentionally not merging the PR until the JS version of this is ready so I don't block the `v13` branch from publishing. v`9.3.0--beta001` will be published from this branch

GovUk.Frontend.ExampleApp/Models/AddressLookupViewModel.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        [MaxLength(500, ErrorMessage = "Shipping county must not exceed 500 characters")]
+        public string? ShippingCounty { get; set; }
+
+        [Required(ErrorMessage = "Shipping country is required")]
+        public string? ShippingCountry { get; set; }
+
+        [Required(ErrorMessage = "Shipping postcode is required")]
+        [MaxLength(20, ErrorMessage = "Shipping postcode must not exceed 20 characters")]
+        public string? ShippingPostcode { get; set; }
```
sussexrick, line 26: This feels like a missed opportunity (and in the Umbraco version) to use the [[UkPostcode] validation attribute](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/docs/components/text-input.md#validate-uk-postcodes).

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tpraddresslookup.config
```diff
@@ -0,0 +1,178 @@
...
+      <SortOrder>6</SortOrder>
+      <Tab Alias="content">Content</Tab>
+      <Variations>Nothing</Variations>
+      <MandatoryMessage></MandatoryMessage>
+      <ValidationRegExpMessage></ValidationRegExpMessage>
+      <LabelOnTop>false</LabelOnTop>
+    </GenericProperty>
+    <GenericProperty>
+      <Key>b78513ce-be31-41f1-a822-8ee61640f3e9</Key>
+      <Name>fieldsetBlocks</Name>
```
sussexrick: This display name should just be 'Blocks' to be consistent with other components.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
```diff
@@ -0,0 +1,167 @@
...
+            RenderGrid = false,
+            RenderWidthContainer = false
+        };
+        <div id="@addressLine1ModelProperty-blocks">
+            @await Html.PartialAsync("GOVUK/BlockList", childBlocks)
+        </div>
+    }
+    @if (!string.IsNullOrEmpty(hintText?.ToString()))
+    {
+            <div class="govuk-hint">@hintText</div>
```
sussexrick, line 96: Non-blocking comment as this is working - for other similar tag helpers, there would be a child tag helper for the hint.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    <govuk-client-side-validation 
+        error-message-required="@(Model.Settings.Value<string>(TprPropertyAliases.AddressLine1ErrorMessageRequired))"
+        error-message-maxlength="@(Model.Settings.Value<string>(TprPropertyAliases.AddressLine1ErrorMessageMaxLength))">
+        <govuk-input name=@addressLine1State.ModelPropertyName value="@addressLine1State.AttemptedValue" input-aria-invalid="@addressLine1State.InvalidAriaLabel" autocomplete="address-line1">
+            <govuk-input-label>@addressLine1State.Label</govuk-input-label>
+            @if (addressLine1State.HasErrorMessage)
+            {
+                <govuk-input-error-message>@addressLine1State.ErrorMessage</govuk-input-error-message>
```
sussexrick, line 106: When I turned off JS in Umbraco and submitted an invalid form, the only field that repopulated with the value I'd entered was country. The rest lost what I'd typed. This was not a problem in the plain ASP.NET version.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
```diff
@@ -0,0 +1,167 @@
...
+
+    var countryState = addressFieldStateHelper.GetFieldState(
+            Model.Settings.Value<string>(TprPropertyAliases.CountryModelProperty),
+            Model.Content.Value<string?>(TprPropertyAliases.CountryLabel),
+            "Country",
+            ViewContext.ModelState
+		);
+}
+
+<tpr-address-lookup class="@cssClasses" described-by="@fieldsetDescribedBy">
```
sussexrick: This `described-by` attribute renders as `described-by`, which I'm not familiar with as a valid HTML attribute. I think you meant `aria-describedby`?

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
```diff
@@ -0,0 +1,167 @@
...
+    var cssClasses = (" " + Model.Settings.Value<string>(PropertyAliases.CssClasses)).TrimEnd();
+    if (hintTextBlocks is not null && hintTextBlocks.Any())
+    {
+        fieldsetDescribedBy = $"{addressLine1ModelProperty}-blocks {fieldsetDescribedBy}".TrimEnd();
+    }
+
+    var addressLine1State = addressFieldStateHelper.GetFieldState(
+            Model.Settings.Value<string>(TprPropertyAliases.AddressLine1ModelProperty),
+            Model.Content.Value<string?>(TprPropertyAliases.AddressLine1Label),
+            "Address line 1",
```
sussexrick, line 40: Non-blocking suggestion - in some other components we have default text coming from a dictionary entry, with the hard-coded default text as the fallback if that is missing.

ThePensionsRegulator.Frontend/TagHelpers/TprAddressLookupContext.cs
```diff
@@ -0,0 +1,35 @@
...
+            }
+
+            _legend = (isPageHeading, attributes, content);
+        }
+
+        public void ThrowIfNotComplete()
+        {
+            if (Legend is null)
+            {
+                throw ExceptionHelper.AChildElementMustBeProvided(TprAddressLookupLegendTagHelper.TagName);
```
sussexrick, line 33: This is the correct thing to do as legend should be required, but the text on the ASP.NET example page says legend is optional. You may want to add a `string.isNullOrWhitespace` check too as you can include the legend tag helper without any content and get an empty legend.

sussexrick (changes requested):

GovUk.Frontend.ExampleApp/Models/AddressLookupViewModel.cs
HarryAmmon: The non-JS version of this component must support international addresses. Using `UkPostcode` would prevent us from doing that

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
HarryAmmon: Will address as part of the part 2 PR

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
HarryAmmon: Will address as part of the part 2 PR

sussexrick (approved):

PR #603, opened 2026-01-23
Restore rich text editor support for v17

sussexrick:
## Context for this PR

## Scope of this PR

So far most backoffice support has already been migrated to Umbraco 17, but rich text editing isn't working. Umbraco 13 hard-coded TinyMCE as the included rich text editor. Umbraco 17 has TipTap as the default rich text editor, but structured so that it's replaceable. They switched away from TinyMCE because of a licence change that means they can't distribute anything above TinyMCE v6 as part of their package, not because of any tech or usability reason. 

I did a lot of experimenting to migrate our support to TipTap, but in the end chose to re-adopt TinyMCE using [the TinyMCE.Umbraco package which is maintained by a third-party](https://github.com/ProWorksCorporation/TinyMCE-Umbraco) but recommended by Umbraco for anyone who wants to stick with TinyMCE. The key reasons for this were:

- [TipTap can't handle legal numbering that we need for text on our corporate website](https://github.com/umbraco/Umbraco-CMS/issues/21146) (this was the key reason).
- With TinyMCE and Umbraco we can add styles to `site.css` in a consuming site and have them appear as choices in the rich text editor dropdown. With TipTap there is an equivalent way of adding custom styles to a _new_ data type, but not of adding them to an existing data type installed by one of our packages.
- Our corporate website editorial team has only just migrated to TinyMCE and it's not a good time to ask them to move again to TipTap, especially if it has workarounds where TinyMCE had more mature features.

This does mean that we're stuck on TinyMCE v6 for now but it's a capable editor that meets our needs. Updating to TipTap may be an option in future once the issues above are ironed out.

## Changes in this PR

- Auto-migrated TipTap settings on data types reverted to TinyMCE
- We have custom `PropertyValueFormatter` classes that update the HTML of rich text properties at render time. In Umbraco 13 these relied on custom property editor types, but that no longer works so now they rely on the aliases of well-known properties that should have the `PropertyValueFormatter` applied. This also meant `IRichTextPropertyEditorAliasProvider` was obsolete.
- Migrates rich text settings to a slightly amended JSON format compatible with the package we're now using for TinyMCE, including updating the documentation.
- Includes the work I did on converting the HTML produced by TipTap into the HTML we want, for future compatibility
- Updates how `govuk-frontend.css` and `tpr.css` are referenced because I needed to see the correct styles, but this is still a workaround to be properly fixed in a later PR
- Updates the unit test mocking method `CreatePropertyType()` to create a better mock, to support testing of modified `PropertyValueFormatter` classes
- Resolves an issue specific to the example app where the text in the context bar was the wrong colour because it was surrounded by `<p class="govuk-body">`. This is resolved by adding a PropertyValueFormatter specifically for the header.
- Adds some missing element type aliases to `ElementTypeAliases.cs` and missing property aliases in `PropertyAliases.cs` and `TprPropertyAliases.cs`

## How to review this PR

To review this PR, follow the [updated instructions for running the Umbraco example app](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/docs/umbraco/run-example-application.md). When following these steps, immediately after cloning the repo switch to the branch for this PR.

Having followed the instructions you should now be able to go to any block with a rich text editor, open the settings pane, and see similar configuration to that which was available in v13. The 'Text' block (used on the home page for example) should have a dropdown list of styles that includes GOV.UK classes.

Custom classes dropdown

Verify that pasting richly-formatted content from Word still sees that content tidied up, without extraneous HTML and styles that don't display properly (ie the 'paste from Word' plugin still works).

## Out-of-scope for this PR

Everything else. 

You can expect the migration to be at the point where:
- .NET code builds
- .NET tests pass

Otherwise you can expect:
- build warnings
- outdated documentation

Also coming later: 


ZebSadiq (approved):
All works well. One question / something to think about: If setting or not setting a Paragraph style results to the same output, can we hide the Paragraph option from the styles?

sussexrick:
> All works well. One question / something to think about: If setting or not setting a Paragraph style results to the same output, can we hide the Paragraph option from the styles?

Good question. I think there was a time when the only way to remove a heading style was to set it to Paragraph, but it appears that's no longer the case. The only difference I could produce now is that if you select "Paragraph" the styles are applied correctly in the backoffice - this is two paragraphs with and without explicit selection:

Different paragraph styles

...but we could fix that with an update to the backoffice stylesheet.

I think it may be a question for content designers as to what seems most intuitive or whether it's confusing.

Issue #535, opened 2025-10-24
Upgrade to Umbraco 17 / .NET 10


sussexrick:
Umbraco 17 and .NET 10 are the next LTS versions and we need to update this repo to continue to be on supported versions. .NET 10 should be an easy backwards-compatible upgrade but Umbraco will need some work, due to the backoffice being rebuilt in new technology and the TinyMCE rich text editor being replaced with TipTap.

- [x] Upgrade to Umbraco 17 / uSync 17 / .NET 10
- [x] Update unit testing to pass on v17
- [x] Review all uses of IPublishedSnapshotAccessor and PropertyCacheLevel.Snapshot (including updating XML comments), which appears to be obsolete
- [x] Upgrade ModelProperty property editor, which is based on Angular JS tech
- [x] Upgrade all backoffice views, which are based on Angular JS tech
- [x] Custom rich text editor data types need a [property editor UI](https://docs.umbraco.com/umbraco-cms/customizing/property-editors/composition/property-editor-ui) registered in the new backoffice tech (https://docs.umbraco.com/umbraco-cms/reference/property-editor-uis looks useful)
- [x] All rich text editor data types need a review of which buttons are available, as TipTap enables more by default
- [x] Upgrade smart paste of Word HTML into rich text editor
- [x] Review all recommended rich text editor settings in appsettings.json, including allowed elements and support for empty table cells
- [x] Review support for custom formats in the rich text editor style selector, and whether blockquote and code block can be removed
- [x] Review approach to client-side asset inclusion following removal of Smidge (close #425 if Smidge is not reintroduced)
- [x] Review SetupContentType() which should now support different types of content type (media types, element types etc)
- [x] Review ServiceContext.CreatePartial() for any new services we should mock for testing
- [x] Ensure all appropriate resources are packaged successfully for downstream consumption
- [x] Ensure all uSync content up-to-date and included in packages (bear in mind move to v17 folder)
- [x] TPRDocuments.cshtml - `documentLink.Content` no longer exists, so will need to use the media cache to look up properties that used to be there
- [x] Side navigation component is not loading JS in Umbraco example app (most likely related to Smidge) and also has missing images
- [x] Remove any Angular JS block labels or replace with Umbraco-flavoured markdown
- [x] [Language variants are now block-level](https://umbraco.com/blog/umbraco-15-release/) - review whether settings are needed on block properties to support this
[...]

PR #605, opened 2026-01-26
Fix regression in grid example page #535

sussexrick:
## Context for this PR

## Scope of this PR

This PR just fixes an error in the Umbraco backoffice on the 'Grid' page of the example app. The block grid property is using the 'GOV.UK Block grid' data type and needs to change to 'TPR Block grid' to support a TPR block that's in the page. 

This happened because this TPR block was added to v13 shortly after the v17 branch split off, and it wasn't fully merged into the v17 branch.

To review this PR, as well as running the Umbraco example app in the normal way you'll also need to compile the backoffice views with the following command:

```cmd
npm run backoffice
```

In the backoffice run a uSync Import to make sure you're up-to-date, then go to the Grid page and notice that the 'Back to menu' is present with a preview.

### Before and after

Before: block is not allowed

After: block is allowed

## Out-of-scope for this PR

Everything else. 

- .NET code builds
- .NET tests pass
- all blocks have a backoffice preview

Otherwise you can expect:
- build warnings
- outdated documentation

Also coming later: 


AlexBeesley (approved):
Image

PR #610, opened 2026-02-02
Add a cache-busting parameter to our stylesheet loaded in the backoffice

sussexrick:
## Context for this PR

## Scope of this PR

In Umbraco 13 [Smidge](https://github.com/shazwazza/smidge) was included by default for bundling and minifying client-side assets. Between v13 and v17 it was removed, and so this PR is part of a complete review of how we include client-side assets. 

This PR adds a cache-busting parameter to our stylesheet when loaded by property editors and block previews in the backoffice. The value for the cache-busting parameter is based on the value we set for the NuGet package version. This ensures that consuming projects will get the latest CSS based on the package version they have imported, without us having to maintain the version in an extra place.

We specify the NuGet version in an MSBuild property, so this PR uses MSBuild to generate a TypeScript file at build time with that same version. We already use MSBuild to generate files in this repo, so this is a consistent pattern. 

It doesn't end the TypeScript line with a semi-colon because that got swallowed by MSBuild every way I tried, but since it's generating a single-line file that doesn't cause a problem with parsing.

## Out-of-scope for this PR

Everything else. 

- .NET code builds
- .NET tests pass

Otherwise you can expect:
- build warnings
- TODO comments in code
- outdated documentation

Also coming later: 

- Further improved support for client-side assets following removal of Smidge between Umbraco 13 and 17

#535

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -435,3 +435,6 @@ GovUk.Frontend.Umbraco/uSync/v17/ContentTypes/
 GovUk.Frontend.Umbraco/uSync/v17/DataTypes/
 ThePensionsRegulator.Frontend.Umbraco/uSync/v17/ContentTypes/
 ThePensionsRegulator.Frontend.Umbraco/uSync/v17/DataTypes/
+
+# Generated files with package version info
+*.generated.*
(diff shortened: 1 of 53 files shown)


ZebSadiq (approved):
Had to perform npm install & npm run build to get this working.

PR #613, opened 2026-02-03
Convert side nav to a partial view for consistency and add caching

sussexrick:
## Context for this PR

## Scope of this PR

The 'TPR side navigation' component was implemented as a view component which, while technically fine, is inconsistent with the rest of the components in our packages. Since the upgrade to v17 is a breaking change, this is the opportunity to change that.

This PR makes the minimal changes to convert it from a view component to a partial view. 

It ensures the arrow image used by the component has a cache-busting parameter, and thereby ensures it also matches the pattern that gets long-term caching headers. It also uses the minified version of the script.

## Out-of-scope for this PR

Everything else. 

- .NET code builds
- .NET tests pass

Otherwise you can expect:
- build warnings
- TODO comments in code
- outdated documentation

Also coming later: 

#535


HarryAmmon (commented):
Pressing the down arrow doesn't not expand to show the child items. Pressing the up arrow does close the list.

I have checked an [instance of this component on the TPR webside](https://www.thepensionsregulator.gov.uk/en/document-library/code-of-practice) which works as expected, so I am raising this here as it appears to be a regression.

Image

An error is thrown in the console 
image

sussexrick:
@HarryAmmon Thank you for spotting that. It was another error created by minification. I've made a slight change to the code which fixes it.

HarryAmmon (approved):

Issue #499, opened 2025-08-29
Review and document breadcrumbs implementation


sussexrick:
Breadcrumbs exists as a tag helper from the `govuk-frontend-aspnetcore` library.

We've extended this with a view component that gets its navigation items from Umbraco, but although this is shown in the example app it's completely undocumented in the README.

We've also got what appears to be a duplicate view `GovUk.Frontend.Umbraco/Views/Shared/GOVUK/Breadcrumbs.cshtml` that is also completely undocumented. 

Our corporate website is our only known consumer of breadcrumbs at the moment and it uses the view component, suggesting that the duplicate is unused.

Review the implementation of this component and determine:
- whether a view component was the right choice given it's the only example in this library that otherwise uses tag helpers (we also use partial views, it may be fine and completely consistent with that approach)
- whether the apparent duplicate view is actually a duplicate, when it was introduced and what its role is or was.

sussexrick:
Comment from @villarin "This was a left over from the breadcrumbs rework - so let's remove"

sussexrick:
There's no reason to use a view component here so moving this to a partial view in v10.0.0 for consistency.

PR #616, opened 2026-02-03
Update search results component for Umbraco v17 

sussexrick:
## Context for this PR

## Scope of this PR

In Umbraco 13 [Smidge](https://github.com/shazwazza/smidge) was included by default for bundling and minifying client-side assets. Between v13 and v17 it was removed, and that removal and the subsequent review of client-side assets had broken the search results component. This PR fixes that with the following changes:

- minor refactoring of the script to eliminate errors introduced by minification (and auto-formatting of whitespace by VS Code)
- include the script in the partial view, removing the need to reference it separately (available for Umbraco only at this stage)
- moved the lib folder to a location consistent with other projects in this repo
- updated documentation 

## Out-of-scope for this PR

Everything else. 

- .NET code builds
- .NET tests pass

Otherwise you can expect:
- build warnings
- TODO comments in code
- outdated documentation

Also coming later: 

#499


ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRSearchResults.cshtml
```diff
@@ -40,4 +41,5 @@
             <a href="@link.Url" target="@link.Target">@link.Name</a>
         }
     </tpr-search-results-footer-links>
-</tpr-search-results>
\ No newline at end of file
+</tpr-search-results>
+<script src="/_Content/ThePensionsRegulator.Frontend/tpr/tpr-search-results.min.js" type="module" asp-append-version="true"></script>
```
HarryAmmon, line 45: Whilst unlikely for this component, if we have 2 instances of the component there will be 2 scripts tags in the HTML. The browser doesn't show any errors with this but it doesn't feel great.

There will be other components, like the upcoming address lookup, that will be on a page more than once so will have a script tag for each instance.

Whilst more work now, would it be better to have a service that determins what script tags to have, based on the HTML that is about to be rendered?  Or a simpler approach, could be minify all of the `tpr-*.js` files into one `.js` file that is loaded on every page?

sussexrick: > better to have a service that determins what script tags to have

Ironically, that was Smidge.

We could add it back, but I think this method is fine. The reason is that when you use `type=module` even if the script tag appears twice the browser only loads it once. That's why you're not seeing any errors.

We don't need to bundle our JS. It's now heavily cached, it's all on the same domain so won't incur multiple DNS lookups, and we have HTTP/2 multiplexing meaning multiple small files are not a problem.

HarryAmmon (approved):

PR #618, opened 2026-02-03
Always treat a rich text property as potentially null in a backoffice view

sussexrick:
## Context for this PR

## Scope of this PR

Backoffice views have already been upgraded to TypeScript for v17. This PR fixes a bug by treating all rich text properties as potentially `null`, which is an error I saw reported in the browser console for one of the views.

Testing this PR will require building the backoffice views, which is done from the command line at the root of the repo. 

```cmd
npm run backoffice
```

If you are working from a fresh clone this has to happen after the .NET build, because the .NET build generates a file the TypeScript build depends on.

## Out-of-scope for this PR

Everything else. 

- .NET code builds
- .NET tests pass

Otherwise you can expect:
- build warnings
- TODO comments in code
- outdated documentation

Also coming later: 

- Further improved support for client-side assets following removal of Smidge between Umbraco 13 and 17

#535


GovUk.Frontend.Umbraco/Backoffice/src/blocks/views/govuk-notification-banner.ts
```diff
@@ -56,8 +56,8 @@ export class GovUkNotificationBannerView extends UmbElementMixin(LitElement) imp
                     ${ this.settings?.type === 'Success' && !(this.settings?.title) ? html`<h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">Success</h2>` : null }
                 </div>
                 <div class="govuk-notification-banner__content" aria-hidden="true">
-                    <h3 class="govuk-notification-banner__heading">${ html`${unsafeHTML(disableLinks(this.content?.heading.markup)) }` }</h3>
-                    ${ this.content?.text?.markup ? html`<p class="govuk-body">${unsafeHTML(disableLinks(this.content.text.markup)) }</p>` : null }
+                    <h3 class="govuk-notification-banner__heading">${ html`${unsafeHTML(disableLinks(this.content?.heading?.markup)) }` }</h3>
+                    ${ this.content?.text?.markup ? html`<p class="govuk-body">${unsafeHTML(disableLinks(this.content.text?.markup)) }</p>` : null }
```
villarin: missing null check on content here? - should be 
```
{unsafeHTML(disableLinks(this.content?.text?.markup)) }

villarin (changes requested):
Null checks look good! Just one minor amend requested

GovUk.Frontend.Umbraco/Backoffice/src/blocks/views/govuk-notification-banner.ts
sussexrick: It's not required because `this.content?.text?.markup` has already evaluated to truthy. But that also means I didn't need to add the null check on `text?` so I'm happy to add it to be consistent if you'd prefer it...

sussexrick: @villarin I've made the suggested amendment

villarin (approved):

PR #619, opened 2026-02-03
Tidy up the notification banner example page

sussexrick:
The notification banner example page in the Umbraco app works before and after this PR. In this PR:

- A legacy setting (Reversed) is swapped for a current one (Secondary).
- The JavaScript is updated to target a specific banner by its class as a better example, rather than the first banner on the page which just happens to be the right one.

Legacy setting removed

#535


HarryAmmon (changes requested):
The picker for the 'model property' property shows an error. This was after a fresh install of Umbraco and running a full uSync import.

Image

villarin (dismissed):

HarryAmmon:
The models builder is reporting that the models are out of date.

image

HarryAmmon:
> The picker for the 'model property' property shows an error. This was after a fresh install of Umbraco and running a full uSync import.
> 
> Image

This was resolved after I ran `npm run backoffice`, which is mentioned in the setup docs

sussexrick:
> > The picker for the 'model property' property shows an error. This was after a fresh install of Umbraco and running a full uSync import.
> 
> This was resolved after I ran `npm run backoffice`, which is mentioned in the setup docs

Yes, this will be a standard part of running the app now. I tried automating it with MSBuild the way we do with other setup steps, but it was blocked by policy on our TPR machines.

sussexrick:
@HarryAmmon I've updated the models (just one change)

HarryAmmon (approved):

PR #623, opened 2026-02-05
#252204 - Video iframe lacks descriptive labelling

AlexBeesley:
This pull request adds support for customizable headings and descriptions to the YouTube video component, improving accessibility and flexibility for content editors. It introduces new properties for video description, heading class, and heading level, updates the relevant models, configuration, and templates, and ensures these new fields are rendered both in the backoffice preview and on the frontend. Additionally, it adds a dictionary entry for a translatable video title.

**Enhancements to YouTube Video Content Model and Data Types:**

* Added a new rich text `description` property to the `TprYouTubeVideo` content type, including its definition as a TinyMCE editor data type, and updated the generated model to expose this property. [[1]](diffhunk://#diff-db673b4203f2d8afdace1f3ff72b7f8ff372e4ed5e8fc3ccea4c0a93859d746dR24-R39) [[2]](diffhunk://#diff-e59a1fac8a044abb2041eeba584261a6989193d1927f2c9c017f3916b90df735R52-R59) [[3]](diffhunk://#diff-19e48628f275399d33b0cd6ddb6bd63313a14eeac11568f091c6194daeb4678eR1-R36)
* Updated the `youtube-video.config` content data to include the new `description` field for relevant video entries. [[1]](diffhunk://#diff-5de837488f50d7ff8fe50d6ed9a83ab5d05a3f04f52aaf713184dc1d39da5519R40) [[2]](diffhunk://#diff-5de837488f50d7ff8fe50d6ed9a83ab5d05a3f04f52aaf713184dc1d39da5519R354-R360) [[3]](diffhunk://#diff-5de837488f50d7ff8fe50d6ed9a83ab5d05a3f04f52aaf713184dc1d39da5519R372) [[4]](diffhunk://#diff-5de837488f50d7ff8fe50d6ed9a83ab5d05a3f04f52aaf713184dc1d39da5519R401)

**Enhancements to YouTube Video Settings:**

* Added new properties `headingClass` and `headingLevel` to the `TprYouTubeVideoSettings` content type and its generated model, allowing editors to set the heading size and level for video titles. [[1]](diffhunk://#diff-b5edf33a2f4a8956042a59d53c660fb842bd3cc66c29efc106df858e70b7c54aR35-R66) [[2]](diffhunk://#diff-1c1f42611167d0138ccc847d6c642bd8d65006211f7de105cddb1b936243699aR59-R74)
* Updated `youtube-video.config` to include `headingClass` and `headingLevel` for video settings. [[1]](diffhunk://#diff-5de837488f50d7ff8fe50d6ed9a83ab5d05a3f04f52aaf713184dc1d39da5519R162-R163) [[2]](diffhunk://#diff-5de837488f50d7ff8fe50d6ed9a83ab5d05a3f04f52aaf713184dc1d39da5519R438-R441) [[3]](diffhunk://#diff-5de837488f50d7ff8fe50d6ed9a83ab5d05a3f04f52aaf713184dc1d39da5519R455-R458) [[4]](diffhunk://#diff-5de837488f50d7ff8fe50d6ed9a83ab5d05a3f04f52aaf713184dc1d39da5519R486-R487)

**Frontend and Backoffice Rendering:**

* Updated the frontend Razor view to render the new description, heading level, and heading class, with sensible defaults and dictionary-based title support for accessibility. [[1]](diffhunk://#diff-552f0d99b6776d3592ca3665472adf8444c5b3d1f2fdb9396d39999db354b318R5-R9) [[2]](diffhunk://#diff-552f0d99b6776d3592ca3665472adf8444c5b3d1f2fdb9396d39999db354b318L36-R38) [[3]](diffhunk://#diff-552f0d99b6776d3592ca3665472adf8444c5b3d1f2fdb9396d39999db354b318L53-R97)
* Modified the HTML generator to render headings and descriptions in the no-cookies embedded player, applying the appropriate classes and ensuring accessibility. [[1]](diffhunk://#diff-974e897f90a3899bf7e976817841ce3f4ce1222d778b22698c7dde65bcd643c5R36-R59) [[2]](diffhunk://#diff-974e897f90a3899bf7e976817841ce3f4ce1222d778b22698c7dde65bcd643c5L51-R71)
* Updated the backoffice block preview to display the video title and description.

**Supporting Infrastructure:**

* Added new property aliases for `description`, `headingLevel`, and `headingClass` in `TprPropertyAliases`.
* Added a dictionary entry for a translatable video title.

**Configuration and Sort Order Adjustments:**

* Adjusted sort orders for properties in the content type and settings configuration to accommodate the new fields and maintain logical ordering. [[1]](diffhunk://#diff-db673b4203f2d8afdace1f3ff72b7f8ff372e4ed5e8fc3ccea4c0a93859d746dL49-R65) [[2]](diffhunk://#diff-db673b4203f2d8afdace1f3ff72b7f8ff372e4ed5e8fc3ccea4c0a93859d746dL65-R81) [[3]](diffhunk://#diff-b5edf33a2f4a8956042a59d53c660fb842bd3cc66c29efc106df858e70b7c54aL51-R83)


ThePensionsRegulator.Frontend.Umbraco/App_Plugins/ThePensionsRegulator.Frontend.Umbraco/blocks/views/TPRYouTubeVideo.html
```diff
@@ -1,7 +1,11 @@
 ﻿<div class="backoffice-block-view" ng-click="block.edit()" ng-focus="block.focus" aria-label="Edit YouTube video">
+    <div class="tpr-youtube-video-preview-content">
+        <h3 ng-if="block.data.title">{{ block.data.title }}</h3>
+        <p ng-if="block.data.description">{{ block.data.description }}</p>
```
sussexrick: This incorrectly represents multiple paragraphs:

Multiple paragraphs in edit view

Single paragraph in preview

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprYouTubeVideo.cs
```diff
@@ -37,6 +37,22 @@ public virtual TagBuilder GenerateTprYouTubeNoCookiesEmbeddedPlayer(TprYouTubeVi
...
+
+            var heading = new TagBuilder("h3");
+            heading.MergeCssClass(video.SmallerHeading ? "govuk-heading-s" : "govuk-heading-m");
+            heading.MergeCssClass("tpr-video-wrapper-no-cookies__heading");
+            heading.InnerHtml.AppendHtml($"{video.Title} (video)");
+            containerTag.InnerHtml.AppendHtml(heading);
+
+            if (!string.IsNullOrWhiteSpace(video.Description))
+            {
+                var description = new TagBuilder("p");
```
sussexrick: This incorrectly renders multiple paragraphs:

Multiple paragraphs in edit view

Rendered as a single paragraph

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRYouTubeVideo.BackOffice.cshtml
sussexrick, line 1: This is a redundant empty file. It can be removed from the PR.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tpryoutubevideo.config
```diff
@@ -21,6 +21,22 @@
   </Info>
   <Structure />
   <GenericProperties>
+    <GenericProperty>
+      <Key>3565e25f-26da-4cf0-bf43-fe713affe0a6</Key>
+      <Name>Description</Name>
+      <Alias>description</Alias>
+      <Definition>c6bac0dd-4ab9-45b1-8e30-e4b619ee5da3</Definition>
+      <Type>Umbraco.TextArea</Type>
```
sussexrick: I think this field should support rich text - people may want to add bold, links, special characters, subscript or superscript for example.

Create a new dedicated data type for it (see 'TPR Rich text editor (Section card)' for an example) so that you can control which editor features are available. Have a good think about which buttons are appropriate to allow and which aren't.

This should also handle the issues flagged elsewhere with multiple paragraphs. You shouldn't need to render a wrapping `p` tag as the content will include its own.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRYouTubeVideo.cshtml
```diff
@@ -52,17 +52,19 @@
 
     var autoplay = Model.Settings.Value<bool>(TprPropertyAliases.VideoAutoplay);
     bool playsInline = Model.Settings.Value<bool>(TprPropertyAliases.VideoPlaysInline)!;
+    var description = Model.Content.Value<string>(TprPropertyAliases.VideoDescription);
```
sussexrick: New properties could be null if they're on a page that hasn't been saved since you updated the data type. Cast to `<string?>` and `<bool?>` and add `?? "some default value"` at the end of the line.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprYouTubeVideo.cs
```diff
@@ -17,7 +17,7 @@ public virtual TagBuilder GenerateTprAblePlayer(TprYouTubeVideo video)
 
 
             var videoTag = new TagBuilder("video");
-            if (!video.Attributes.ContainsKey("id")) { video.Attributes.Add("id", Guid.NewGuid().ToString()); } // Able Player requires id
```
sussexrick, line 20: Any reason this comment is removed? It's a useful reminder why never to remove the default on the id setting.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprYouTubeVideo.cs
```diff
@@ -47,8 +63,9 @@ public virtual TagBuilder GenerateTprYouTubeNoCookiesEmbeddedPlayer(TprYouTubeVi
                 src += "?autoplay=1&mute=1";
             }
 
+            var iframeTitle = $"Video: {video.Title} (video)";
```
sussexrick: Including both a `Video: ` prefix and a ` (video)` suffix is unnecessary repetition. You need one or the other. Also, videos appear on Welsh pages so this needs to be localisable text (which usually means injected from outside the tag helper so you offload language management to the app).

In Umbraco it's probably best kept in the Translations/Dictionary section so it only has to be translated once.

It would be best to put the whole string in there, with a token to replace with the title. That way if a different language has different syntax (one needs a prefix, another needs a suffix, or they have different quote marks around the title) you will support it.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprYouTubeVideo.cs
```diff
@@ -37,6 +37,22 @@ public virtual TagBuilder GenerateTprYouTubeNoCookiesEmbeddedPlayer(TprYouTubeVi
             var containerTag = new TagBuilder("div");
             containerTag.MergeAttributes(video.Attributes);
             containerTag.MergeCssClass("tpr-video-wrapper-no-cookies");
+
+            var heading = new TagBuilder("h3");
```
sussexrick: Both here and in the backoffice preview you've hard-coded the heading level to h3. That will often be wrong. In fact it's wrong on your example page, where you jump from h1 to h3 without an h2.

Look at the 'GOV.UK Heading level (Accordion)' data type and how its used for an example of allowing the heading level to be chosen by the user based on the context they're in. Have a sensible default if they don't select anything (that may be h2 or h3).

The appearance of the heading is a separate concern to the heading level. "Smaller heading" is vague, and inconsistent. Give the user control by allowing them to select from a list of classes. Have a sensible default if they don't select anything (see next comment). You could re-use the 'GOV.UK Heading size' data type to offer a choice of classes, or define a new similar data type 'TPR Heading size (YouTube video)' if you only want to offer `-s` and `-m` sizes.

sussexrick: You can call `IGovUkHeadingClassProvider` and call `HeadingClasses` to get the appropriate default class if none is selected by the user. See `GovUkCaption.cshtml` for an example. Other headings on the page use this and the result can change depending on a sitewide setting, so it's important to use this rather than hard-coding to stay consistent.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/YouTubeVideoSettingsSmallerHeadingToggle.config
sussexrick, line 1: I think you can remove this based on other comments, but in any case you wouldn't need a dedicated data type for a boolean property. There are only two possible settings and we have existing data types for both.

sussexrick (changes requested):

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprYouTubeVideo.cs
AlexBeesley: image

Fixed by an earlier commit switch this to an RTE.

sussexrick: It's better, but you're still surrounding the RTE output with a `p` tag, but it now brings its own paragraphs and `p` within `p` is invalid, so you need to remove the outer one.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprYouTubeVideo.cs
```diff
@@ -33,10 +33,30 @@ public virtual TagBuilder GenerateTprAblePlayer(TprYouTubeVideo video)
...
             containerTag.MergeAttributes(video.Attributes);
             containerTag.MergeCssClass("tpr-video-wrapper-no-cookies");
+
+            var heading = new TagBuilder(video.HeadingLevel);
+            if (!string.IsNullOrWhiteSpace(video.HeadingSize))
+            {
+                heading.MergeCssClass(video.HeadingSize);
+            }
+            heading.MergeCssClass("tpr-video-wrapper-no-cookies__heading");
+            heading.InnerHtml.AppendHtml(video.IframeTitle);
```
sussexrick, line 48: I don't think " (video)" is meant to be visible text on the end of the heading?

Image

It makes sense for the iframe title, which is hidden text that will be exposed to a screen-reader, and it tells those users what they're dealing with. For a visual user, they can already see it's a video. The problem is you're re-using the iframe title property for a different context, which is not an iframe title.

It's also inconsistent with the backoffice preview, though I think it's probably the preview that's correct.

ThePensionsRegulator.Frontend/TagHelpers/TprYouTubeVideoTagHelper.cs
```diff
@@ -33,9 +37,13 @@ public class TprYouTubeVideoTagHelper : TagHelper
         private bool _playsInline = true;
         private string _preload = ComponentGenerator.YouTubeVideoDefaultPreload;
         private bool _useAblePlayer = false;
+        private string? _description = null;
+        private string _headingLevel = "h3";
```
sussexrick: Why does heading level default to h3 here but h2 in your switch expression?

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/video-title.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Dictionary Key="58868176-7af3-459a-bb77-a1b16a67edb8" Alias="Video title" Level="0">
+  <Info />
+  <Translations>
+    <Translation Language="cy">{{VIDEO_TITLE}} (fideo)</Translation>
```
sussexrick, line 5: Nice work including Welsh support 👍

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tpryoutubevideosettings.config
```diff
@@ -32,13 +32,45 @@
...
+      <Validation></Validation>
+      <Description><![CDATA[Select the heading level for your text.]]></Description>
       <SortOrder>0</SortOrder>
       <Tab Alias="settings">Settings</Tab>
       <Variations>Nothing</Variations>
       <MandatoryMessage></MandatoryMessage>
       <ValidationRegExpMessage></ValidationRegExpMessage>
       <LabelOnTop>false</LabelOnTop>
     </GenericProperty>
+    <GenericProperty>
```
sussexrick, line 58: You should be able to configure heading level and heading size to appear next to each other:

Image

Consistency with section cards would be good, so "heading size" rather than "heading class" for example. Notice how the help text here tells you important information, whereas yours repeats the information in the property name using more words.

Image

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRYouTubeVideo.cshtml
```diff
@@ -50,19 +52,47 @@
...
+
+    var headingLevelProperty = Model.Settings.Value<string>(TprPropertyAliases.VideoHeadingLevel) ?? string.Empty;
+    var headingLevelNumber = headingLevelProperty.Replace("heading ", string.Empty, StringComparison.OrdinalIgnoreCase).Trim();
+    if (!int.TryParse(headingLevelNumber, out int numericHeadingLevel) || numericHeadingLevel < 1 || numericHeadingLevel > 6)
+    {
+        numericHeadingLevel = 2;
+    }
+    var headingLevel = $"h{numericHeadingLevel}";
+
+    var headingClasses = headingClassProvider.HeadingClasses(Umbraco.AssignedContentItem);
```
sussexrick, line 67: Nice work supporting variable heading sizes 👍

sussexrick (dismissed):

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprYouTubeVideo.cs
AlexBeesley: image
This was John Haynes design for this work.

villarin (commented):
HI @AlexBeesley please raise the build package version for this one.
I can see all comments has been dealt with now - so I have resolve those and happy to approve

villarin (dismissed):

villarin (approved):
updated vulnerable dep

PR #633, opened 2026-02-12
256262 FAQs with multiple CTA don't go to widget

AlexBeesley:
This pull request includes a version bump and a JavaScript enhancement to improve accessibility handling for navigation buttons on the search results page.

Versioning:

* Updated the project version in `Directory.Build.props` from `9.2.0` to `9.2.1`.

Frontend accessibility improvement:

* Modified the logic in `tpr-search-results.js` so that accessibility-related changes (removal of visually hidden spans) are applied to all elements with the `tpr-search-results__nav-button` class, instead of just the first one. This ensures consistent behavior across multiple navigation buttons.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,5 +1,5 @@
 <Project>
   <PropertyGroup>
-    <Version>9.2.0</Version>
+    <Version>9.2.1</Version>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


villarin (approved):

HarryAmmon (approved):
Tested by adding a second `tpr-search-results__nav-button` to the search results example page and verified both linked to the search results component.

Image

PR #636, opened 2026-02-13
Refactor packaging of govuk-umbraco-backoffice.css

sussexrick:
## Context for this PR

## Scope of this PR

This PR continues a review of the MSBuild that runs when our packages are installed.

Packages designed to be referenced by web projects have sometimes been referenced by class libraries, and the class library projects get the files even though they don't want them. One example of this is the `govuk-umbraco-backoffice.css` file, which appears in a `Content` folder (see screenshot in #343).

This happens because the project file has configuration to pack it inside the `contentFiles` folder in the package, and this is how `contentFiles` behave.

When our packages are installed, the `<project-name>.targets` file for that project is run when the consuming project is built. This PR uses that mechanism to avoid having the folder appear:

1. instead of building the SASS into CSS in the `Styles` folder, it's built under `wwwroot`
2. this causes it to be included in the NuGet package as a static asset (served from the DLL) rather than a content file (served from disk)
3. Umbraco needs it to be a file on disk, so our `.targets` file finds the package folder, and copies the file into the project at the location expected by Umbraco
4. gitignore already ignored it there

This PR also removes the minification step for this file. It was added only in this series of PRs, but testing showed this broke the mechanism for adding custom styles in the rich text editor, which depends on comments being preserved in the CSS output.

## How to test this PR

Since we are now looking at the behaviour of the script that runs when a package is installed, testing this PR should involve:

1. Change the version number to something like `10.0.0-alpha001` in `Directory.Build.props` at the top of the repo and generate local packages.
6.  Set up one or more local NuGet package sources to point to the alpha packages. 
3 .Create a consuming project and consume the packages. 
7. Verify that the Content folder does not appear, but `govuk-umbraco-backoffice.css` does appear under `wwwroot/css` and styles are visible in the rich text editor.

## Out-of-scope for this PR

Everything else. 

- .NET code builds
- .NET tests pass

Otherwise you can expect:
- build warnings
- TODO comments in code
- outdated documentation

Also coming later: 

- Fix the _other_ unwanted extra files included when referencing packages outside of the main web project


villarin (approved):
Tested from vanilla solution. Build working as expected and application running locally after content and scheme update via usync

PR #637, opened 2026-02-16
Review docs for new app setup, and modernise example apps using top-level statements

sussexrick:
## Context for this PR

## Scope of this PR

This PR modernises example apps by moving to `Program.cs` with top-level statements instead of the older `Startup.cs`, so that they can serve as true examples to copy for a new app. It also reviews documentation and application code for new projects so that they match.

It also includes a small JS fix for pages which have no `<title>` element (there should always be one, but I noticed it wasn't handled while working on this PR).

## Out-of-scope for this PR

Everything else. 

- .NET code builds
- .NET tests pass

Otherwise you can expect:
- build warnings
- TODO comments in code
- outdated documentation

Also coming later: 

- Fix unwanted extra files included when referencing packages outside of the main web project

#535


GovUk.Frontend.Umbraco.ExampleApp/appsettings.Development.json
```diff
@@ -0,0 +1,45 @@
+{
```
pjoblingtpr, line 1: Just checking you want to include this file? Was previously excluded.

sussexrick: Yes, it's deliberate. It's because the default Visual Studio template for an Umbraco app now includes an appsettings.Development.json whereas it never used to, so I'm updating the example app to match the recommendations. As an example app there's nothing sensitive, nor ever likely to be.

pjoblingtpr (approved):

PR #641, opened 2026-02-19
Fix example app views that are broken in GOV.UK mode 

sussexrick:
## Context for this PR

## Scope of this PR

This PR addresses some tech debt by fixing the GOV.UK-only view of example app pages. 

When you set the `UseTPRStyles` setting to `false` in either example app, you get the pure GOV.UK version of the components. We typically review just the TPR version, hence some pages were broken, and others were just blank.

For TPR-only component pages where there is nothing useful to see without TPR styles, the existing pattern of displaying text explaining that situation is implemented for pages that didn't have it.

You can test this PR by setting  `UseTPRStyles` to `false` and verifying that each page in each example app loads and has some reasonable content.

PJ: Additionally, I've added a strongly-typed, extendable solution for application config - see AppConfig class and Program.cs for details. You can now inject app config in anywhere as a class. 

## Out-of-scope for this PR

Everything else. 

- .NET code builds
- .NET tests pass
- consuming the package works but still creates unwanted folders in some cases

Otherwise you can expect:
- C# build errors and TODO comments
- outdated documentation

Also coming later: 

- Fix unwanted extra files included when referencing packages outside of the main web project

#535


GovUk.Frontend.Umbraco.ExampleApp/Views/SideNavigation.cshtml
```diff
@@ -1,9 +1,21 @@
-﻿@using ThePensionsRegulator.Umbraco
+﻿@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<ContentModels.SideNavigation>
+@using ThePensionsRegulator.Umbraco
 @using Umbraco.Cms.Web.Common.PublishedModels;
-@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<ContentModels.SideNavigation>
 @using ContentModels = Umbraco.Cms.Web.Common.PublishedModels;
+@using Microsoft.Extensions.Configuration;
+@inject IConfiguration Configuration
+@{
+    var useTPRstyles = Configuration.GetValue<bool>("TPRStyles");
```
pjoblingtpr: Should we consider a strongly-typed solution to Configuration.GetValue<bool>("TPRStyles"), to avoid the magic string everywhere. Like loading it into a config class in program.cs and injecting that where needed?

sussexrick: That would be an improvement. Given your current access level would you be able to implement it please?

pjoblingtpr: sure

pjoblingtpr: Implemented and tested this. Please have a look and let me know what you think.

SaiAndol5 (dismissed):

GovUk.Frontend.Umbraco.ExampleApp/Views/SideNavigation.cshtml
pjoblingtpr: I got Sai to review -- so we are all good.

pjoblingtpr (dismissed):
Tested and all working OK so approved.

SharedResourceLibrary/AppConfig.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace GovUk.Frontend.ExampleSharedResource
```
sussexrick: This is the wrong place for this class. Just have a version of it in each example app. They're meant to be independent of each other.

This library belongs to `GovUk.Frontend.ExampleApp` and exists to demonstrate the ability to read localised strings from a shared library. We don't need to do that in the Umbraco app because Umbraco string localisation is handled by Umbraco instead of resource files.

GovUk.Frontend.Umbraco.ExampleApp/GovUk.Frontend.Umbraco.ExampleApp.csproj
```diff
@@ -22,6 +22,7 @@
       <PackageReference Include="uSync" Version="17.0.1" />
       <ProjectReference Include="..\GovUk.Frontend.AspNetCore.Extensions\GovUk.Frontend.AspNetCore.Extensions.csproj" />
       <ProjectReference Include="..\GovUk.Frontend.Umbraco\GovUk.Frontend.Umbraco.csproj" />
+      <ProjectReference Include="..\SharedResourceLibrary\GovUk.Frontend.ExampleSharedResource.csproj" />
```
sussexrick: See comment elsewhere for details, but this reference shouldn't be here.

SharedResourceLibrary/AppConfig.cs
pjoblingtpr: Ok - understood.

pjoblingtpr: fixed this and done as you requested.

GovUk.Frontend.Umbraco.ExampleApp/GovUk.Frontend.Umbraco.ExampleApp.csproj
pjoblingtpr: Removed.

pjoblingtpr (approved):

PR #642, opened 2026-02-19
#256775 Updating heading level in accordion doesn't reduce visible font size

Ethan-Lambourne:
In this PR:
- Created a `_tpr-accordion.scss` file;
- Populated the `_tpr-accordion.scss` file with styles that change the font size of the heading within the accordion depending on which `<h2-4>` element is generated by the gov-uk accordion tag helper.

Before these changes were applied, the font size attribute was being overridden by the default `.govuk-accordion__section-button` class, so no matter which heading level was inputted into the tag helper, the actual font size for the heading would never change.

This is related to the work item: https://dev.azure.com/thepensionsregulator/TPR/_workitems/edit/256775

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,5 +1,5 @@
 <Project>
   <PropertyGroup>
-    <Version>9.2.2</Version>
+    <Version>9.2.3</Version>
   </PropertyGroup>
 </Project>
--- a/ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAccordion.cshtml
+++ b/ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAccordion.cshtml
@@ -1,2 +1,45 @@
-﻿@addTagHelper *, Microsoft.AspNetCore.Mvc.TagHelpers
-<partial name="GOVUK/GovUkAccordion" />
+﻿@using GovUk.Frontend.Umbraco
+@using GovUk.Frontend.Umbraco.Blocks
+@using ThePensionsRegulator.Umbraco
+@using ThePensionsRegulator.Umbraco.Blocks
+@inherits Umbraco.Cms.Web.Common.Views.UmbracoViewPage<IOverridableBlockReference<IOverridablePublishedElement,IOverridablePublishedElement>>;
+@addTagHelper *, GovUk.Frontend.AspNetCore
+
+@{
+    var cssClasses = Model.Settings?.Value<string>(PropertyAliases.CssClasses);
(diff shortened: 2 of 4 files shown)


sussexrick:
I don't think this is the right way to tackle the problem, for two reasons:

- the default size of headings can change, depending on a sitewide setting about which size to start with, and your CSS doesn't reflect that
- other components with a similar issue have a dropdown of size classes, so this should work in a consistent way

Take a look at the discussion on #623 for more info.

Ethan-Lambourne:
@sussexrick I think I've actioned your comment on this PR in the latest commits now, so if you could re-review this whenever you get a moment that would be great - as I'd like these changes to be included in release 21 of the corp site if possible.

villarin (approved):
Seems to be working as expected locally.
I can also see that the implementation is no longer reliant on custom css classes and that we are using the standard govu-heading sizing classes instead.

sussexrick:
@villarin @Ethan-Lambourne Sorry I didn't get to review in time, but this PR was not ready. Please can you prepare new ones for both v13 and v17 that addresses the following issues:

### Unexplained SASS change

You don't say why the `font-size: inherit` setting was added. In testing I don't see it making a difference. Without it the font-size is set by GOV.UK code, which is what we want.

### No GOV.UK support

This is not a TPR-specific behaviour, it's a GOV.UK one. As such the changes should be in the GOV.UK accordion view, and the TPR one should've remained as just a reference to the GOV.UK one as it was before. Currently if you use an accordion in a GOV.UK block list/grid it will still have the bug you were trying to solve.

### Heading sizes are fixed

You haven't addressed the fact that the default size of headings can change, depending on a sitewide setting about which size to start with. If you go to the home page of the example app in the backoffice you will find a setting where you can change the starting size for headings across the site:

Settings tab of the home page in the backoffice

This is set to `govuk-heading-l` in the example app. If you change it to `govuk-heading-xl` you will see heading sizes change across the app, including the h1 on the accordion page. Your accordion heading size does not change. This is because you need to get the class name from `IGovUkHeadingClassProvider` as shown in PR #623 that I pointed you to.

## Size customisation

This is optional but PR #623 shows how you can also add the option for users to override the default size, by adding a separate property with a dropdown list of sizes, and then only applying the class from `IGovUkHeadingClassProvider` if nothing was selected from the list:

```csharp
var headingClass = Model.Settings.Value<string>(TprPropertyAliases.VideoHeadingSize);
    if (string.IsNullOrWhiteSpace(headingClass))
    {
       /// get size from IGovUkHeadingClassProvider
    }
```

This would add greater flexibility and be consistent with other similar support for customising heading levels within components.

sussexrick:
@villarin @Ethan-Lambourne I've created issue #664 to track this

PR #653, opened 2026-02-25
#253327 ROS 1 accessibility highs civic (v17)

Max-Emery:
`Note:` These are the updates to the TPR Header Menu that are being used in v13, only minor changes were made to JS syntax to make these changes compatible with the v17 solution.

Why:

Accessibility issues flagged with Header Menu Component

High:

- Logo cannot be activated when mobile navigation is triggered
- Main navigation label is unclear for screen reader users
- Navigation links act as both dropdown triggers and links, causing confusion

Medium:

- Focus on mobile button is not colour compliant (Safari)
- Insufficient focus contrast on dropdown submenu links

In this PR:

- Header menu container and logo adjusted to ensure they usable in mobile view.
- `<i>` elements replaced with native button elements, ARIA labelling and expanded values moved onto buttons.
- Dropdown functionality separated from menu items, implemented dropdown buttons for desktop view. With supporting JS
for the following scenarios: open and close logic, user is no longer viewing or clicks outside of sub menu prompts the menu to close.
- Focus style added to mobile menu toggle.
- Updates to focus style for all anchor and button elements in header menu.


sussexrick (approved):
I'm approving because it's CIVIC's recommendation and I can't see anything wrong with your code, but there is some unusual behaviour when testing with NVDA. I'd suggest it's worth going back to them for feedback.

When tabbing to one of the down arrows NVDA announces "E button collapsed" for Employers, "T button collapsed" for Trustees and so on. To me, this seems less helpful than the combined buttons we had before...

PR #660, opened 2026-02-27
#253767 ROS-1 The open mobile menu obscures other interactive elements

Max-Emery:
Why:

Accessibility issue raised by CIVIC:

- The mobile menu remains open when it loses focus. This obscures the view of other interactive elements.

In this PR:

- Added additional functionality to support mobile menu close when user tabs out of the menu.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3776,9 +3776,9 @@
       }
     },
     "node_modules/minimatch": {
-      "version": "10.2.2",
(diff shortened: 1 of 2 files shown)


package-lock.json
villarin, line 1: Hi @Max-Emery  do we really need this without any updates to package.json?

villarin: Ignore that has been replaced with the latest npm package updates

villarin (approved):

PR #673, opened 2026-03-04
Collapse all in side navigation

georgeboydratcliff:
Update to the Side Navigation view and javascript to support a link to close all opened side navigation items


villarin (changes requested):
- Need to change the pr target to v13 
- Possibly branched out of develop. So may need to recreate the branch from v13 and re-apply your commit

ThePensionsRegulator.Frontend/Views/Components/TprSideNavigation/SideNavigation.cshtml
```diff
@@ -23,6 +23,7 @@
                 <a class="govuk-link govuk-link--no-visited-state tpr-side-nav__desktop-link" href="@Model.TitleLink.Url">@Model.TitleLink.Name</a>
             </h2>
         }
+        <a href="#" class="tpr-side-nav__collapse-all govuk-link">Collapse all</a>
```
villarin: This label needs translating so please make it part of ThePensionsRegulator.Frontend.Models.TprSideNavigationViewModel and update the navigation service to populate the text from there. This can be defined as:
- A static text for the example app service implementation 
- A dictionary entry for the umbraco example app service implementation

sussexrick:
Your link is not aligned with others:

Misaligned link

sussexrick:
You're not handling the no-JS scenario. Since this is a JS-only feature you should probably insert the link using JS. Currently it says it will collapse all but it actually reloads the page with no changes.

sussexrick:
Currently you have a link saying 'Collapse all' that does what it says. But after you've collapsed all, you still have a link that says 'Collapse all' which does nothing.

sussexrick:
Since this is a function which does a thing rather than goes to a place, it should be a button, not a link.

ThePensionsRegulator.Frontend/wwwroot/tpr/tpr-side-navigation.js
```diff
@@ -81,6 +88,40 @@ document.addEventListener("DOMContentLoaded", function () {
         button.firstElementChild.classList.remove("tpr-side-nav__list-item-arrow--up");
         button.closest("li").classList.remove("tpr-side-nav__list-item--expanded");
     }
+
+    function collapseAll() {
+        // collapse mobile first-level nav if expanded
+        const mobileToggleBtn = sideNav.querySelector(".tpr-side-nav__mobile-expand-toggle[aria-expanded='true']");
```
sussexrick, line 84: Would it be possible to select all the chevrons indicating an open menu and trigger a click() event, rather than repeating the collapse logic?

sussexrick:
Visually this is a fairly ugly/intrusive solution. It's not typical to see such a prominent link in a menu for this function. 

I can't access the original accessibility report right now but I can see they suggest supporting the Esc key as a possible solution. 

They are flagging the lack of a bypass mechanism, so a hidden skip link like the one we already have at the top of the page would be another typical solution:

skip link

Max-Emery:
`suggestion:` I noticed that when the menu closes, focus doesn’t return anywhere meaningful. If the user had focus inside the side nav, they lose context completely. Could we restore focus to the element that opened the menu?

villarin:
@georgeboydratcliff In the interest of  releasing the collapse via escape functionality as per the accessibility audit. I have created issue #730 for you to follow up.

villarin (approved):

PR #674, opened 2026-03-04
#253762 ROS-1 Logo and mobile menu button do not adapt correctly on zoom

Max-Emery:
Why: 

- Under high magnification the mobile menu toggle would overlap the logo element, this can result in a difficult experience for mobile users with low vision, as key header elements become harder to perceive or interact with.

logo overlap

In this PR: 

- Amended media query to allow mobile menu toggle to stack underneath the logo under high magnification or on very small screens.

### DIFF
--- a/ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
+++ b/ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
@@ -318,17 +318,17 @@
         outline: 3px solid transparent;
     }
 
-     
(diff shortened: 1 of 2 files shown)


sussexrick:
You're losing the left padding when it wraps. Before wrapping:

image

Wrapped:

image

Edge case - you're also losing the right padding, although this would only matter if translated to a language where the word for "Menu" is a lot longer.

sussexrick (approved):

PR #675, opened 2026-03-05
AB#257068 - Aria attribute adjustments for error placeholder on searc…

villarin:
This PR contains accessibility improvements for the search results widget as indicated by the latest audit as follows:

- Replaced pagination Show more/few link with button tag while preserving link appearance
- Enforce correct aria attributes for error placeholder 
image

### DIFF
--- a/ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchResultsFooterLinks.cs
+++ b/ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchResultsFooterLinks.cs
@@ -28,7 +28,7 @@ namespace ThePensionsRegulator.Frontend.HtmlGeneration
             var builder = new HtmlContentBuilder();
             builder.Append(DefaultShowMoreQuestionsContent);
 
-            var li = CreateLink(showMoreQuestionsAttributes, builder);
+            var li = CreateButton(showMoreQuestionsAttributes, builder);
 
             list.InnerHtml.AppendHtml(li);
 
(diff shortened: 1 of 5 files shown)


AlexBeesley (approved):

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprSearchResultsFooterLinks.cs
```diff
@@ -50,6 +50,17 @@ private TagBuilder CreateLink(AttributeDictionary dictionary, IHtmlContent conte
...
+        }
+        private TagBuilder CreateButton(AttributeDictionary dictionary, IHtmlContent content)
+        {
+            var li = new TagBuilder("li");
+            var button = new TagBuilder("button");
+            button.MergeAttributes(dictionary);
+            button.MergeCssClass("govuk-link");
+            button.InnerHtml.AppendHtml(content);
+            li.InnerHtml.AppendHtml(button);
+
```
sussexrick, line 63: @villarin @AlexBeesley When doing things like this you should check how the GOV.UK code samples do it. In this case you are missing an attribute `data-module="govuk-button"`

ThePensionsRegulator.Frontend/Styles/_tpr-search-results.scss
```diff
@@ -68,6 +68,20 @@
     #tpr-search-results-show-more-questions {
         cursor: pointer;
         color: $govuk-link-colour;
+        background: none;
+        border: none;
+        padding: 0 !important;
```
sussexrick, line 73: You don't need `!important` here as your selector already has enough weight to override existing styles (and `!important` should always be avoided if possible)

You don't need text-decoration: underline because it's part of `.govuk-link` which you've already applied.

PR #690, opened 2026-03-09
Feature/recursive model property binding

dcr670:
Change to allow recursive model binding to properties of complex types.  Mainly to support SchemeAddress.AddressLine1.


dcr670:
Created in error.

PR #691, opened 2026-03-09
Changes to allow binding to properties of complex types when Model Bi…

dcr670:
Change to allow properties of complex types to be bound.  Developed primarily for SchemeAddress.AddressLine1.


sussexrick (changes requested):
This needs tests to verify it, but also to document it. What should be supported and what shouldn't? When you make decimal a special case, should float also be a special case? What about int, and uint? When you make class a special case, what about records?

GovUk.Frontend.Umbraco.Tests/Validation/ModelPropertyControllerTests.cs
sussexrick, line 1: There's a typo in the filename

sussexrick:
@copilot Please review again, and focus on, having discovered an assembly, what is the code that scans the assembly checking for? What will it miss? What is likely to return an unexpected result?

sussexrick:
@dcr670 That wasn't the workflow I expected from Copilot but I think its feedback has some useful tweaks. However I got this feedback from Copilot in Visual Studio that is a more thorough review. I don't mean you need to address all of this, but you might want to cherry-pick some issues that you hadn't expected...

# ModelPropertyController Implementation Analysis

## What It's Checking For

### **Type Categories**
1. **Primitives & Simple Types**: `int`, `bool`, `double`, `decimal`, `string`, `enum` → Treated as terminal (leaf) nodes
2. **Collections**: Arrays, `List<T>`, `IEnumerable<T>` → Examines element type
3. **Umbraco Types**: `PublishedContentModel`, `OverridableBlockListModel`, `OverridableBlockGridModel` → **Skipped entirely**
4. **Custom Classes/Interfaces**: Recursively explored
5. **System Types**: Those starting with `"System."` → Added as-is (fallback)

### **Safety Mechanisms**
- **Depth limit**: Maximum 5 levels of nesting
- **Cycle detection**: Tracks current path using `Stack<Type>` to prevent `A → B → A` loops
- **Scope**: Only public instance properties via `BindingFlags.Public | BindingFlags.Instance`

---

## What It Will Miss

### **1. Nullable Types**
```csharp
public int? NullableCount { get; set; }
public decimal? OptionalPrice { get; set; }
```
These are **NOT** treated as primitives. `int?` is actually `Nullable<int>`, a generic struct, so it hits the **fallback case** and gets added by name only—it won't recognize it as a simple type.

### **2. DateTime, DateTimeOffset, Guid, TimeSpan**
```csharp
public DateTime CreatedDate { get; set; }
public Guid Id { get; set; }
```
These are `System.` types but **not primitives**, so they hit the fallback and get added as property names without recursion. This is probably correct, but inconsistent with how `decimal` is explicitly handled.

### **3. Tuples and ValueTuples**
```csharp
public (string Name, int Age) PersonInfo { get; set; }
```
These won't be properly decomposed—likely treated as `System.ValueTuple` and added as-is.

### **4. Dictionaries**
```csharp
public Dictionary<string, string> Metadata { get; set; }
```
`Dictionary<TKey, TValue>` implements `IEnumerable<KeyValuePair<TKey, TValue>>`, so it will recurse into `KeyValuePair<string, string>` and return:
```
Metadata.Key
Metadata.Value
```
This is **probably not useful** for validation binding, as you'd normally want `Metadata["someKey"]` syntax.

### **5. Non-Generic IEnumerable**
```csharp
public ArrayList LegacyItems { get; set; }
public IEnumerable NonGenericCollection { get; set; }
```
`GetEnumerableElementType` returns `null` for these, so they're added as a single property name without recursion.

### **6. Private/Protected Properties**
```csharp
private string InternalId { get; set; }
protected int ProtectedCount { get; set; }
```
Correctly excluded due to `BindingFlags.Public`, but also means **internal state** won't be discovered.

### **7. Static Properties**
```csharp
public static string SharedSetting { get; set; }
```
Excluded by `BindingFlags.Instance`.

### **8. Read-Only Properties (by design)**
Actually **included** since there's no check for `CanWrite`. This might be intentional for validation display, but these can't be bound:
```csharp
public string FullName => $"{FirstName} {LastName}";
```

---

## What's Likely to Return Unexpected Results

### **1. Collection Edge Cases**
For a collection of primitives, it returns the collection name:
```csharp
public List<string> Tags { get; set; }
// Returns: ["Tags"]
```
But for complex types, it recurses:
```csharp
public List<Address> Addresses { get; set; }
// Returns: ["Addresses.Line1", "Addresses.City", "Addresses.Postcode"]
```
This is **inconsistent** from a binding perspective—you'd need `Addresses[0].City` for actual model binding.

### **2. Interface Properties**
```csharp
public ICustomInterface Metadata { get; set; }
```
It **will recurse** into the interface definition, showing properties defined by the interface. However, if the concrete implementation has additional properties, they **won't be discovered**.

### **3. Depth Limit Truncation**
With 5+ levels of nesting:
```csharp
public Level1 { Level2 { Level3 { Level4 { Level5 { Level6 { ... } } } } } }
```
Properties beyond `Level5` are **silently ignored**. No indication that truncation occurred.

### **4. Circular References**
```csharp
public class Node {
    public string Name { get; set; }
    public Node Parent { get; set; }
}
```
The cycle detection stops the loop, but **only on the current recursion path**. If you reach the same type via different paths:
```
Root.NodeA.Child (Node type)
Root.NodeB.Child (Node type)  ← Will still expand even though Node was seen before
```
This is actually **correct behavior** but could lead to property duplication.

### **5. Struct vs Class**
Structs are treated the same as classes and will be recursed into, which is probably correct but could hit the depth limit unexpectedly with nested value types.

### **6. Assembly Loading Side Effects**
```csharp
Assembly.LoadFrom(file)
```
This loads **every DLL** in the directory (except `System.*` and `Microsoft.*`). This could:
- Trigger static constructors
- Load dependencies into the AppDomain
- Cause performance issues with many DLLs
- Fail if DLLs are locked or incompatible

---

## Summary

The implementation is **sophisticated** for basic scenarios but has quirks around:
- **Nullable types** (not recognized as primitives)
- **Dictionaries** (returns `Key`/`Value` instead of indexed access)
- **Collection indexing** (returns `Items.Name` not `Items[0].Name`)
- **DateTime/Guid** (inconsistent with `decimal` handling)
- **Assembly scanning overhead** (loads all DLLs unnecessarily)

These issues may or may not matter depending on how the property paths are consumed downstream.

GovUk.Frontend.Umbraco.Tests/Controllers/ModelPropertyConrollerTests.cs
```diff
@@ -0,0 +1,702 @@
+﻿using GovUk.Frontend.Umbraco.Validation;
+using NUnit.Framework;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace GovUk.Frontend.Umbraco.Tests.Validation
```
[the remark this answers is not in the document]
sussexrick: @dcr670 It's got a point here. The code under test is in a `Validation` namespace and folder so the tests should match that.

GovUk.Frontend.Umbraco/Validation/ModelPropertyController.cs
```diff
@@ -1,17 +1,15 @@
...
+using System.Collections;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using ThePensionsRegulator.Umbraco.Blocks;
 using Umbraco.Cms.Core.Models.PublishedContent;
 using Umbraco.Cms.Web.BackOffice.Controllers;
 using Umbraco.Cms.Web.Common.Attributes;
 using Umbraco.Cms.Web.Common.Controllers;
 using Umbraco.Extensions;
+[assembly: InternalsVisibleTo("GovUk.Frontend.Umbraco.Tests.ModelPropertyControllerTests")]
```
sussexrick: [We already have this](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/v13/GovUk.Frontend.Umbraco/InternalsVisibleTo.cs).

GovUk.Frontend.Umbraco/Validation/ModelPropertyController.cs
```diff
@@ -44,15 +40,102 @@ public IEnumerable<string> ForDocumentType(string alias)
...
+                        return propNames;
                     }
                 }
+            
             }
 
             return Array.Empty<string>();
         }
+
+        public void CollectProperties(Type type, string prefix, int depth, int maxDepth, Stack<Type> pathStack, List<string> propNames)
```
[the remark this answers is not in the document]
sussexrick: @dcr670 I agree with this. Nullable reference types should be enough on their own to make it clear at compile time whether null is acceptable and possible.

sussexrick:
> @sussexrick I've opened a new pull request, #697, to work on those changes. Once the pull request is ready, I'll request review from you.

@dcr670 There's some useful feedback in here to apply, about using `DirectorySeparatorChar` and catching potential exceptions coming from `Assembly.LoadFrom()`.

GovUk.Frontend.Umbraco/Validation/ModelPropertyController.cs
```diff
@@ -44,15 +48,153 @@ public IEnumerable<string> ForDocumentType(string alias)
...
+
+                // Interfaces and user-defined classes: recurse into them
+                if (propType.IsInterface || (propType.IsClass && !propType.FullName!.StartsWith("System.", StringComparison.Ordinal)))
+                {
+                    CollectProperties(propType, fullName, depth + 1, maxDepth, pathStack, propNames);
+                    continue;
+                }
+
+                // Handle tuples: expand tuple element names
+                if (IsTupleType(propType))
```
sussexrick: It's good to support tuples but there are no tests covering this.

GovUk.Frontend.Umbraco.Tests/Validation/ModelPropertyControllerTests.cs
```diff
@@ -0,0 +1,702 @@
...
+            var propNames = new List<string>();
+            var pathStack = new Stack<Type>();
+            var expected = new[] { "Status" };
+
+            // Act
+            _controller.CollectProperties(typeof(ModelWithEnum), string.Empty, 0, 5, pathStack, propNames);
+
+            // Assert
+            Assert.That(propNames, Is.EqualTo(expected));
+        }
```
sussexrick, line 173: Whilst I agree with your comment that this doesn't need to exhaustively test every possible thing you could do in C#, it is missing tests for types that we commonly use as properties on view models:

int?
decimal?
DateTime
DateTime?
DateTime
DateOnly?
DateTimeOffset
DateTimeOffset?
Guid
Guid?

sussexrick (changes requested):

sussexrick (approved):

Issue #664, opened 2026-03-02
Heading size control in Accordion does not support sitewide setting


sussexrick:
See #642 for context.

### Unexplained SASS change

You don't say why the `font-size: inherit` setting was added. In testing I don't see it making a difference. Without it the font-size is set by GOV.UK code, which is what we want.

### No GOV.UK support

This is not a TPR-specific behaviour, it's a GOV.UK one. As such the changes should be in the GOV.UK accordion view, and the TPR one should've remained as just a reference to the GOV.UK one as it was before. Currently if you use an accordion in a GOV.UK block list/grid it will still have the bug you were trying to solve.

### Heading sizes are fixed

You haven't addressed the fact that the default size of headings can change, depending on a sitewide setting about which size to start with. If you go to the home page of the example app in the backoffice you will find a setting where you can change the starting size for headings across the site:

<img width="459" height="320" alt="Settings tab of the home page in the backoffice" src="https://github.com/user-attachments/assets/4e2e256d-eb52-40a0-b03e-619f258d81b8" />

This is set to `govuk-heading-l` in the example app. If you change it to `govuk-heading-xl` you will see heading sizes change across the app, including the h1 on the accordion page. Your accordion heading size does not change. This is because you need to get the class name from `IGovUkHeadingClassProvider` as shown in PR #623 that I pointed you to.

## Size customisation

This is optional but PR #623 shows how you can also add the option for users to override the default size, by adding a separate property with a dropdown list of sizes, and then only applying the class from `IGovUkHeadingClassProvider` if nothing was selected from the list:

```csharp
var headingClass = Model.Settings.Value<string>(TprPropertyAliases.VideoHeadingSize);
    if (string.IsNullOrWhiteSpace(headingClass))
    {
       /// get size from IGovUkHeadingClassProvider
    }
```

This would add greater flexibility and be consistent with other similar support for customising heading levels within components.

_Originally posted by @sussexrick in https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/issues/642#issuecomment-3973750905_

PR #700, opened 2026-03-12
Accordion heading size control improvements (v13)

Ethan-Lambourne:
These changes are intended to solve the open issue: https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/issues/664

Closes https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/issues/664

In this PR:
- Moved changes to the GovUk version of the accordion instead of the TPR version;
- Included support for default heading size changes by getting the class names from `IGovUkHeadingClassProvider`;
- Included additional size customisation for the accordion headers by adding a dropdown that allows the user to apply a govuk heading class to the accordions headings, falling back to the class name obtained via the `IGovUkHeadingClassProvider` if left blank.

The `font-size: inherit` sass change has been left in because without it the font size for the heading is overridden by the `govuk-accordion__section-button` class inherited from the main https://github.com/x-govuk/govuk-frontend-aspnetcore repository - so without this extra styling no matter what classes are applied to the heading element within the accordion, the font size for the heading would never change.


sussexrick:
@Ethan-Lambourne A little Github tip - if you put "Closes #664" or "Fixes #664" instead of your version it'll close the issue when the PR is merged.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/DataTypes/GOVUKHeadingLevelAccordion.config
```diff
@@ -8,6 +8,10 @@
   </Info>
   <Config><![CDATA[{
   "Items": [
+    {
+      "id": 5,
+      "value": "Heading 1"
```
sussexrick: Have we identified a use case where h1 would be appropriate? Normally multiple h1 elements on a page is considered an accessibility problem.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/govukaccordionsettingsshared.config
```diff
@@ -21,6 +21,22 @@
...
   <GenericProperties>
+    <GenericProperty>
+      <Key>0d413e7e-d713-4dcb-9525-464ff9d674b0</Key>
+      <Name>Heading class</Name>
+      <Alias>headingClass</Alias>
+      <Definition>6ffa4d83-3f52-4c27-abb2-8b85d8232f4b</Definition>
+      <Type>Umbraco.DropDown.Flexible</Type>
+      <Mandatory>false</Mandatory>
+      <Validation></Validation>
+      <Description><![CDATA[Defaults to "govuk-heading-m" if left blank.]]></Description>
```
sussexrick, line 32: This text is consistent with the section cards component, but the other field isn't - that should be "Defaults to 'Heading 2' if left blank." rather than "Defaults to 2 if left blank."

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAccordion.cshtml
```diff
@@ -1,45 +1,3 @@
...
-                <govuk-accordion-item-summary>@(section.Content.Value<string>(PropertyAliases.AccordionSectionSummary))</govuk-accordion-item-summary>
-            }
-            <govuk-accordion-item-content>
-                @await Html.PartialAsync("GOVUK/BlockGrid", childBlocks)
-            </govuk-accordion-item-content>
-        </govuk-accordion-item>
-    }
-</govuk-accordion>
\ No newline at end of file
+<partial name="GOVUK/GovUkAccordion" />
```
sussexrick, line 3: You've moved most of the functionality back to GOV.UK level which is great, but the issue you identified which you fixed with `font-size: inherit` is still only applied to the TPR version. The same issue happens in GOV.UK styles.

sussexrick (changes requested):

GovUk.Frontend.AspNetCore.Extensions/Styles/_govuk-accordion.scss
sussexrick, line 1: The media query in here should be using a helper rather than hard-coding a breakpoint:

```scss
@include govuk-media-query($from: tablet) {
    // your changes
}

sussexrick (changes requested):
This PR looks good now, but I just spotted one bit of tidying up from the earlier PR on this.

sussexrick (approved):

PR #701, opened 2026-03-13
#256048 - Tables do not adapt well on small screens v13

AlexBeesley:
To support accessibility improvements, on the corporate website, in response to [this](https://tpr-ae.a11y.civiccomputing.com/december2025/tables-do-not-adapt-well-on-small-screens-m.html#tables-do-not-adapt-well-on-small-screens-m) Civic report, this pull request adds a new feature that enables CSV download functionality for HTML tables rendered in block grid content.

**Changes**

* Added the `EnableTableCsvDownload` option to the `GovUkFrontendUmbracoOptions` class, allowing the CSV download button to be enabled or disabled for tables in block grid content.
* Updated service registration in `Startup.cs` to pass the new option to both `AddTprFrontendUmbraco` and `AddGovUkFrontendUmbraco`, ensuring the feature is enabled when configured.
* Introduced `table-csv-download.js`, which adds a CSV download button to each table, generates CSV from table data, and triggers file download. The button text can be customised via a meta tag.
* Modified the body closing partial (`UmbracoBodyClosing.cshtml`) to conditionally include the CSV download script based on the option value, ensuring the script is only loaded when needed.
* Added a dictionary configuration file for the CSV download button text, supporting both English and Welsh translations.
* Registered the new JavaScript asset bundle for the CSV download functionality in `ApplicationBuilderExtensions.cs`.


GovUk.Frontend.Umbraco/wwwroot/govuk/table-csv-download.js
```diff
@@ -0,0 +1,94 @@
+(function () {
+    "use strict";
+
+    var DEFAULT_FALLBACK_BUTTON_TEXT = "Download table data (CSV)";
```
sussexrick: `var` is legacy JS syntax. Always use `let` or `const` (`const` here).

GovUk.Frontend.Umbraco/wwwroot/govuk/table-csv-download.js
```diff
@@ -0,0 +1,94 @@
...
+
+    function getButtonText() {
+        var meta = document.querySelector('meta[name="table-csv-download-button-text"]');
```
sussexrick: Using a meta tag is also a bit old-school. If you want a page-wide setting accessible to JS a more modern approach (and also consistent with other code in govuk-frontend and this repo) would be to use a `data-*` attribute. 

For example, you could look for `<body data-tpr-table-csv-download-text="My custom text">`. Note the `tpr` namespacing so there's no chance it could clash with some imported library.

GovUk.Frontend.Umbraco/wwwroot/govuk/table-csv-download.js
```diff
@@ -0,0 +1,94 @@
...
+        if (value.indexOf(",") !== -1 || value.indexOf('"') !== -1 || value.indexOf("\n") !== -1 || value.indexOf("\r") !== -1) {
+            return '"' + value.replace(/"/g, '""') + '"';
+        }
+        return value;
+    }
+
+    function tableToCsv(table) {
+        var csv = [];
+        var rows = table.querySelectorAll("tr");
+        for (var i = 0; i < rows.length; i++) {
```
sussexrick: Have you considered what happens if there are merged cells or rows? It looks to me as if it won't crash, but it would produce potentially invalid CSV data, eg merged cells in a row would lead to different numbers of columns in each row of a CSV. Is this acceptable? It would be very difficult to represent these kinds of tables in CSV, but would it be better return potentially bad data or to detect merged cells and not offer a download?

GovUk.Frontend.Umbraco/wwwroot/govuk/table-csv-download.js
```diff
@@ -0,0 +1,94 @@
+(function () {
```
sussexrick: This JS needs to be covered by Jest tests.

GovUk.Frontend.Umbraco/wwwroot/govuk/table-csv-download.js
```diff
@@ -0,0 +1,94 @@
...
+        link.style.display = "none";
+        document.body.appendChild(link);
+        link.click();
+        document.body.removeChild(link);
+        URL.revokeObjectURL(url);
+    }
+
+    function initTableCsvDownload() {
+        var buttonText = getButtonText();
+        var tables = document.querySelectorAll("table");
```
sussexrick: This should probably look for `.govuk-table` just in case there's some other table on the page, outside of the main content, for some reason.

GovUk.Frontend.Umbraco/wwwroot/govuk/table-csv-download.js
```diff
@@ -0,0 +1,94 @@
...
+
+        for (var i = 0; i < tables.length; i++) {
+            var table = tables[i];
+
+            var caption = table.querySelector("caption");
+            var fileName = sanitizeFileName(caption ? caption.innerText || caption.textContent : null);
+
+            var wrapper = document.createElement("div");
+            wrapper.className = "tpr-table-wrapper";
```
sussexrick: The ownership of this feature seems a bit confused. It's in the GOV.UK project, but the class identifies it as "tpr". Since it's not part of the GOV.UK design system I would say it's TPR and belongs in ThePensionsRegulator.Frontend.

GovUk.Frontend.Umbraco/wwwroot/govuk/table-csv-download.js
```diff
@@ -0,0 +1,94 @@
...
+
+    function downloadCsv(csvContent, fileName) {
+        var BOM = "\uFEFF";
+        var blob = new Blob([BOM + csvContent], { type: "text/csv;charset=utf-8;" });
+        var url = URL.createObjectURL(blob);
+        var link = document.createElement("a");
+        link.setAttribute("href", url);
+        link.setAttribute("download", fileName + ".csv");
+        link.style.display = "none";
+        document.body.appendChild(link);
```
sussexrick: You don't need to add this link to `document.body()` to click it (according to CoPilot). You lose compatibility with some older browsers, but we don't support very old browsers anyway. 

Avoiding DOM manipulation would make your code easier to test.

GovUk.Frontend.Umbraco/wwwroot/govuk/table-csv-download.js
```diff
@@ -0,0 +1,94 @@
...
+            wrapper.appendChild(table);
+
+            var button = document.createElement("button");
+            button.type = "button";
+            button.className = "govuk-button govuk-button--secondary";
+            button.setAttribute("data-module", "govuk-button");
+            button.textContent = buttonText;
+
+            (function (tbl, fn) {
+                button.addEventListener("click", function () {
```
sussexrick: If you use `let i = 0` in the definition of the `for` loop, instead of `var`, then according to CoPilot you no longer need to wrap this event listener in a self-calling function, so your code could be simpler.

Image

GovUk.Frontend.Umbraco/GovUkFrontendUmbracoOptions.cs
```diff
@@ -6,5 +6,13 @@ public class GovUkFrontendUmbracoOptions
         /// Render a .govuk-width-container around each block, for use when the layout page does not contain .govuk-width-container.
         /// </summary>
         public bool RenderWidthContainerForBlocks { get; set; }
+
+        /// <summary>
+        /// Enable a CSV download button on all HTML tables rendered in block grid content.
+        /// The button text defaults to "Download table data (CSV)" but can be customised by adding
+        /// a &lt;meta name="table-csv-download-button-text" content="your text" /&gt; tag to the page.
+        /// Tables already inside a .tpr-table-wrapper element are skipped.
```
sussexrick: Are they skipped? I don't see any code doing that.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/table.config
```diff
@@ -61,7 +61,7 @@
...
+        "markup": "<table><caption>Table caption</caption>\n<thead>\n<tr>\n<th scope=\"col\">Column header 1</th>\n<th scope=\"col\">Column header 2</th>\n<th scope=\"col\">Column header 3</th>\n<th class=\"govuk-table__header--numeric\" scope=\"col\">Numeric data</th>\n</tr>\n</thead>\n<tbody>\n<tr>\n<th scope=\"row\">Row header 1</th>\n<td>Data cell 1</td>\n<td>Data cell 2</td>\n<td class=\"govuk-table__cell--numeric\">1</td>\n</tr>\n<tr>\n<th scope=\"row\">Row header 2</th>\n<td>Data cell 3</td>\n<td>Data cell 4</td>\n<td class=\"govuk-table__cell--numeric\">2</td>\n</tr>\n<tr>\n<th scope=\"row\">Row header 3</th>\n<td>Data cell 5</td>\n<td>Data cell 6</td>\n<td class=\"govuk-table__cell--numeric\">3</td>\n</tr>\n</tbody>\n</table>\n<p>The CSV download button above is automatically added below every table in block grid content provided<span>&nbsp;</span><strong>EnableTableCsvDownload&nbsp;</strong>is set to true in<strong>&nbsp;GovUkFrontendUmbracoOptions</strong>; this is for improved accessibility.</p>",
```
sussexrick: > "below every table in block grid"

It's not just block grid though. This is actually a lot better than that. It'll work with or without Umbraco, and doesn't care if it's a block grid or list or static HTML rendering the table.

As such it shouldn't live in an Umbraco specific project because there's nothing Umbraco specific about it (aside from your dictionary support... which I don't actually see used anywhere). ThePensionsRegulator.Frontend would be a good home for this.

sussexrick (changes requested):
This is a clever solution and it works. 

I just think it needs some polish and a new home. Saying it's specific to Umbraco is under-selling it. It's better than that.

It needs tests. It needs documentation. And it needs to know what it is - a TPR feature that doesn't require Umbraco.

sussexrick:
@AlexBeesley Can we conform to [this UK gov standard for CSV data](https://www.gov.uk/government/publications/recommended-open-standards-for-government/tabular-data-standard)? Maybe we already do - if so let's document that.

GovUk.Frontend.Umbraco/wwwroot/govuk/table-csv-download.js
```diff
@@ -0,0 +1,94 @@
...
+
+            var wrapper = document.createElement("div");
+            wrapper.className = "tpr-table-wrapper";
+            table.parentNode.insertBefore(wrapper, table);
+            wrapper.appendChild(table);
+
+            var button = document.createElement("button");
+            button.type = "button";
+            button.className = "govuk-button govuk-button--secondary";
+            button.setAttribute("data-module", "govuk-button");
```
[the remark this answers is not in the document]
sussexrick: What function is the wrapper serving? It doesn't have any CSS applied or JS targeting it.

sussexrick (changes requested):

docs/components/tpr-table-csv-download.md
```diff
@@ -0,0 +1,99 @@
+# TPR table CSV download
+
+Adds a CSV download button below each `.govuk-table` element on the page. This is a TPR accessibility feature that allows users to download table data in CSV format, which can be easier to read in a spreadsheet application than a complex HTML table on a small screen.
```
sussexrick, line 3: Good docs, but they're not linked from anywhere. You need to a add a link in the root README.md.

GovUk.Frontend.ExampleApp/Views/Home/Index.cshtml
```diff
@@ -48,6 +48,7 @@
     <li><a class="govuk-link" href="/relatedlinks">related links</a></li>
     <li><a class="govuk-link" href="/searchresults">search results</a></li>
     <li><a class="govuk-link" href="/sectioncards">section cards</a></li>
+    <li><a class="govuk-link" href="/table">table</a></li>
```
sussexrick: Thank you for adding a table example to this app, but table is part of the GOV.UK design system so it belongs in the upper section

GovUk.Frontend.ExampleApp/Views/Table/Index.cshtml
```diff
@@ -0,0 +1,63 @@
+@{
+    ViewData["Title"] = "Table CSV download";
```
sussexrick: I think the title is just "Table" because that's what you linked to, and because this is demonstrating the table styling, not just the CSV function.

GovUk.Frontend.ExampleApp/Views/Table/Index.cshtml
```diff
@@ -0,0 +1,63 @@
...
+}
+<div class="govuk-width-container">
+    <span class="govuk-caption-xl">Example application</span>
+    <div class="govuk-grid-row govuk-grid-row--tpr-divider">
+        <div class="govuk-grid-column govuk-grid-column-full">
+            <h1 class="govuk-heading-l">@ViewData["Title"]</h1>
+        </div>
+    </div>
+
+    <p class="govuk-body">
```
sussexrick: All this content should be in a two-thirds column

GovUk.Frontend.ExampleApp/Views/Table/Index.cshtml
```diff
@@ -0,0 +1,63 @@
...
+                <th scope="col" class="govuk-table__header">Expenses</th>
+                <th scope="col" class="govuk-table__header govuk-table__header--numeric">Profit</th>
+            </tr>
+        </thead>
+        <tbody class="govuk-table__body">
+            <tr class="govuk-table__row">
+                <th scope="row" class="govuk-table__header">Q1 2025</th>
+                <td class="govuk-table__cell">£12,500</td>
+                <td class="govuk-table__cell">£8,200</td>
+                <td class="govuk-table__cell govuk-table__cell--numeric">£4,300</td>
```
sussexrick: I would apply the numeric class to columns 2, 3 and 4 as they are all numeric. Or change the content to be the same as the Umbraco app for which these styles are correct.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/table.config
sussexrick: In this Umbraco context it would still be "block grid or block list"

sussexrick: "The CSV download button above" is not actually there. You need to enable the option in `Startup.cs` and also add it to the bundle in `ThePensionsRegulator.Frontend.Umbraco.ApplicationBuilderExtensions`.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/table.config
```diff
@@ -61,7 +61,7 @@
...
         "blocks": {
```
sussexrick, line 65: The setting is in `TprFrontendOptions`, not `GovUkFrontendUmbracoOptions`.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/table-csv-download-button-text.config
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Dictionary Key="a7e3b2c1-4d5f-6a89-b0c1-d2e3f4a5b6c7" Alias="Table CSV Download Button Text" Level="0">
```
sussexrick, line 2: This is still not being used anywhere. You could either remove it, or demonstrate it being used by feeding the value to the data attribute on the body tag in `_Layout.cshtml`.

ThePensionsRegulator.Frontend/__tests__/tpr-table-csv-download.tests.js
```diff
@@ -0,0 +1,299 @@
...
+  mergedCells = false,
+} = {}) => {
+  const headerHtml = headers
+    .map((h) => `<th scope="col">${h}</th>`)
+    .join("");
+  const rowHtml = rows
+    .map((r) => `<tr>${r.map((c) => `<td>${c}</td>`).join("")}</tr>`)
+    .join("");
+
+  let mergedAttr = "";
```
sussexrick: `mergedAttr` is unused and can be removed

ThePensionsRegulator.Frontend/__tests__/tpr-table-csv-download.tests.js
```diff
@@ -0,0 +1,299 @@
...
+    document.body.innerHTML = `<div>${createTable()}</div>`;
+    initTableCsvDownload();
+    const table = document.querySelector(".govuk-table");
+    const button = table.nextElementSibling;
+    expect(button).not.toBeNull();
+    expect(button.hasAttribute("data-tpr-table-csv-button")).toBe(true);
+  });
+
+  it("should derive the file name from the table caption", () => {
+    document.body.innerHTML = createTable({ caption: "My Report 2024" });
```
sussexrick, line 278: I don't think this is testing what it says it is. What it actually tests appears to be a duplicate of the test above.

sussexrick (changes requested):
This is looking a lot better. The JS looks much more robust and the tests look good. Just a few small things to tidy up.

ThePensionsRegulator.Frontend/Views/Shared/TPR/BodyClosing.cshtml
```diff
@@ -1,3 +1,4 @@
 ﻿<script src="~/govuk-frontend.min.js?v=5.13.0" type="module"></script>
 <script src="/_content/ThePensionsRegulator.GovUk.Frontend/govuk/govuk-js-init.js" type="module"></script>
-<script src="/_content/ThePensionsRegulator.Frontend/tpr/tpr-back-to-top.js"></script>
\ No newline at end of file
+<script src="/_content/ThePensionsRegulator.Frontend/tpr/tpr-back-to-top.js"></script>
+<script src="/_content/ThePensionsRegulator.Frontend/tpr/tpr-table-csv-download.js" type="module"></script>
```
sussexrick: You're not respecting the option in `TprFrontendOptions`.

AlexBeesley:
@copilot resolve the merge conflicts in this pull request

AlexBeesley:
@copilot resolve the merge conflicts in this pull request

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Dictionary/table-csv-download-button-text.config
sussexrick: The dictionary key you're using in _Layout.cshtml is not the same as the one in this dictionary entry. The example app still isn't using this.

sussexrick (changes requested):

sussexrick (approved):

PR #704, opened 2026-03-16
I have simply changed the toggle button text on the model to conform …

georgeboydratcliff:
I have simply changed the toggle button text on the model to conform with what was requested in the accessibility ticket (ie not to have separate expand and collapse text, but to have a 'toggle' text). The 'expandED' and 'collapsED' text remains the same as I believe it to be a far superior solution to what was requested in the ticket!


villarin (approved):

sussexrick:
Don't forget to create another PR to merge this change into develop.

villarin (approved):

PR #715, opened 2026-03-17
#259676 Breadcrumb does not adapt correctly on zoom

Max-Emery:
Why:

- Breadcrumb component was flagged as not adapting correctly on zoom, this meant the breadcrumb was obstructed and required horizontal scrolling to view.

image

In this PR: 

- Added additional styling to allow the breadcrumb to collapse down underneath parent, when under high magnification on smaller screens

 
breadcrumb solution

### DIFF
--- a/ThePensionsRegulator.Frontend/Styles/_tpr-breadcrumb.scss
+++ b/ThePensionsRegulator.Frontend/Styles/_tpr-breadcrumb.scss
@@ -1,0 +1,12 @@
+﻿
+
+@media(max-width: 15em){
+
+    .govuk-breadcrumbs__list {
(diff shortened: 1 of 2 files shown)


villarin (approved):

sussexrick:
@Max-Emery This looks like a good improvement, but why limit it to when TPR styling is in use? Breadcrumbs is a GOV.UK component. A better solution would be to run the example app with "UseTPRStyles: false" in appsettings.json and see whether the problem also exists when just GOV.UK styles are in play. If so, put your fix in ThePensionsRegulator.GovUk.Frontend rather than ThePensionsRegulator.Frontend.

PR #717, opened 2026-03-18
#255078 - ROS-02 Main heading and content do not adapt correctly when page is zoomed (v17)

Max-Emery:
Why:

- When viewing a page at 200% zoom on smaller screens, main content and headings would not adapt correctly forcing the user to scroll horizontally to view.

In this PR:

- Added TPR specific styling of GOVUK typography, ensuring relevant classes enforce word-break to prevent horizontal scrolling.


ThePensionsRegulator.Frontend/Styles/tpr.scss
```diff
@@ -71,6 +71,7 @@
 @import '_tpr-table.scss';
 @import '_tpr-tag.scss';
 @import '_tpr-textarea.scss';
+@import '_tpr-typography.scss';
```
sussexrick: This is a good fix for the problem, but it's not a problem caused by TPR styles. The problem also exists when using just GOV.UK styles: 

Long words in h1 causing horizontal scrolling

It would make sense to put the fix in `ThePensionsRegulator.GovUk.Frontend` rather than `ThePensionsRegulator.Frontend`, so that pure GOV.UK styling benefits too. 

We [already have similar fixes](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/ThePensionsRegulator.GovUk.Frontend/Styles/_govuk-panel.scss) in there for issues that we see in pure GOV.UK styling.

ThePensionsRegulator.Frontend/Styles/_tpr-typography.scss
```diff
@@ -0,0 +1,8 @@
+.govuk-phase-banner__text,
+.govuk-caption-l,
```
sussexrick: You need to support `govuk-caption-xl` and `govuk-heading-xl` too. Also `govuk-caption-m` exists, so why not support that given you've also chosen to support headings at that size?

sussexrick (changes requested):

sussexrick (approved):

PR #719, opened 2026-03-18
#259676 Breadcrumb does not adapt correctly on zoom (v17)

Max-Emery:
Why:

- Breadcrumb component was flagged as not adapting correctly on zoom, this meant the breadcrumb was obstructed and required horizontal scrolling to view.

image

In this PR:

- Added additional styling to allow the breadcrumb to collapse down underneath parent, when under high magnification on smaller screens.

image


sussexrick:
@Max-Emery This looks like a good improvement, but why limit it to when TPR styling is in use? Breadcrumbs is a GOV.UK component. A better solution would be to run the example app with "UseTPRStyles: false" in appsettings.json and see whether the problem also exists when just GOV.UK styles are in play. If so, put your fix in ThePensionsRegulator.GovUk.Frontend rather than ThePensionsRegulator.Frontend.

ThePensionsRegulator.GovUk.Frontend/Styles/_govuk-breadcrumb.scss
```diff
@@ -0,0 +1,8 @@
+
+
+@media(max-width: 15em) {
```
sussexrick: What purpose is the media query serving? I did a little bit of testing without it and `flex-wrap: wrap` didn't seem to cause a problem at larger sizes.

Max-Emery: The only reason for inclusion was that would wouldn't get as much of the breadcrumb collapsing down - with the current page node wrapping underneath before the home node had started collapsing, without the media query the fix is more robust but just means that most mobile views would see it collapsed underneath rather than just at high zoom scenarios. Happy to amend if you think its not needed.

sussexrick: I noticed your screenshot had a long title for the home page, whereas I was testing with 'Home', so I've done some further testing. I think this is the appropriate media query and selector:

```sass
@include govuk-media-query($until: tablet) {
    .govuk-breadcrumbs--collapse-on-mobile .govuk-breadcrumbs__list {
        flex-wrap: wrap;
    }
}
```

With the media query active to larger sizes you get a clearer layout:

Second breadcrumb wraps below the first

...compared to when it's inactive:

Each breadcrumb wraps individually

The media query and selector I've used matches that used by GOV.UK Frontend to apply `display: flex`. Without `display: flex` the new property `flex-wrap: wrap` is ignored anyway. This ensures there are two scenarios to test (flex on or off) as opposed to three scenarios (flex on with wrap, flex on without wrap, flex off).

sussexrick (changes requested):

ThePensionsRegulator.GovUk.Frontend/Styles/_govuk-breadcrumb.scss
Max-Emery: Have updated with your suggested media query and selector.

sussexrick (approved):

PR #720, opened 2026-03-18
#255083 Anchor links are not tagged as navigation elements (v13)

Ethan-Lambourne:
In this PR:
- Altered the html generation for the `TPR Related Links` component to make it wrapped within a `<nav>` element;
- Made the decorative blue bar that sits above the the heading toggleable;
- Updated the example app implementation for the 'TPR Related Links' component as well as the markdown documentation to reflect the new changes;
- Created the `TPR Navigational Links` component to provide an easy way for users to create a list of links wrapped within a `<nav>` element with dynamic accessibility labelling. The main input for the `TPR Navigational Links` component is a custom rich text editor which allows for users to easily copy/paste the existing implementations of in-page anchor links into it.

These changes are related to the work item [AB#255083](https://dev.azure.com/thepensionsregulator/TPR/_workitems/edit/255083) which outlines the changes the external accessibility auditor CIVIC have requested.


sussexrick:
When referencing Azure DevOps work items please use the syntax AB#255083, which causes Azure DevOps and Github to actually link the PR to the work item. That works when used in PR descriptions and commit messages. Not sure if it'll work from this comment.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRRelatedLinks.cshtml
```diff
@@ -6,13 +6,14 @@
 @addTagHelper *, ThePensionsRegulator.Frontend
 
 @{
-    var cssClasses = Model.Settings.Value<string>(GovUkPropertyAliases.CssClasses);
+    var cssClasses = Model.Settings.Value<bool>(TprPropertyAliases.RelatedLinksBlueBar) ? "blue-bar" : string.Empty;
```
sussexrick: Before this change every use of this component has the blue bar, but none of them have this property. 

When you add the property the value will be either "" or null, which will be `false` when converted to `bool` which means the blue bar is gone. The default needs to be to have it there, otherwise you'll have to find and update every page that uses this when updating an existing site.

This is the result on the example page after pulling this branch and then importing only the Settings, which should recreate the scenario of existing content.

No blue bar

docs/components/tpr-related-links.md
```diff
@@ -20,30 +20,34 @@ In ASP.NET applications the related links component is structured like so, repla
...
-    </li>
-    <li>
-      <a href="/" class="govuk-link">Example link 2</a>
-    </li>
-    <li>
-      <a href="/" class="govuk-link">Example link 3</a>
-    </li>
-  </ul>
-</div>
+<nav>
```
sussexrick: You could use `nav` instead of `div` rather than having both.

sussexrick: Each `nav` element needs to have an `aria-label` so that users can identify the purpose of it when there are multiple `nav` elements on the page.

Re-using the same component here means you can't just have a default `aria-label` that works most of the time (with the ability to change it). Instead you could at best have a default that worked half the time, and ask content designers to remember to change it the other half. And sometimes that would get missed.

This is why the use case you're working on was originally identified as a separate component in AB#206788. Looking at the comments there the story was closed without proper analysis that would've identified the need for the `nav` element. I think it still makes sense to make it a separate component. Although they'd be similar, the appearance is different and it would let you have a default `aria-label` that's right most of the time.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tprrelatedlinkssettings.config
```diff
@@ -22,6 +22,31 @@
     <AllowedTemplates />
   </Info>
   <Structure />
-  <GenericProperties />
-  <Tabs />
+  <GenericProperties>
+    <GenericProperty>
+      <Key>dd3a2e4c-b3be-40fd-83ee-f82471e88f63</Key>
+      <Name>Blue bar</Name>
+      <Alias>blueBar</Alias>
```
sussexrick: This isn't a future-proof way to handle this. What if the design changes and the bar is TPR purple? Now you have a `blueBar` property and a `.blue-bar` class that add a purple bar. When naming CSS classes and also properties like this, name them for the purpose, not for the appearance. 

The same goes for C# properties - calling the C# constant `RelatedLinksBlueBar` means that when the appearance changes you have to make a breaking change to the C# which, since we use [semantic versioning](https://semver.org), is a new major version of the package. A tweak to the appearance shouldn't cause that.

ThePensionsRegulator.Frontend/Styles/_tpr-related-links.scss
```diff
@@ -1,15 +1,17 @@
 ﻿@import '_tpr-variables.scss';
 
 .tpr-related-links {
-    border-top: solid 4px $tpr-colour-royal-blue;
     padding-top: govuk-spacing(4);
     @include govuk-responsive-margin(6, "bottom");
 }
 
+.tpr-related-links.blue-bar {
```
sussexrick: We follow govuk-frontend in using the [BEM naming convention](https://getbem.com/introduction/) for CSS classes. That means the class would be `tpr-related-links--something` rather than `blue-bar`.

sussexrick (dismissed):

Ethan-Lambourne:
@sussexrick I believe I've actioned your comments in the latest commits - I've updated the description of this PR to reflect the new changes.

villarin (approved):

PR #737, opened 2026-03-27
#259678 If you set any of the colours on the body or a element you must set all of them

Max-Emery:
Why:

Raised in [AB#259678](https://dev.azure.com/thepensionsregulator/TPR/_boards/board/t/Web%20Platform%20Team/Stories?System.Tags=Corp%20Web%20Accessibility%2CCorp%20Website%20Migration&workitem=259678) that not all colour styling had been set meaning that some content is not available for users with alternate browser colour schemes. 

Areas identified through testing with forced colours set to both light and dark mode include:

-  the logo (Header and footer)
- search icon 
- mobile menu toggle hamburger icon 
- mobile menu toggle focus style.

In this PR:

- Added support for different logo's displaying depending on users chosen colour scheme
- Added styling for light mode hamburger icon to ensure visibility
- Added styling for search icon to increase visibility with dark colour scheme 
- Moved focus to the outside of the toggle, this is consistent with focus style of the logo and ensure it is visible with forced colour schemes.

`Note`: Similar to the header a print logo was also added for the footer logo, this is to address a situation where a user may load the page without forced colours but decides to turn them on. Without the print logo the wrong logo will be used meaning its not visible. This is a bit of an edge case as this is not present if the page is loaded with forced colours already enabled.

### DIFF
--- a/ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
+++ b/ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
@@ -69,12 +69,18 @@
     display: flex;
 }
 
-    .tpr-mobile-menu__svg.tpr-mobile-menu__svg--hide {
-        display: none;
-    }
+.tpr-mobile-menu__svg.tpr-mobile-menu__svg--hide {
+    display: none;
+}
 
-.tpr-mobile-menu__toggle .tpr-header-menu__menu-icon {
+ .tpr-header-menu__menu-icon {
     fill: $tpr-colour-white;
+}
+
+@media (prefers-color-scheme: light) {
+    .tpr-header-menu__menu-icon {
+        fill: $tpr-colour-black;
+    }
 }
 
 .tpr-header-menu__nav-container {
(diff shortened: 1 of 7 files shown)


georgeboydratcliff (dismissed):

ThePensionsRegulator.Frontend/Styles/_tpr-header-menu.scss
```diff
@@ -18,8 +18,9 @@
     margin-bottom: -0.6rem;
     cursor: pointer;
 
-    &:focus {
-        outline: $govuk-focus-width solid $govuk-focus-colour;
+    &:focus-within {
+        box-shadow: 0 0 0 4px $tpr-colour-tangerine-yellow, 0 0 0 8px $tpr-colour-black;
+        outline: 3px solid transparent;
```
sussexrick: We should still use `$govuk-focus-width` here rather than hard-coding 3px.

sussexrick (commented):
One minor thing noted, otherwise looks good. Nice to see attention paid to forced colours mode, which is usually forgotten. 👍

sussexrick (approved):

Issue #739, opened 2026-03-30
Upgrade to govuk-frontend-aspnetcore v4.0.0 and GOV.UK Frontend v6.0.0 for v17


sussexrick:
https://github.com/x-govuk/govuk-frontend-aspnetcore/releases/tag/v4.0.0

PR #741, opened 2026-03-30
Upgrade to govuk-frontend-aspnetcore v4.0.0 and GOV.UK Frontend v6.0.0

sussexrick:
This PR upgrades our core dependencies to:

- [govuk-frontend-aspnetcore v4.0.0](https://github.com/x-govuk/govuk-frontend-aspnetcore/releases/tag/v4.0.0)
- [GOV.UK Frontend v6.0.0](https://github.com/alphagov/govuk-frontend/releases/tag/v6.0.0)

The first upgrade removes the `MergeCssClass` extension method we were using, so code is updated to use the standard `AddCssClass` instead.

The second introduces changes in the SASS code that are significant for us:
- We upgrade our SASS compiler package to get the latest version of the Dart SASS compiler.
- The `govuk-tint()` and `govuk-shade()` functions that we were using are gone, so define variables for the tints and shades we had previously defined using the functions, to leave the end result unchanged.
- Some core SASS variables defining colours are now set and consumed differently, using the `govuk-functional-colour()` function, so we update our code to leave the end result unchanged.
- A typo is corrected in the class name for the pagination ellipsis, so we update our customisation of the class to leave the end result unchanged.
- GOV.UK set disabled buttons to `0.5` opacity. We reset that to leave the end result unchanged.
- GOV.UK removed size `14` from their typography scale, so we do too. We were not using it.

 #739

### DIFF
--- a/ThePensionsRegulator.Frontend/ThePensionsRegulator.Frontend.csproj
+++ b/ThePensionsRegulator.Frontend/ThePensionsRegulator.Frontend.csproj
@@ -30,9 +30,9 @@
 	</ItemGroup>
 
 	<ItemGroup>
-		<PackageReference Include="AspNetCore.SassCompiler" Version="1.86.2" />
+		<PackageReference Include="AspNetCore.SassCompiler" Version="1.98.0" />
 		<PackageReference Include="BuildBundlerMinifier" Version="3.2.449" />
-		<PackageReference Include="GovUk.Frontend.AspNetCore" Version="3.5.1" />
+		<PackageReference Include="GovUk.Frontend.AspNetCore" Version="4.0.0" />
 		<PackageReference Include="HtmlAgilityPack" Version="1.12.4" />
 	</ItemGroup>
 
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <Version>10.0.0-beta003</Version>
+    <Version>10.0.0-beta004</Version>
     <InformationalVersion>$(Version)</InformationalVersion>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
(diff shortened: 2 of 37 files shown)


ThePensionsRegulator.Frontend/ThePensionsRegulator.Frontend.csproj
```diff
@@ -30,9 +30,9 @@
 	</ItemGroup>
```
pjoblingtpr, line 30: This project isn't building for me @sussexrick  :

Severity	Code	Description	Project	File	Line	Suppression State
Error (active)		"Unknown font size `14` - expected a point from the type scale."	ThePensionsRegulator.Frontend	C:\Dev\govuk-frontend-aspnetcore-extensions\ThePensionsRegulator.Frontend\EXEC	1	

Can you check to see if you can reproduce?

sussexrick: @pjoblingtpr I'm unable to reproduce, either on my existing install or a fresh clone. Can you find an exact location that's triggering the error? Size 14 is removed, so would make sense...

pjoblingtpr: line 25 actually in Directory.Build.Targets

<Exec Command="powershell -NonInterac .....

pjoblingtpr: I will try a fresh clone

sussexrick: I managed to reproduce it, and noticed it was reporting that it was using GOV.UK Frontend v5.14.0. 

I ran `git submodule update` at the root of the repo and now it reports GOV.UK Frontend v6.0.0 and works.

pjoblingtpr: any changes required?

sussexrick: Not if you're able to confirm it works for you

pjoblingtpr: working

pjoblingtpr (approved):

PR #745, opened 2026-04-01
Review which components are supported on TPR block list and block grid data types

### LINKED ISSUE #735: v17: Review components in TPR block grid and block list
The components listed in the 'TPR block grid' and 'TPR block list' should mostly be consistent, but they're out of sync.

Exceptions:
- layout components like 'two equal columns' belong in the grid only
- components intended always to go in the right column belong in the grid only

The 'Link' component has been added to the 'TPR block grid' data type at some point and should be removed. This is intended specifically for use within the 'Button group' component. Links can be added to a 'TPR block grid' or 'TPR block list' using the 'Text' component.

Review 'GOV.UK block grid' and 'GOV.UK block list' too, though these are less likely to be out of sync.

sussexrick:
Most components should be supported equally well on both the block list and block grid data types. This PR is the result of reviewing support across the two and resolving differences that were not appropriate.

1. Support section cards and featured image on 'TPR Block list'. These were supported on 'TPR Block grid' only. A simple update to the view makes them compatible with block list. I also moved them up the list a bit, above some lesser-used components. This is a non-breaking enhancement to block list.

2. Support row and column classes on image and featured image. This is a non-breaking enhancement for both block list and block grid, rather than a question of parity. By adding existing row and column compositions to existing settings types for these blocks, we gain the ability to apply custom classes to rows and columns, and do things like setting an image to be a half-column wide.

3. Removes the 'Link' component from 'TPR Block grid'. This is a breaking change for block grid, but it should never have been there. You can add a link to a block grid using a 'Text' block. The 'Link' component is intended specifically to enable the scenario of adding a link (like 'Cancel') to a button group, and is on the 'GOV.UK Buttons' data type for that reason.

4. Support 'Navigation links' on both 'TPR Block list' and 'TPR Block grid' (previously just grid). Update the settings for its rich text data type to include the two stylesheets we must always include, causing the style dropdown in the rich text area to offer the correct styles. Add a backoffice preview for the 'Navigational links' component.

This PR also includes other updates to uSync data from doing a fresh export of settings.

Closes #735


pjoblingtpr:
I get a build error with this Rick: "The type or namespace name 'NavigationalLinks' could not be found (are you missing a using directive or an assembly reference?)" - Umbraco ExampleApp

pjoblingtpr (approved):

PR #750, opened 2026-04-07
Address lookup JavaScript support (v13)

HarryAmmon:
Adds JavaScript support to the address lookup component. This uses the OS Places API.

Changes:

- Removes 'legend is page heading' support as this it proved too difficult to maintain the expected behaviour of this using JS components, as in some states the legend is removed
- Introduces the concept of 'primary' and 'secondary' address lookup components
- Adds appsettings to allow the component to query the OS Places API


HarryAmmon:
Comment from @sussexrick carried over from #598 - "Non-blocking comment as this is working - for other similar tag helpers, there would be a child tag helper for the hint."

HarryAmmon:
Comment from @sussexrick  carried over from #598 - "Non-blocking suggestion - in some other components we have default text coming from a dictionary entry, with the hard-coded default text as the fallback if that is missing."

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/component-builder.js
```diff
@@ -0,0 +1,229 @@
...
+
+    createLinkList(links) {
+        const linkList = document.createElement('ul');
+        linkList.setAttribute(this.config.ATTRIBUTES.BASE, this.config.DATA_ATTRIBUTES.LINK_LIST);
+        linkList.className = this.config.CSS_CLASSES.LINK_LIST;
+        if (links !== undefined && links.length !== 0) {
+            links.forEach(link => {
+                linkList.appendChild(link);
+            });
+        }
```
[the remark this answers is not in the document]
HarryAmmon: @copilot This comment seems incomplete, can you re-word it?

ThePensionsRegulator.Frontend/TagHelpers/TprAddressLookupTagHelper.cs
```diff
@@ -1,8 +1,10 @@
-﻿using GovUk.Frontend.AspNetCore.Extensions;
+﻿using System;
+using System.Threading.Tasks;
+using GovUk.Frontend.AspNetCore.Extensions;
+using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Html;
 using Microsoft.AspNetCore.Mvc.TagHelpers;
 using Microsoft.AspNetCore.Razor.TagHelpers;
```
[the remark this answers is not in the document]
HarryAmmon: @copilot the Razor.TagHelpers version doesn't appear to be available

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/index.js
```diff
@@ -0,0 +1,522 @@
...
+        postcode = this.postcodeSanitiser.sanitise(postcode);
+
+        const searchResults = await this.apiService.searchAddresses(postcode, building);
+        this.JsonResults = searchResults.results;
+
+        if (searchResults.length === 1) {
+            const address = this.addressMapper.mapFromDpaResult(searchResults[0]);
+            this.stateMachine.transition(AddressLookupStateMachine.STATES.CONFIRMED, { address });
+        } else if (searchResults.length > 1) {
+            this.stateMachine.transition(AddressLookupStateMachine.STATES.SELECT, { results: searchResults });
```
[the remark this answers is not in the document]
HarryAmmon: `this.JsonResults` was never used, so it has been removed

HarryAmmon:
Note to self: this is a regression. This should appear as a fieldset error 
image

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprAddressLookup.cs
```diff
@@ -2,18 +2,32 @@
...
+            {
+                container.Attributes.Add("data-address-lookup-same-as-primary-checkbox-label", sameAsPrimaryCheckboxLabel);
+            }
+
+            var fieldSet = new TagBuilder("fieldset");
+            fieldSet.AddCssClass("govuk-fieldset");
+            if (!string.IsNullOrEmpty(fieldsetDescribedBy))
+            {
+                fieldSet.Attributes.Add("described-by", fieldsetDescribedBy);
+            }
```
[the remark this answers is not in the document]
sussexrick: This was not a problem because `described-by` is a tag helper attribute which is consistent with GOV.UK tag helpers, and which is rendered as `aria-describedby`. However it was correctly referencing an id for the hint text, but the hint text did not have that id. That's now fixed.

sussexrick:
> Comment from @sussexrick carried over from #598 - "Non-blocking comment as this is working - for other similar tag helpers, there would be a child tag helper for the hint."

Done

sussexrick:
> Note to self: this is a regression. This should appear as a fieldset error image

Done (by Harry)

sussexrick:
> Comment from @sussexrick carried over from #598 - "Non-blocking suggestion - in some other components we have default text coming from a dictionary entry, with the hard-coded default text as the fallback if that is missing."

Done

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
```diff
@@ -33,52 +34,62 @@
...
+        ViewContext.ModelState
+    );
 }
 
-<tpr-address-lookup class="@cssClasses" aria-describedby="@fieldsetDescribedBy">
-    <tpr-address-lookup-legend is-page-heading="@legendIsPageHeading" class="@legendClass">@legend</tpr-address-lookup-legend>
+<tpr-address-lookup class="@cssClasses" described-by="@fieldsetDescribedBy" same-as-primary-checkbox-label="@checkboxLabel" role="@role" data-address-lookup-search-url="@addressLookupEndpointUrlProvider.GetAddressLookupSearchEndpoint()" data-address-lookup-id-url="@addressLookupEndpointUrlProvider.GetAddressLookupIdEndpoint()">
+    <tpr-address-lookup-legend>@legend</tpr-address-lookup-legend>
 
     @if (hintTextBlocks is not null && hintTextBlocks.Any())
```
sussexrick, line 103: Text in the hint and child blocks only shows with JS disabled.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
```diff
@@ -91,15 +102,15 @@
...
     {
-            <div class="govuk-hint">@hintText</div>
+        <tpr-address-lookup-hint id="@addressLine1ModelProperty-hint">@hintText</tpr-address-lookup-hint>
     }
 
     <govuk-client-side-validation 
         error-message-required="@(Model.Settings.Value<string>(TprPropertyAliases.AddressLine1ErrorMessageRequired))"
         error-message-maxlength="@(Model.Settings.Value<string>(TprPropertyAliases.AddressLine1ErrorMessageMaxLength))">
-        <govuk-input name=@addressLine1State.ModelPropertyName value="@addressLine1State.AttemptedValue" input-aria-invalid="@addressLine1State.InvalidAriaLabel" autocomplete="address-line1">
+        <govuk-input name="@addressLine1State.ModelPropertyName" value="@addressLine1State.AttemptedValue" input-aria-invalid="@addressLine1State.InvalidAriaLabel" autocomplete="address-line1" input-attributes="@(new Dictionary<string, string?>{{ "data-address-lookup", "address-line-1" }})">
```
sussexrick, line 122: We offer the ability to customise the field labels, but JS doesn't respect that.

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/component-builder.js
```diff
@@ -0,0 +1,229 @@
...
+        return link;
+    }
+
+    createLinkList(links) {
+        const linkList = document.createElement('ul');
+        linkList.setAttribute(this.config.ATTRIBUTES.BASE, this.config.DATA_ATTRIBUTES.LINK_LIST);
+        linkList.className = this.config.CSS_CLASSES.LINK_LIST;
+        if (links !== undefined && links.length !== 0) {
+            links.forEach(link => {
+                linkList.appendChild(this.createListItem(link));
```
sussexrick, line 109: When you click a link in these lists it causes the form to validate (possibly by triggering the blur event on a form element, hooked up in validator.js). This is turn causes the link not to be followed. You have to click it a second time to follow the link. One example is when you choose to enter an international address and then immediately choose to go back to postcode lookup,

Also these links should be buttons since they are changing the state of a component rather than navigating to a new location.

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
```diff
@@ -13,73 +12,111 @@
...
+            This should be the URL for the <a class="govuk-link" href="https://docs.os.uk/os-apis/accessing-os-apis/os-places-api/technical-specification/postcode" target="_blank">postcode endpoint</a> of the Ordnance Survey API,
+            or an equivalent service.
+        </p>
+
+        <h3 class="govuk-heading-s">
+            AddressByIdEndpoint
+        </h3>
+
+        <p class="govuk-body">
+            This should be the URL for the <a class="govuk-link" href="https://docs.os.uk/os-apis/accessing-os-apis/os-places-api/technical-specification/uprn" target="_blank">postcode endpoint</a> of the Ordnance Survey API,
```
sussexrick, line 88: The text of this link should refer to the UPRN endpoint

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tpraddresslookup.config
```diff
@@ -53,6 +53,22 @@
...
     </GenericProperty>
+    <GenericProperty>
+      <Key>0b4def25-cb4e-4dcc-b39b-596684eba760</Key>
+      <Name>Address lookup checkbox label</Name>
+      <Alias>addressLookupCheckboxLabel</Alias>
+      <Definition>0cc0eba1-9960-42c9-bf9b-60e150b429ae</Definition>
+      <Type>Umbraco.TextBox</Type>
+      <Mandatory>false</Mandatory>
+      <Validation></Validation>
+      <Description><![CDATA[Label for 'same as' checbox for a seconday address lookup component.]]></Description>
```
sussexrick, line 64: Two typos - checkbox and secondary

ThePensionsRegulator.Frontend.Umbraco/TprPropertyAliases.cs
```diff
@@ -74,6 +77,7 @@ public static class TprPropertyAliases
         public const string TownOrCityErrorMessageRequired = "townOrCityErrorMessageRequired";
         public const string TownOrCityErrorMessageMaxLength = "townOrCityErrorMessageMaxLength";
         public const string TownOrCityLabel = "townOrCityLabel";
+        public const string UPRNModelProperty = "uPRNModelProperty";
```
sussexrick, line 81: This alias could be changed to better casing. And the C# property should be `UprnModelProperty` to follow MS standards.

ThePensionsRegulator.Frontend/TagHelpers/TprAddressLookupTagHelper.cs
```diff
@@ -25,9 +30,30 @@ public TprAddressLookupTagHelper() : this(null)
...
+        [HtmlAttributeName(RoleAttributeName)]
+        public AddressLookupRole Role { get; set; } = AddressLookupRole.Primary;
+
         /// <inheritdoc/>
         public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
         {
+
+            if (Role == AddressLookupRole.Secondary && string.IsNullOrEmpty(SameAsPrimaryCheckboxLabel))
+            {
+                throw new InvalidOperationException(
```
sussexrick, line 53: **non-blocking suggestion**: Could we default to "Same as previous address" instead of throwing an exception?

ThePensionsRegulator.Frontend/ThePensionsRegulator.Frontend.csproj
```diff
@@ -322,4 +322,8 @@
 	<ItemGroup>
 	  <UpToDateCheckInput Remove="styles\_tpr-section-cards.scss" />
 	</ItemGroup>
+	
+	<ItemGroup>
+	  <Folder Include="__mocks__\_content\ThePensionsRegulator.GovUk.Frontend\govuk\" />
```
sussexrick, line 327: This appears to refer to an empty folder

ThePensionsRegulator.Frontend/__tests__/address-lookup/postcode-sanitiser.tests.js
```diff
@@ -0,0 +1,33 @@
...
+
+import { jest } from '@jest/globals';
+import { PostcodeSanitiser } from "../../wwwroot/tpr/address-lookup/postcode-sanitiser.js";
+
+describe("Postcode sanitiser", () => {
+    const sanitiser = new PostcodeSanitiser();
+
+    it.each([
+        ["bn1 4dw", "BN1 4DW"],
+        ["bn14dw", "BN14DW"],
```
sussexrick, line 11: Is the business OK with getting postcodes with or without a space in the middle, so the same postcode could appear in two formats? Perhaps we need normalisation that also addresses that, rather than sanitisation that just removes problems.

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/config.js
```diff
@@ -0,0 +1,113 @@
...
+        X_SMALL: "x-small",
+        SMALL: "small",
+        MEDIUM: "medium",
+        LARGE: "large",
+        X_LARGE: "x-large",
+        XX_LARGE: "xx-large"
+    },
+
+    LABELS: {
+        ADDRESS_CONFIRMED: "Confirmed address",
```
sussexrick, line 55: Currently all of these strings:
a) are not localisable 
b) overwrite any customisation in Umbraco

These can be defaults but the postcode lookup should attempt to replace them when it initialises with values written by the server-side code. These can be values already present on the screen, or data attributes for others.

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/config.js
```diff
@@ -0,0 +1,113 @@
...
+    FIELD_NAMES: {
+        SELECT_ADDRESS: "Find-address"
+    },
+
+    VALIDATION: {
+        POSTCODE_REQUIRED: true,
+        SELECT_REQUIRED: true
+    },
+    PATTERNS: {
+        POSTCODE: "^\\s*[A-Za-z]{1,2}[0-9]{1,2}[A-Za-z]?[\\s\\-]*[0-9][ABDEFGHJLNPQRSTUWXYZabdefghjlnpqrstuwxyz]{2}\\s*$"
```
sussexrick, line 106: Our postcode validation regex should be consistent on client and server. 

They're close, but [the server version allows more punctation](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/ThePensionsRegulator.GovUk.Frontend/Validation/UkPostcodeAttribute.cs), at every position that `ALLOWED_PUNCTUATION` is inserted.

We should have the [same set of tests](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/develop/ThePensionsRegulator.GovUk.Frontend.UnitTests/UkPostcodeAttributeTests.cs) (which might also mean updating the server-side tests with examples from `postcode-validation.tests.js`).

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/api-service.js
```diff
@@ -0,0 +1,60 @@
...
+        if (!results.results) {
+            return [];
+        }
+
+        let filteredResults = results.results.map(result => result.DPA);
+
+        if (building && building !== '') {
+            building = building.trim();
+            if (Number.isInteger(Number(building))) {
+                filteredResults = filteredResults.filter(result => result.BUILDING_NUMBER === building);
```
sussexrick, line 29: Should this be handling suffixes like 12a and ranges like 110-112, like you have when converting the address, rather than just checking for an integer?

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/validator.js
```diff
@@ -0,0 +1,136 @@
...
+        });
+
+        return allValid;
+    }
+
+    validAddress(address) {
+        return !!(address.addressLine1 && address.town && address.postcode);
+    }
+
+    addOrUpdateCustomFieldsetError(fieldset, message) {
```
sussexrick, line 85: Happy to accept this as tech debt for now, but it feels like at least some of this should've been incorporated into `govuk-validation.js` as improved overall validation support, rather than in the code for a specific component. We should log as issues any features that were found to be missing, requiring this code to be written.

docs/components/tpr-address-lookup.md
```diff
@@ -0,0 +1,325 @@
...
+
+```razor
+@addTagHelper *, ThePensionsRegulator.Frontend
+
+<form method="post" novalidate>
+    <tpr-address-lookup
+        data-address-lookup-search-url="/api/address/search"
+        data-address-lookup-id-url="/api/address/byid">
+        <tpr-address-lookup-legend>Enter your address</tpr-address-lookup-legend>
+        <govuk-client-side-validation>
```
sussexrick, line 17: Needs an example of adding a hint

sussexrick, line 17: This shows one instance of `<govuk-client-side-validation>` surrounding all inputs, but in the actual code you have separate instances to support separate error messages.

docs/components/tpr-address-lookup.md
```diff
@@ -0,0 +1,325 @@
...
+                <govuk-input-label>Country</govuk-input-label>
+            </govuk-input>
+            <govuk-input for="Postcode" input-class="govuk-input--width-10" autocomplete="postal-code"
+                         data-address-lookup="postcode">
+                <govuk-input-label>Postcode</govuk-input-label>
+            </govuk-input>
+        </govuk-client-side-validation>
+    </tpr-address-lookup>
+
+    <govuk-button type="submit">Submit</govuk-button>
```
sussexrick, line 45: UPRN and country hidden fields are missing

docs/components/tpr-address-lookup.md
```diff
@@ -0,0 +1,325 @@
...
+        "POST_TOWN": "LONDON",
+        "POSTCODE": "SW1A 1AA",
+        "ORGANISATION_NAME": ""
+      }
+    }
+  ]
+}
+```
+
+If a building name or number is provided, results are filtered client-side: numeric values match `BUILDING_NUMBER` exactly, non-numeric values match `BUILDING_NAME` (case-insensitive substring).
```
sussexrick, line 219: Numeric including 12a and ranges like 110-119?

docs/components/tpr-address-lookup.md
```diff
@@ -0,0 +1,325 @@
...
+| `index.js`              | Entry point. Bootstraps `TprAddressLookup` instances on `DOMContentLoaded`.            |
+| `config.js`             | Central configuration: selectors, labels, error messages, patterns.                    |
+| `state-machine.js`      | `AddressLookupStateMachine` — manages states and valid transitions.                    |
+| `api-service.js`        | `AddressLookupApiService` — fetches addresses from search and ID endpoints.            |
+| `component-builder.js`  | `AddressLookupComponentBuilder` — creates GOV.UK-styled DOM elements.                  |
+| `mapper.js`             | `AddressMapper` — maps between DPA results, form inputs, and internal address objects. |
+| `validator.js`          | `AddressLookupValidator` — integrates with jQuery Validation Unobtrusive.              |
+| `input-builder.js`      | `InputBuilder` — fluent API for adding validation attributes to inputs.                |
+| `postcode-sanitiser.js` | `PostcodeSanitiser` — trims, uppercases, and removes hyphens from postcodes.           |
+| `utils.js`              | Utility functions (e.g. `toSentenceCase`).                                             |
```
sussexrick, line 310: `toSentenceCase` is now `toTitleCase` - I missed this one

ThePensionsRegulator.Frontend/Styles/_tpr-address-lookup.scss
```diff
@@ -0,0 +1,8 @@
+@import '_tpr-variables.scss';
+
+.tpr-address-lookup__links {
+    a {
+        cursor: pointer;
```
sussexrick, line 5: This doesn't seem to achieve anything because:

- `cursor: pointer` is the default for a elements anyway
- `color: $govuk-link-colour` will always be overridden by the more specific selector `govuk-link:link` (or other states)

Perhaps you are trying to override the visited colour, but if so then a better behaviour might be to make the link destination `#some-id-unique-to-this-lookup-instance-and-link` so that the visited state represents whether the user has visited that state of the component or not.

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tpraddresslookupsettings.config
```diff
@@ -19,7 +19,6 @@
       <Composition Key="bf69e8d6-97a6-4b50-bec7-c5993e120339">govukCssClasses</Composition>
       <Composition Key="d508995b-a958-4273-8ab3-50735220a1b2">govukGrid</Composition>
       <Composition Key="d7016563-0b6f-4c3a-b75d-b0bbe66419fd">govukGridColumnClasses</Composition>
-      <Composition Key="8ba759bf-131b-488e-b7e6-98ff91bb076b">govukLegendIsPageHeading</Composition>
```
sussexrick, line 22: With 'legend as page heading' removed, you are saying we only support the correct HTML for having multiple instances on a page, and do not support the correct HTML for a single thing per page.

This addresses your immediate need but it is tech debt that needs to be addressed later, so it should be logged as an issue.

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
```diff
@@ -100,6 +100,14 @@
 }</code></pre>
         </govuk-inset-text>
 
+        <govuk-warning-text icon-fallback-text="Warning">
+            Consuming services must implement ITprCountryRepository to allow this components full functionality.
```
sussexrick, line 104: Typo "component's"

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/Content/address-lookup.config
```diff
@@ -126,7 +138,7 @@
...
+        "markup": "<p>If JavaScript is not available then the component will display input fields for entering an international address. The labels for these fields can be configured in Umbraco.</p>\n<p></p>",
```
sussexrick, line 141: Empty para can be removed here.

docs/components/tpr-address-lookup.md
```diff
@@ -258,6 +258,26 @@ When using the Umbraco integration (`ThePensionsRegulator.Frontend.Umbraco`), co
 
 These values are read by `TprAddressLookupEndpointUrlProvider`, which implements `ITprAddressLookupEndpointUrlProvider`. You can register a custom implementation if you need to resolve endpoints differently.
 
+### Country repository
+
+You must register an implementation of `ITprCountryRepository` in the dependency injection container. The view injects this service and serialises the result of `GetCountries()` into the `data-address-lookup-countries` attribute on the component element. The JavaScript reads this attribute to populate the country dropdown in the manual international address entry view.
```
sussexrick, line 263: Could we register a default implementation? Something so that the page doesn't crash the moment you add your first 'address lookup' block to it. For example it could read a list of countries from appsettings.json, and return an empty array if missing.

GovUk.Frontend.ExampleApp/Views/AddressLookup/Index.cshtml
```diff
@@ -100,6 +100,14 @@
...
+        </govuk-warning-text>
+
+        <p class="gouvk-body">
```
sussexrick, line 107: Typo in the class name

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
```diff
@@ -1,12 +1,15 @@
-﻿@model AddressLookupViewModel
+﻿@using System.Text.Json
+@model AddressLookupViewModel
 @inject ITprAddressLookupEndpointUrlProvider urlProvider
+@inject ITprCountryRepository countryRepository
```
sussexrick, line 5: This heading is on every demo page in the ASP.NET app, but I think it's unnecessary. 'Shipping address' and 'Billing address' do the job.

Address h2

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
```diff
@@ -79,15 +82,18 @@
     </govuk-client-side-validation>
 
     <input type="hidden" id="ShippingUPRN" name="ShippingUPRN" value="@Model?.ShippingUPRN" data-address-lookup="UPRN" />
+    <input type="hidden" id="ShippingCountryCode" name="ShippingCountryCode" value="@Model?.ShippingCountryCode" data-address-lookup="country-code" />
 </tpr-address-lookup>
 
 <h3 class="govuk-heading-m">Billing address</h3>
```
sussexrick, line 88: I think your HTML structure is wrong. Looking at search view you have an actual fieldset (shown in red below) around building name/number and postcode, and a legend of 'Search by postcode'. I think this legend is unnecessary. 

The actual set of fields that together form a single 'thing' also includes the potential 'same as above' checkbox and the secondary button. The legend is actually 'Shipping address' or 'Billing address'.

If I tab to this component NVDA announces 'Search by postcode' but it doesn't tell me it's a shipping address, which is crucial information. A 'Postcode' field and 'Find address' button tell me I can search by postcode.

Example component in search view with fieldsets marked

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
```diff
@@ -79,15 +82,18 @@
...
 <p class="govuk-body">Tell us where we should bill your order to.</p>
 
 <tpr-address-lookup role="Secondary"
```
sussexrick, line 91: This field label is way too big:

Select an address, looking like h1

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
```diff
@@ -79,15 +82,18 @@
...
                     same-as-primary-checkbox-label="Same as shipping address"
```
sussexrick, line 92: Using NVDA when I enter a postcode only and press 'Find address' it takes me to the 'select address' view and announces "choose an address collapsed required invalid entry". Why invalid? It doesn't look invalid. It's because the blank entry is selected - if I select another and go back to blank it says "invalid entry" again. If you update the `<option>` to have display text like 'Select an address' but still no value you have the same behaviour for your validators but avoid the "invalid entry" announcement.

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
```diff
@@ -79,15 +82,18 @@
...
-                    data-address-lookup-search-url=@urlProvider.GetAddressLookupSearchEndpoint()
-                    data-address-lookup-id-url=@urlProvider.GetAddressLookupIdEndpoint()>
+                    data-address-lookup-search-url="@urlProvider.GetAddressLookupSearchEndpoint()"
```
sussexrick, line 93: When I enter a postcode only and press 'Find address' it takes me to the 'select address' view. If I press Alt+⬅️ to go back, it takes me back to the previous page. I would expect to go back to the previous state of this component. You can use the JS history API to make each state part of the back button history.

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
```diff
@@ -79,15 +82,18 @@
...
+                    data-address-lookup-id-url="@urlProvider.GetAddressLookupIdEndpoint()"
```
sussexrick, line 94: On the 'select address' view if I select the top blank option and click 'Confirm address' there is no audible feedback in NVDA. I think it should move focus back to the first invalid field.

Similar for manual UK address entry.

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
```diff
@@ -79,15 +82,18 @@
...
+                    data-address-lookup-countries="@JsonSerializer.Serialize(countryRepository.GetCountries())">
```
sussexrick, line 95: I would expect the postcode box to be shorter here (in the ASP.NET app):

Postcode same width as town/county

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
```diff
@@ -79,15 +82,18 @@
...
+
```
sussexrick, line 96: If I enter a valid building number and postcode I get to here:

Address displayed as text

Based on the 'Edit address' label I'd expect to see this address loaded into UK address text inputs so that I could edit it. I understand from your docs that this is not intended. However you use the text 'Return to postcode search' in other contexts to signal more clearly the intended behaviour.

Also this text leads me to expect that the address I've already selected would be pre-populated in the next view, even if that is the postcode search view. 'Edit' does not suggest 'start again'.

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/validator.js
```diff
@@ -0,0 +1,136 @@
...
+
+            fieldset.parentNode.insertBefore(formGroup, fieldset);
+            formGroup.appendChild(fieldset);
+        }
+
+        let errorElement = fieldset.querySelector(".govuk-error-message");
+        if (!errorElement) {
+            const errorParaSpan = document.createElement("span");
+            errorParaSpan.classList.add("govuk-visually-hidden");
+            errorParaSpan.innerText = "Error: ";
```
sussexrick, line 100: This text should be localisable.

On checking this I discovered we already have this issue (and a solution) in `govuk-validation.js` and I've just added #765 as a tech debt item to fix it. You could add your use case to that ticket, but should not just ignore it.

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/config.js
```diff
@@ -0,0 +1,115 @@
...
+    },
+
+    ERROR_MESSAGES: {
+        SELECT_REQUIRED: "Select an address",
+        MAX_LENGTH_500: "Your answer must not exceed 500 characters",
+        MAX_LENGTH_100: "Must be no more than 100 characters",
+        MAX_LENGTH_20: "Your answer must not exceed 20 characters",
+        INVALID_COUNTRY: "This is not a valid country",
+        COUNTRY_NOT_UK: "This page is for international addresses",
+        REQUIRED: "This field must be populated",
```
sussexrick, line 91: Given these messages are hard-coded (and even if they became defaults that could be overridden) they should follow GOV.UK guidance. 

> if the error relates to a specific field within the question, ... refer to that field in the error message - for example: “you must enter a year”

Here's the reason:

Error summary with many identical messages

Also the text used for a required validator is always "Enter <field label>", not "This field must be populated". Check with our content designers, but I think the word "valid" is also discouraged.

https://design-system.service.gov.uk/components/error-message/

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
```diff
@@ -0,0 +1,172 @@
...
+<tpr-address-lookup role="Secondary"
+                    same-as-primary-checkbox-label="Same as shipping address"
+                    data-address-lookup-search-url="@urlProvider.GetAddressLookupSearchEndpoint()"
+                    data-address-lookup-id-url="@urlProvider.GetAddressLookupIdEndpoint()"
+                    data-address-lookup-countries="@JsonSerializer.Serialize(countryRepository.GetCountries())">
+    
+    <tpr-address-lookup-legend>Billing address</tpr-address-lookup-legend>
+
+    <govuk-client-side-validation error-message-required="This field is required"
```
sussexrick, line 99: If I click 'Same as shipping address' when the secondary address lookup is in the search view, I would expect the fields to be hidden.

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
```diff
@@ -0,0 +1,172 @@
...
+                                  error-message-maxlength="Cannot exceed the maximum length">
+        <govuk-input for="BillingAddressLine1" autocomplete="address-line1"
```
sussexrick, line 101: If I press Enter in the secondary address it triggers the button (and any errors) in the primary address. This would also be true if other fields were on the same page, when Enter should fire the primary action button for the whole form.

The correct solution is to have the buttons within this component use `<button type="button">` and trap the enter key using JS _within this component only_.

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
sussexrick: This example shows more clearly that the heading is redundant:

Heading and legend both saying 'Shipping address'

GovUk.Frontend.ExampleApp/Views/AddressLookup/JavaScriptDisabled.cshtml
```diff
@@ -0,0 +1,28 @@
...
+        <p class="govuk-body">
+            If JavaScript is disabled, the address component shows text inputs that would accept an international address.
+            If data is provided, the relevant fields are populated with this data. When a user saves an address, the
+            configured API is not queried.
+        </p>
+
+        <h2 class="govuk-heading-l">Address</h2>
+
+        <form asp-controller="AddressLookup" asp-action="PostJavaScriptDisabled" method="post" novalidate>
+            <partial name="_AddressLookupForm" model="Model" />
```
sussexrick, line 25: These field widths should be appropriate to the content:

All the same width

GovUk.Frontend.ExampleApp/Views/AddressLookup/JavaScriptDisabled.cshtml
```diff
@@ -0,0 +1,28 @@
...
+        </form>
```
sussexrick, line 26: Country with JS is a dropdown:

Country as dropdown

Country with JS disabled is free text:

Country as text input

Since the countries are available server-side this should also be a dropdown. And if the dropdown is rendered server-side that's also a better place to read the data into JS, rather than duplicating it in a data attribute.

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/component-builder.js
```diff
@@ -0,0 +1,234 @@
...
+        link.setAttribute(this.config.ATTRIBUTES.BASE, dataAddressLookupValue);
+        link.classList = this.config.CSS_CLASSES.LINK;
+        link.innerText = linkText;
+        link.href = "#";
+
+        return link;
+    }
+
+    createLinkList(links) {
+        const linkList = document.createElement('ul');
```
sussexrick, line 104: There's nearly always just one item so this isn't really a list. `<nav>` would be more appropriate, and `<nav>` always needs an 'aria-label` attribute (which should be localisable) explaining what it's the navigation for.

sussexrick (changes requested):

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/api-service.js
HarryAmmon: This is now supported

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
HarryAmmon: I don't agree that this should be the behaviour. The component contains navigation links that allow the user to transition to other states, including back to the search state. When there are 2 components on the screen (as they are in the service that implements this component) how would I know what component the user wanted to go back on?

I will run this comment past our UI/UX person, maybe your comment highlights that the links on the component are not enough and could be improved.

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
HarryAmmon: Recorded as #787

ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml
HarryAmmon: All input field labels can now be customised. This is done by either respecting the field labels that come from the view, or if there isn't a corresponding element rendered by the non JS view, a `data-address-lookup-*` attribute.

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/component-builder.js
HarryAmmon: Captured by #788

ThePensionsRegulator.Frontend/TagHelpers/TprAddressLookupTagHelper.cs
HarryAmmon: Captured as #789

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/config.js
HarryAmmon: `config.js` has been restructured so these type of strings are now defaults. `field-defaults.js` now quieries the original inputs, `data-address-lookup-{data-attribute}-{suffix}` and then `config.js` (in that order) to determine the label or error message.

This has not been done for every label, piece of text or error message yet, only the ones raised by the test team as an issue in the service currently implementing this component.

Captured by #791

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/config.js
HarryAmmon: Captured by #792

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/validator.js
HarryAmmon: Captured as it's own issue #793 (not as a sub-issue of the address lookup issue) as improved validation

ThePensionsRegulator.Frontend/Styles/_tpr-address-lookup.scss
HarryAmmon: Not sure what I was trying to achieve, removed

GovUk.Frontend.Umbraco.ExampleApp/uSync/v9/ContentTypes/tpraddresslookupsettings.config
HarryAmmon: Captured as #795

docs/components/tpr-address-lookup.md
HarryAmmon: Captured as #796

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
HarryAmmon: Captured as #797

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
HarryAmmon: Captured in #798

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
HarryAmmon: Captured in #799

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/config.js
HarryAmmon: The defaults have been updated to closer align with GOV.UK guidance, likely not perfect. Error messages should now be customisble, either by respecting the data-val-* attribute on an exsiting element, or by using `data-address-lookup-{ELEMENT}-{SUFFIX}`. Umbraco properties exist for these as well. 

I have raised that these messages need reviewing with our content editor.

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
HarryAmmon: I don't know what you mean by this. From how I've interpretted your comment, this is how it works.
image
image

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
HarryAmmon: Captured by #800

GovUk.Frontend.ExampleApp/Views/AddressLookup/JavaScriptDisabled.cshtml
HarryAmmon: Widths updated to match [GOV.UK example](https://design-system.service.gov.uk/patterns/addresses/)

image

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/component-builder.js
HarryAmmon: Captured by #801

ThePensionsRegulator.Frontend/__tests__/address-lookup/postcode-sanitiser.tests.js
HarryAmmon: Lifted and shifted the code that did this from the original system

GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
HarryAmmon: This has been captured as something to look into once our service goes live

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/validator.js
HarryAmmon: Updated #765 with this use case.

HarryAmmon:
This PR is being abandoned and `v9.3.0` will never get a full release. This is because the address lookup component and the work in this branch is being migrated over to `v10.*` which consuming servies are moving to.

PR #755, opened 2026-04-08
#60926 Invalid mobile menu structure with nested interactive elements

Max-Emery:
Why:

The mobile menu toggle contained invalid HTML structure, with a nested button element inside an anchor tag and incorrect use of `aria-live` attribute. This can lead to inconsistent behaviour across browsers and assistive technologies.

[AB#260926](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/260926)

In this PR:

- Removed `aria-live` attribute from anchor tag. 
- Changed container element to `div` element.
- Ensured both icon and text are part of the button element.
- Provided new fallback link for when JS is disabled.


ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
```diff
@@ -359,22 +363,22 @@ function restoreDefaultMenuState(e) {
...
+    const toggle = document.querySelector(".tpr-header-menu__button");
+    toggle.classList.toggle("tpr-header-menu__button--open");
 
-    const svg = document.querySelector(".tpr-mobile-menu__svg");
-    svg?.classList.toggle("tpr-mobile-menu__svg--hide");
+    const svgs = document.querySelectorAll(".tpr-mobile-menu__svg");
+    svgs.forEach(s => s.classList.toggle("tpr-mobile-menu__svg--hide"));
 
-    const button = document.querySelector(".tpr-header-menu__button");
+    const button = document.querySelector(".tpr-header-menu__button-inner");
```
sussexrick, line 372: I noticed that there can be an `aria-expanded` attribute on `-tpr-header-menu__button-inner` that never changes its value from `false`. This is just a comment which wouldn't block approval, because I couldn't find where in the code it happens, but perhaps you can because you know it better. I think it's a real edge case as the attribute only appears at larger screen sizes when the mobile menu is not normally rendered.

Image

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderMenu.cs
```diff
@@ -7,19 +7,66 @@ partial class ComponentGenerator
...
 
             if (tprMobileMenu.Attributes != null)
             {
                 mobileMenuToggle.MergeAttributes(mobileMenuToggle.Attributes);
             }
 
+            var closeButton = new TagBuilder("span");
+            closeButton.AddCssClass("tpr-header-menu__button-inner");
+
+            if (!string.IsNullOrEmpty(tprHeaderBar.HeaderMenuToggleOpen))
```
sussexrick: It's possible on the tag helper not to set these attributes or to set them to an empty string. I think they need to be required (throw an exception if missing) or have a default value if not set (which could be hard-coded in English). As it stands this currently renders the icon with no text, which is announced as "button expanded" and "button collapsed" without any indication of what the button does.

sussexrick (changes requested):

ThePensionsRegulator.Frontend/wwwroot/tpr/headermenu.js
Max-Emery: This was some left over logic I missed and was applied to wrong element, have moved to `.tpr-header-menu__button` so when screen is resized ARIA expanded value reflects menu state.

ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderMenu.cs
Max-Emery: Have updated these properties, along with ARIA label properties to provide a default value if empty.

Max-Emery:
I've also included a fix for an issue I spotted when reviewing another task, as its only a small update I've included it here.

When mobile menu child menu is focused, the associated arrow button updates to its focus colour, despite not being in focus anymore - this presents a contrast issue when the menu item is highlighted as the current section.

contrast issue

When no longer focused arrow button will return to default styling.

sussexrick (approved):

Issue #761, opened 2026-04-16
Example apps should require trusted types for script


sussexrick:
Our newer applications enforce the use of `require-trusted-types-for 'script'` in the content security policy, so both example apps should also enable this.

This will ensure that components are tested in an environment with realistic security restrictions.

PR #762, opened 2026-04-16
Enable trusted types and permissions policy in example apps (v13)

sussexrick:
Recent apps that we build have more a more restrictive security environment than the example apps. Specifically they enable permissions policy and trusted types for scripts.

This PR updates the .NET 8 version of example apps to implement both of those directives using code copied from our actual applications.

#761


AlexBeesley (commented):
Same comments as I left on the v17 version of this PR. Happy to approve once you have seen those.

AlexBeesley (dismissed):

AlexBeesley (approved):

PR #764, opened 2026-04-17
Enable trusted types and permissions policy in example apps (v17)

sussexrick:
Recent apps that we build have more a more restrictive security environment than the example apps. Specifically they enable permissions policy and trusted types for scripts.

This PR updates the .NET 10 version of example apps to implement both of those directives using code copied from our actual applications.

#761


GovUk.Frontend.ExampleApp/Middleware/SecurityHeadersMiddleware.cs
```diff
@@ -14,16 +14,25 @@ public SecurityHeadersMiddleware(RequestDelegate next, IWebHostEnvironment webHo
 
         public async Task InvokeAsync(HttpContext context, INonceProvider nonceProvider)
         {
-            var connectSrcForHotReload = _webHostEnvironment.IsDevelopment() ? "'self' ws://localhost:* http://localhost:64300" : string.Empty;
-            var styleSrcForAblePlayer = "'sha384-xBuQ/xzmlsLoJpyjoggmTEz8OWUFM0/RC5BsqQBDX2v5cMvDHcMakNTNrHIW2I5f' 'sha384-ETDm/j6COkRSUfVFsGNM5WYE4WjyRgfDhy4Pf4Fsc8eNw/eYEMqYZWuxTzMX6FBa'";
+            var connectSrcForVisualStudioBrowserLink = _webHostEnvironment.IsDevelopment() ? "'self' ws://localhost:* http://localhost:*" : string.Empty;
```
AlexBeesley: **nitpick** A better name should describe what the directive actually permits (dev-only localhost HTTP + WS), not a specific tooling feature. Something like `connectSrcForDevLocalhost`.

GovUk.Frontend.Umbraco.ExampleApp/wwwroot/js/trusted-types-policy.js
```diff
@@ -0,0 +1,14 @@
+﻿if (typeof(trustedTypes) != undefined && trustedTypes.createPolicy && typeof (DOMPurify) !== 'undefined') {
```
AlexBeesley: **question** I think `typeof trustedTypes` returns a string (e.g. "undefined"), so comparing to undefined will always be true? 

Maybe this would be better here? 
`if (window.trustedTypes && trustedTypes.createPolicy && typeof DOMPurify !== 'undefined') { ... }`

AlexBeesley (commented):
A few comments, but looks good.

AlexBeesley (commented):
deleted

GovUk.Frontend.ExampleApp/Middleware/SecurityHeadersMiddleware.cs
sussexrick: done

GovUk.Frontend.Umbraco.ExampleApp/wwwroot/js/trusted-types-policy.js
sussexrick: Good spot, updated

AlexBeesley (approved):

PR #769, opened 2026-04-21
#60926 Invalid mobile menu structure with nested interactive elements (v17)

Max-Emery:
Why:

The mobile menu toggle contained invalid HTML structure, with a nested button element inside an anchor tag and incorrect use of aria-live attribute. This can lead to inconsistent behaviour across browsers and assistive technologies.

[#AB260926](https://dev.azure.com/thepensionsregulator/TPR/_boards/board/t/Web%20Platform%20Team/Stories?System.Tags=Corp%20Web%20Accessibility%2CCorp%20Website%20Migration&System.AssignedTo=Max.Emery%40tpr.gov.uk&workitem=260926)

In this PR:

- Removed aria-live attribute from anchor tag.
- Changed container element to div element.
- Ensured both icon and text are part of the button element.
- Provided new fallback link for when JS is disabled.


sussexrick:
@Max-Emery Please check this has all recent changes from your v13 version. I can see the default text for buttons is missing for example.

Max-Emery:
Hi @sussexrick, default button value should now be showing. Also ensured that arrow receives correct focus style when clicked. PR is up to date, I'll also raise a PR to make these changes available for v13

sussexrick (changes requested):
When I run the Umbraco example app, uSync import everything, and access the TPR header demo page I get the following error:

The collection already contains an entry with data-close-label

Max-Emery:
whoops, was applying the same attribute twice through the previous null check. Removed redundant code and made the same update for v13.

sussexrick (approved):

PR #783, opened 2026-04-24
Bugfix/button not showing in backoffice

ZebSadiq:
Button StyleOfButton undefined fix
Vulnerability fix for OpenMcdf


GovUk.Frontend.ExampleApp/GovUk.Frontend.ExampleApp.csproj
```diff
@@ -8,6 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation" Version="10.0.1" />
+    <PackageReference Include="OpenMcdf" Version="3.1.3" />
```
sussexrick: This was already fixed in #779 with fewer changes, and including an upgrade to the package that depends on it since upgrading only this is a potential breaking change. You should remove your fix and just merge the latest from develop.

sussexrick (changes requested):

GovUk.Frontend.ExampleApp/GovUk.Frontend.ExampleApp.csproj
ZebSadiq: Ok done.

sussexrick (approved):

PR #785, opened 2026-04-24
Support royal blue and green boxes

### LINKED ISSUE #746: TPR box needs to support a darker blue
Identify the colour used here in the TPR colour palette and add it as an option on the 'TPR box' block. Needed for v17 only.

Box with darker blue background

sussexrick:
TPR designs are using royal blue and green boxes to reflect statuses. We need to support these colour options for the 'TPR box' and 'TPR nested box' components.

Green, royal blue and grey boxes indicating statuses

## Changes in this PR

- updates the 'TPR Box colour' data type in Umbraco with the new options
- adds new CSS classes, and updates views to map the selection in Umbraco to the expanded range of classes
- inverts the colour of content within the new boxes to maintain contrast
- adds examples to both example apps, including an example of all supported content within a box
- fixes the display of tables in the backoffice, which had regressed from showing TPR styles to showing GOV.UK styles

There would be several ways to invert the colour of the content inside these boxes in the final rendered output. However, when used in the Umbraco backoffice the elements that need to be inverted are blocks that are descendants of the box block. Each block has its own shadow root, so we need styles in a descendant shadow root to adapt to a class applied in an ancestor shadow root. CSS variables are the only method that can cross this boundary and work in the backoffice as well as on the final rendered component. 

Each element that needs to update has its colour(s) set to the variables defined here, with a fallback to the default colour for that element. These variables are normally undefined, so the fallback default is used. When a selector for the new box colours defines the variables, the variable value is used instead of the fallback and the colours are inverted.

Due to the complexity of inverting all styles for all components I have created examples of content within boxes and called that the supported components within a coloured box. We may need to add more later, but the range I've added support for should cover all common scenarios.

## Examples of box colours

Example showing the range of box colours

Content colours within a grey box

Content colours within a royal blue box

Content colours within a green box

## Examples in Umbraco backoffice

Content colours within a grey box in Umbraco

Content colours within a royal blue box in Umbraco

Content colours within a green box in Umbraco

Closes #746.

### DIFF
--- a/GovUk.Frontend.ExampleApp/Views/DividerLinesAndBoxes/Index.cshtml
+++ b/GovUk.Frontend.ExampleApp/Views/DividerLinesAndBoxes/Index.cshtml
@@ -71,6 +71,9 @@
         <div class="tpr-box tpr-box--royal-blue">
             <p class="govuk-body">This box has its background colour set to royal blue.</p>
         </div>
+        <div class="tpr-box tpr-box--green">
+            <p class="govuk-body">This box has its background colour set to green.</p>
+        </div>
     </div>
 </div>
 <div class="govuk-grid-row govuk-grid-row--tpr-divider tpr-box">
@@ -87,7 +90,12 @@
         </div>
         <div class="tpr-box tpr-box--royal-blue">
             <partial name="_BoxContentPart1" />
-            <p class="govuk-body">This box is set to Royal Blue. Inverse colours must be used when the box has a dark background.</p>
+            <p class="govuk-body">This box is set to royal blue. Inverse colours must be used when the box has a dark background.</p>
+            <partial name="_BoxContentPart2" />
+        </div>
+        <div class="tpr-box tpr-box--green">
+            <partial name="_BoxContentPart1" />
+            <p class="govuk-body">This box is set to green. Inverse colours must be used when the box has a dark background.</p>
             <partial name="_BoxContentPart2" />
         </div>
     </div>
--- a/GovUk.Frontend.ExampleApp/Views/DividerLinesAndBoxes/Index.cshtml
+++ b/GovUk.Frontend.ExampleApp/Views/DividerLinesAndBoxes/Index.cshtml
@@ -60,10 +60,16 @@
         <div class="tpr-box">
             <p class="govuk-body">Boxes can be added within columns.</p>
         </div>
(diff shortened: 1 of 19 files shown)


JodiB-TPR (changes requested):
I can't approve this in its current form. It's proposing to support the use of these two colours to represent **statuses**. I don't think these colours are appropriate for these components for a few reasons.

When talking about colour variables, the GOV.UK Design System guidance clearly says “Only use the variables in the context they’re designed for.” ([Colour – GOV.UK Design System](https://design-system.service.gov.uk/styles/colour/#main-colours)). So, the colours being proposed conflict with those intended semantics.

First, the green being used is GOV.UK's green, not TPR's (Camarone — #1e7e34). But more importantly, that colour is currently **reserved for 'Success' and 'Confirmation' messages** (e.g., 'Notification banners' and 'Confirmation panels').

Second, the Royal Blue is **reserved for links and buttons**. Using it for non-interactive boxes would create a 'false affordance' — people reasonably associate blue with something clickable. These boxes are informational, not interactive, so supporting a link/button colour here increases confusion risk.

Third, as mentioned, **these are statuses**, not calls to action. Statuses like “Incomplete” and “Cannot start yet” already have established styling in GOV.UK Task List patterns, which we emulate. Supporting these two colours on these components would make it easier for this pattern to be reused elsewhere, even though the underlying semantics are wrong.

Simply put: **a status should look like a status**, not a control.

johnhayn-es:
Hello, a quick summary from me: 
- These cards/boxes were based on panels (https://design-system.service.gov.uk/components/panel/), and arguably being used in the right context (our user testing backs this up, no one has thought these are buttons, so far)
- I agree with @JodiB-TPR onthe colours but don't know what the paradigm is at TPR - happy to update to the right ones if there's an obvious suggestion/correction
- I think @JodiB-TPR's quote about colour use is taken out of context and doesn't represent how we choose colours, but instead is there to guide developers about when to apply a functional colour vs. directly linking to the web palette. The quote is `Only use the variables in the context they’re designed for. In all other cases, you should reference the [GOV.UK web palette](https://design-system.service.gov.uk/styles/colour/#govuk-web-palette) directly. For example, if you wanted to use red, you should use govuk-colour("red") rather than govuk-functional-colour("error").`

PR #815, opened 2026-05-27
260864 timeline heading hierarchy

georgeboydratcliff:
Updated the TPR Timeline for accessibility, to have <h3> headings instead of <p> headings


villarin:
Pr raised against incorrect branch. A follow up pr has been merged now for both v13 and v17

PR #817, opened 2026-05-29
260864 tpr timeline h3 accessibility

georgeboydratcliff:
Updated TPR Timeline to have H3 headings instead of P headings for accessibility (change copied over from v13)


villarin:
Pr raised against incorrect branch. A follow up pr has been merged now for both v13 and v17

PR #827, opened 2026-06-03
Provide IPublishedValueFallback rather than relying on Umbraco's StaticServiceProvider

sussexrick:
When you use an overload of the `.Value<T>()` method in Umbraco that skips passing `IPublishedValueFallback`, Umbraco gets it instead from its `StaticServiceProvider` instance.

This is a problem in testing because `UmbracoTestContext` sets up the static service provider, but multiple tests or test suites might do this in parallel and, since it is static, enter a race condition to get it set up correctly. We get flaky tests.

We have a Roslyn analyzer for that now, and I've added that to each project that references Umbraco. This is _not_ added as an extra dependency to the packages we publish.

This PR ensures that we always pass `IPublishedValueFallback` to `.Value<T>()` in our own code, and allow consumers of our packaging to do so as well.

I've also improved the test coverage for `DependentFieldsActionFilter` so that it tests support within checkboxes as well as radio buttons.

This PR is non-breaking ~~, save for an edge case that we can be confident is just theoretical - if any consumers were to implement  `IOverridablePublishedElement` instead of using our `IOverridablePublishedElement` they would have to implement a new overload~~.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <Version>10.0.1</Version>
+    <Version>10.0.4</Version>
     <InformationalVersion>$(Version)</InformationalVersion>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
--- a/GovUk.Frontend.Umbraco.ExampleApp/Controllers/FilterAndOverrideBlocksController.cs
+++ b/GovUk.Frontend.Umbraco.ExampleApp/Controllers/FilterAndOverrideBlocksController.cs
@@ -32,8 +32,8 @@ namespace GovUk.Frontend.Umbraco.ExampleApp.Controllers
             };
 
             // Filter out a block in the block list and block grid
-            viewModel.Page.BlockList!.Filter = block => block.Settings?.Value<string>(nameof(GovukGrid.CssClassesForRow)) != "filter-this";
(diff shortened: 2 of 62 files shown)


ThePensionsRegulator.Umbraco.Core/IOverridablePublishedElement.cs
```diff
@@ -9,5 +9,6 @@ public interface IOverridablePublishedElement : IPublishedElement
     {
         void OverrideValue(string alias, object value);
         T? Value<T>(string alias, string? culture = null, string? segment = null, Fallback fallback = default, T? defaultValue = default);
+        T? Value<T>(IPublishedValueFallback publishedValueFallback, string alias, string? culture = null, string? segment = null, Fallback fallback = default, T? defaultValue = default);
```
HarryAmmon, line 12: Given using the original method causes problems, can it be marked as obselute, like has been done for the HeaderLockup?

HarryAmmon: Ah, if I'd scrolled down a tiny bit more I would have seen you have a solution for this. Though I still think it could be made obsolute and removed in the next breaking change

sussexrick: The PR no longer modifies this file as #833 did a more thorough job.

HarryAmmon (approved):

Issue #814, opened 2026-05-26
Mock of .Value needs to support override that specifies IPublishedValueFallback


sussexrick:
When writing unit tests for overriding blocks the mock that is setup by our testing library supports the first line here but not the second:

```csharp
block.Settings.Value<string>("someProperty")); // supported
block.Settings.Value<string>(_testContext.PublishedValueFallback.Object, "someProperty")); // not supported
```

The mock needs to be updated to support both overloads, especially since we are encouraging the use of the second one to avoid flaky tests that rely on Umbraco's static service provider.

PR #833, opened 2026-06-09
Support more variants of .Value() extension method (v17) #814

sussexrick:
We now prefer to call the overload of `.Value()` that takes `IPublishedValueFallback` because it avoids Umbraco's `StaticServiceProvider` and therefore avoids race conditions in unit testing.

#814 identified a gap in our mocking support for that overload, which this PR fixes for the v17 (`develop`) branch.

However, analysis of this and other flaky test results demonstrated more problems with this support. Some scenarios for value types (eg setting up a property as `int` and reading back the value as `int?`) were not supported.

This appeared to be covered by existing tests, but subtle C# behaviours meant it wasn't. 

The helper in the test class has a type constraint which means its generic parameter (was `T`, now `TContentType`) always resolves to `IPublishedElement` even when `IOverridablePublishedElement` is passed. This meant the tests were calling the built-in Umbraco method rather than our replacement for it that supports overriding values.

Testing `.Value<T>` and `.Value<T?>` together also turns out to be futile, since C# resolves both to the same type. One is tested twice and the other isn't tested. We now have extra helper methods to ensure both are supported and tested.

The test suite for `UmbracoPublishedElementExtensions` is now much more comprehensive. It tests all combinations of writing and reading nullable and non-nullable variants, for both `IPublishedContent` (properties set up on a page) and `IOverridablePublishedElement` (properties set up on a block). `UmbracoPublishedElementExtensions` itself has fixes for the bugs and omissions identified by these improved tests.

### DIFF
--- a/ThePensionsRegulator.Umbraco.Core/IOverridablePublishedElement.cs
+++ b/ThePensionsRegulator.Umbraco.Core/IOverridablePublishedElement.cs
@@ -9,5 +9,6 @@ namespace ThePensionsRegulator.Umbraco.Core
     {
         void OverrideValue(string alias, object value);
         T? Value<T>(string alias, string? culture = null, string? segment = null, Fallback fallback = default, T? defaultValue = default);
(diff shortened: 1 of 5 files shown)


HarryAmmon (approved):
#LGTM

PR #834, opened 2026-06-10
Pass IPublishedValueFallback to OverridableBlockListModel and OverridableBlockGridModel (v13)

sussexrick:
Calling `.Value()` without specifying the `IPublishedValueFallback` leads to Umbraco using its `StaticServiceProvider` to get it. In unit testing this can lead to a race condition between tests, so we recommend to use the overload that specifies `IPublishedValueFallback`.

This PR refactors our existing code targeting v13 to use that, mostly in `OverridableBlockListModel` and `OverridableBlockGridModel` where `.Value()` is called in the base class.

[AB#269110](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/269110)

### DIFF
--- a/GovUk.Frontend.Umbraco/Blocks/IOverridablePublishedElementExtensions.cs
+++ b/GovUk.Frontend.Umbraco/Blocks/IOverridablePublishedElementExtensions.cs
@@ -219,7 +219,6 @@ namespace GovUk.Frontend.Umbraco.Blocks
         /// <param name="publishedSnapshotAccessor">Accessor for a published snapshot, which is a point-in-time capture of the current state of everything that is "published".</param>
         /// <param name="publishedValueFallback">The published value fallback provider.</param>
         /// <exception cref="ArgumentNullException">Thrown if any argument is <c>null</c>.</exception>
-        [Obsolete("Use an overload which specifies an IPublishedValueFallback.")]
         public static void OverrideSelectOptions(this IOverridablePublishedElement blockContent,
             IEnumerable<SelectOption> items,
             IPublishedSnapshotAccessor publishedSnapshotAccessor,
--- a/GovUk.Frontend.Umbraco/Blocks/IOverridablePublishedElementExtensions.cs
+++ b/GovUk.Frontend.Umbraco/Blocks/IOverridablePublishedElementExtensions.cs
@@ -1,10 +1,9 @@
 ﻿using GovUk.Frontend.Umbraco.Models;
-using System;
-using System.Collections.Generic;
-using System.Linq;
+using Microsoft.Extensions.DependencyInjection;
 using ThePensionsRegulator.Umbraco;
 using ThePensionsRegulator.Umbraco.Blocks;
 using Umbraco.Cms.Core;
+using Umbraco.Cms.Core.DependencyInjection;
 using Umbraco.Cms.Core.Models.Blocks;
 using Umbraco.Cms.Core.Models.PublishedContent;
 using Umbraco.Cms.Core.PublishedCache;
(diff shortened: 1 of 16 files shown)


GovUk.Frontend.Umbraco/Blocks/IOverridablePublishedElementExtensions.cs
```diff
@@ -145,22 +205,56 @@ public static void OverrideRadioButtons(this IOverridablePublishedElement blockC
...
+        /// <summary>
+        /// Replaces the select options configured in Umbraco with those supplied as an argument.
+        /// </summary>
+        /// <param name="blockContent">The content of a block list item based on the GOV.UK Select component.</param>
+        /// <param name="items">The select options.</param>
+        /// <param name="publishedSnapshotAccessor">Accessor for a published snapshot, which is a point-in-time capture of the current state of everything that is "published".</param>
+        /// <param name="publishedValueFallback">The published value fallback provider.</param>
+        /// <exception cref="ArgumentNullException">Thrown if any argument is <c>null</c>.</exception>
+        [Obsolete("Use an overload which specifies an IPublishedValueFallback.")]
+        public static void OverrideSelectOptions(this IOverridablePublishedElement blockContent,
```
HarryAmmon, line 222: This method does specifify an IPublishedValueFallback

ThePensionsRegulator.Umbraco.Testing/UmbracoBlockGridFactory.cs
```diff
@@ -54,11 +62,17 @@ public static OverridableBlockGridModel CreateOverridableBlockGridModel(BlockGri
...
+        public static OverridableBlockGridModel CreateOverridableBlockGridModel(IPublishedValueFallback publishedValueFallback, IEnumerable<BlockGridItem> blockGridItems)
         {
-            return new OverridableBlockGridModel(blockGridItems, null, OverridableBlockGridItem.NoopPublishedElementFactory);
+            return new OverridableBlockGridModel(publishedValueFallback, blockGridItems, null, OverridableBlockGridItem.NoopPublishedElementFactory);
         }
 
+        /// <summary>
+        /// Create an <see cref="OverridableBlockGridModel"/> containing multiple instances of <see cref="BlockGridItem"/> which can have their property values overridden at runtime.
+        /// </summary>
+        public static OverridableBlockGridModel CreateOverridableBlockGridModel(IEnumerable<BlockGridItem> blockGridItems)
```
HarryAmmon, line 73: Obsolete?

ThePensionsRegulator.Umbraco.Testing/UmbracoBlockListFactory.cs
```diff
@@ -30,19 +30,31 @@ public static BlockListModel CreateBlockListModel(IEnumerable<BlockListItem> blo
...
+        public static OverridableBlockListModel CreateOverridableBlockListModel(IPublishedValueFallback publishedValueFallback, IEnumerable<BlockListItem> blockListItems)
         {
-            return new OverridableBlockListModel(blockListItems, null, OverridableBlockListItem.NoopPublishedElementFactory);
+            return new OverridableBlockListModel(publishedValueFallback, blockListItems, null, OverridableBlockListItem.NoopPublishedElementFactory);
         }
 
+        /// <summary>
+        /// Create an <see cref="OverridableBlockListModel"/> containing multiple instances of <see cref="BlockListItem"/> which can have their property values overridden at runtime.
+        /// </summary>
+        public static OverridableBlockListModel CreateOverridableBlockListModel(IEnumerable<BlockListItem> blockListItems)
```
HarryAmmon, line 55: Obsolete?

ThePensionsRegulator.Umbraco.Testing/UmbracoBlockListFactory.cs
```diff
@@ -30,19 +30,31 @@ public static BlockListModel CreateBlockListModel(IEnumerable<BlockListItem> blo
...
+        public static OverridableBlockListModel CreateOverridableBlockListModel(IPublishedValueFallback publishedValueFallback, BlockListItem blockListItem)
         {
-            return CreateOverridableBlockListModel(new[] { blockListItem });
+            return CreateOverridableBlockListModel(publishedValueFallback, new[] { blockListItem });
         }
 
+        /// <summary>
+        /// Create an <see cref="OverridableBlockListModel"/> containing a single <see cref="BlockListItem"/> which can have its property values overridden at runtime.
+        /// </summary>
+        public static OverridableBlockListModel CreateOverridableBlockListModel(BlockListItem blockListItem)
```
HarryAmmon, line 41: Obsolete?

HarryAmmon (changes requested):

ThePensionsRegulator.Umbraco.Testing/UmbracoBlockGridFactory.cs
sussexrick: No, I don't think so. Because this is test code we can fall back to providing a mock, rather than falling back to the static service provider, so it shouldn't cause flaky tests. 

If it causes a failure it should be a consistent one, which will be because you were customising and testing the behaviour of the `IPublishedValueFallback` instance and then not passing in that instance to be used.

ThePensionsRegulator.Umbraco.Testing/UmbracoBlockListFactory.cs
sussexrick: Same comment as above.

ThePensionsRegulator.Umbraco.Testing/UmbracoBlockListFactory.cs
sussexrick: Same comment as above.

GovUk.Frontend.Umbraco/Blocks/IOverridablePublishedElementExtensions.cs
sussexrick: Well spotted. I'll credit over-zealous AI autocomplete with that one. Fixed.

HarryAmmon (approved):

PR #848, opened 2026-07-01
Permissions issue on download table data as a CSV.

AlexBeesley:
## Issue found in testing on SIT: 

image

## Root cause

When the Umbraco HTML table component renders a table with `EnableTableCsvDownload = true`, it emits a server-side `<form class="tpr-table-download-form">` that POSTs the table HTML to `/api/table/download-csv`. The `tpr-table-csv-download.js` script detected this form and skipped adding its own client-side button, leaving the server form as the sole download mechanism.

On affected pages, Umbraco's routing pipeline was intercepting the POST to `/api/table/download-csv` before the `TableDownloadController` MVC route could handle it, resulting in Umbraco serving its access-denied content page. Tables on other pages worked correctly because they had no server-side form and the script added a client-side button instead.

## Fix

Changed `tpr-table-csv-download.js` so that when a server-side `tpr-table-download-form` is detected, the script intercepts its `submit` event, prevents the default POST to the server, and performs the CSV conversion and file download entirely in the browser — consistent with how client-side buttons work for all other tables.

### Changes

- **`tpr-table-csv-download.js`**
  - Added exported helper `findServerSideDownloadForm(table)` — returns the server-side form element following a table, or `null`.
  - Updated `initTableCsvDownload()` to attach a `submit` listener on any detected server-side form. The listener calls `e.preventDefault()` then runs `downloadCsv(tableToCsv(table), fileName)`. A `data-tpr-table-csv-intercepted` attribute prevents duplicate listeners if the function is called more than once (idempotency).

- **`tpr-table-csv-download.tests.js`**
  - Added import for `findServerSideDownloadForm`.
  - Added `describe("findServerSideDownloadForm")` test suite covering direct sibling, nested sibling, additional classes, missing button, and unrelated form cases.
  - Added `describe("initTableCsvDownload — server-side form interception")` test suite verifying: submit triggers client-side download, no extra button is added, and the listener is not duplicated on repeated `init` calls.

## Testing

- Existing tests all continue to pass.
- Manually verified on the affected page: clicking "Download table data (CSV)" now downloads the file client-side without navigating to the API URL.

## Related

[AB #269788](https://dev.azure.com/thepensionsregulator/TPR/_workitems/edit/269110)

### DIFF
--- a/ThePensionsRegulator.Frontend.Umbraco/ServiceCollectionExtensions.cs
+++ b/ThePensionsRegulator.Frontend.Umbraco/ServiceCollectionExtensions.cs
@@ -82,7 +82,16 @@ namespace ThePensionsRegulator.Frontend.Umbraco
             services.AddTprGovUkFrontendUmbraco(configureGovUkOptions, configureGovUkUmbracoOptions);
 
             // ThePensionsRegulator.Frontend
-            services.AddTprFrontend(configureGovUkOptions, configureTprOptions);
+            // Keep these registrations explicit here rather than calling AddTprFrontend(), because
+            // AddTprGovUkFrontendUmbraco() above has already registered the GOV.UK frontend services.
+            services.AddTransient<IContextAwareHostUpdater, TprHostUpdater>();
+            services.AddTransient<IConsentCookieReader, TprConsentCookieReader>();
+            services.AddTransient<IStaticFileCachePolicy, TprStaticFileCachePolicy>();
+            services.AddTransient<ITableCsvService, TableCsvService>();
+
+            var tprFrontendOptions = new TprFrontendOptions();
+            configureTprOptions(tprFrontendOptions);
+            services.AddTransient((services) => Options.Create(tprFrontendOptions));
 
             // ThePensionsRegulator.Frontend.Umbraco
             services.Configure<RazorViewEngineOptions>(options => options.ViewLocationFormats.Add("/Views/Shared/TPR/{0}.cshtml"));
--- a/ThePensionsRegulator.Frontend.Umbraco/ServiceCollectionExtensions.cs
+++ b/ThePensionsRegulator.Frontend.Umbraco/ServiceCollectionExtensions.cs
@@ -82,13 +82,7 @@ namespace ThePensionsRegulator.Frontend.Umbraco
             services.AddTprGovUkFrontendUmbraco(configureGovUkOptions, configureGovUkUmbracoOptions);
 
             // ThePensionsRegulator.Frontend
-            services.AddTransient<IConsentCookieReader, TprConsentCookieReader>();
-            services.AddTransient<IContextAwareHostUpdater, TprHostUpdater>();
-            services.AddTransient<IStaticFileCachePolicy, TprStaticFileCachePolicy>();
-
-            var tprFrontendOptions = new TprFrontendOptions();
-            if (configureTprOptions is not null) { configureTprOptions(tprFrontendOptions); }
-            services.AddTransient((services) => Options.Create(tprFrontendOptions));
+            services.AddTprFrontend(configureGovUkOptions, configureTprOptions);
 
             // ThePensionsRegulator.Frontend.Umbraco
             services.Configure<RazorViewEngineOptions>(options => options.ViewLocationFormats.Add("/Views/Shared/TPR/{0}.cshtml"));
--- a/ThePensionsRegulator.Frontend/Controllers/TableDownloadController.cs
+++ b/ThePensionsRegulator.Frontend/Controllers/TableDownloadController.cs
@@ -2,7 +2,7 @@ using Microsoft.AspNetCore.Mvc;
 using System.Text.RegularExpressions;
 using ThePensionsRegulator.Frontend.Services;
 
-namespace ThePensionsRegulator.Frontend.Umbraco.Controllers;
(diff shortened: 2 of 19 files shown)


villarin (commented):
Please check [moderate vulnerability](https://dev.azure.com/thepensionsregulator/TPR/_build/results?buildId=276494&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=300e099e-8e44-55d6-a32c-adee27218fa0&l=42) and npm audit fix

ThePensionsRegulator.Frontend.Umbraco/ServiceCollectionExtensions.cs
```diff
@@ -82,13 +82,7 @@ public static IServiceCollection AddTprFrontendUmbraco(
...
 
             // ThePensionsRegulator.Frontend
-            services.AddTransient<IConsentCookieReader, TprConsentCookieReader>();
-            services.AddTransient<IContextAwareHostUpdater, TprHostUpdater>();
-            services.AddTransient<IStaticFileCachePolicy, TprStaticFileCachePolicy>();
-
-            var tprFrontendOptions = new TprFrontendOptions();
-            if (configureTprOptions is not null) { configureTprOptions(tprFrontendOptions); }
-            services.AddTransient((services) => Options.Create(tprFrontendOptions));
+            services.AddTprFrontend(configureGovUkOptions, configureTprOptions);
```
sussexrick: These lines were repeated for a reason - if you call `services.AppTprFrontend()` that in turn calls `services.AddGovUkTprFrontend()`, but that has already been called by `services.AddTprGovUkFrontendUmbraco()`.

ThePensionsRegulator.Frontend/Controllers/TableDownloadController.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Text.RegularExpressions;
+using ThePensionsRegulator.Frontend.Services;
+
+namespace ThePensionsRegulator.Frontend.Umbraco.Controllers;
+
+/// <summary>
+/// Handles CSV download requests for HTML tables.
+/// Provides both server-side fallback for no-JS scenarios and client-side form interception.
+/// </summary>
+[ApiController]
```
sussexrick, line 13: Why is this in the Umbraco project? The reference to it is in `ThePensionsRegulator.Frontend`, but that would be a broken link if you tried to use it without Umbraco.

I think it makes sense to move this to `ThePensionsRegulator.Frontend` alongside the code it supports. Then the only gap is how to add the form to make no-JS work if you're not using Umbraco. That needs to be solved. At it's simplest it could just be a documentation update.

ThePensionsRegulator.Frontend/Services/TableHtmlCsvFormGenerator.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            var tableInput = doc.CreateElement("input");
+            tableInput.SetAttributeValue("type", "hidden");
+            tableInput.SetAttributeValue("name", "tableHtml");
+            tableInput.SetAttributeValue("value", HttpUtility.HtmlEncode(table.OuterHtml));
+            form.AppendChild(tableInput);
+
+            var fileNameInput = doc.CreateElement("input");
+            fileNameInput.SetAttributeValue("type", "hidden");
+            fileNameInput.SetAttributeValue("name", "fileName");
```
sussexrick, line 55: I was concerned that untrusted client data was being posted for the filename and HTML so I asked CoPilot to review. I thought I was only asking but it generated suggested fixes too, so !855 is available for you to review and merge into your branch if you agree with the changes. They look like improvements to me.

ThePensionsRegulator.Frontend/Services/TableHtmlCsvFormGenerator.cs
```diff
@@ -0,0 +1,75 @@
...
+            tokenInput.SetAttributeValue("type", "hidden");
+            tokenInput.SetAttributeValue("name", "__RequestVerificationToken");
+            tokenInput.SetAttributeValue("value", antiForgeryToken);
+            form.AppendChild(tokenInput);
+
+            var submitButton = doc.CreateElement("button");
+            submitButton.SetAttributeValue("type", "submit");
+            submitButton.SetAttributeValue("class", "govuk-button govuk-button--secondary");
+            submitButton.SetAttributeValue("data-module", "govuk-button");
+            submitButton.InnerHtml = "Download table data (CSV)";
```
sussexrick: Should support localisation to Welsh.

sussexrick (dismissed):
This is better but there's still some confusion over whether it's a feature that requires Umbraco or not.

The server-side solution works well. So well that I'm not sure how much value the client-side solution now adds. (The client-side solution also works well.)

ThePensionsRegulator.Frontend/Services/TableHtmlCsvFormGenerator.cs
villarin: merged now

ThePensionsRegulator.Frontend/Controllers/TableDownloadController.cs
villarin: Looks good now. @AlexBeesley  please bump release version

villarin (approved):

PR #849, opened 2026-07-01
Upgrade umbraco to v17.5.0

villarin:
This PR includes:

- Upgrade umbraco to v17.5.0 to deal with vulnerable dependencies
- Upgrade auxiliary  .net core dependencies
- Fix breaking changes for controller based components using null IPublishedValueFallback dependency
- Fixed mock setup extensions to cater for umbraco 17.5.0 updates affecting PublishedContentExtensions

### DIFF
--- a/ThePensionsRegulator.Umbraco.Testing/UmbracoPublishedContentExtensions.cs
+++ b/ThePensionsRegulator.Umbraco.Testing/UmbracoPublishedContentExtensions.cs
@@ -28,6 +28,11 @@ namespace ThePensionsRegulator.Umbraco.Testing
                     It.IsAny<string?>()))
                 .Returns(parent is null ? [] : [parent]);
 
+            filteringService
+                .Setup(x => x.Unfiltered(
+                    It.Is<IEnumerable<Guid>>(keys => parent == null ? !keys.Any() : keys.Count() == 1 && keys.First() == parentKey)))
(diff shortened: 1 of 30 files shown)


ThePensionsRegulator.Umbraco.Testing/UmbracoPublishedContentExtensions.cs
```diff
@@ -28,6 +28,11 @@ private static Mock<IPublishedContent> SetupParent(this Mock<IPublishedContent>
                     It.IsAny<string?>()))
                 .Returns(parent is null ? [] : [parent]);
 
+            filteringService
```
sussexrick, line 31: It's good to support extra functionality but you should add tests for it.

sussexrick (changes requested):

ThePensionsRegulator.Umbraco.Testing/UmbracoPublishedContentExtensions.cs
villarin: We use these extensions to provide setup mock inputs for node ancestors/descendants within other unit tests like the sidenav, etc. So although no specific unit tests are provided for this class they are used across multiple test cases throughout the unit tests projects. The update here ensure we are up to date with changes released for Umbraco's PublishedContentExtensions as part of release 17.5.0

sussexrick (approved):

PR #850, opened 2026-07-02
Add Address Line 3 support to address lookup feature

tpr-martyn-hills:
Extended address lookup to include Address Line 3 across UI, validation, data mapping, and configuration.

Updated Razor views, Umbraco models, and config files.

Enhanced address parsing logic and expanded unit tests.

Adjusted JSON test data and improved code consistency.

[AB#270325](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/270325)


HarryAmmon (changes requested):
I don't think we should show the user the address line 3 input, it isn't part of the design. We can add it at a later date but for now this should be a hidden input.

ThePensionsRegulator.Frontend/wwwroot/tpr/address-lookup/mapper.js
```diff
@@ -125,27 +199,23 @@
         return null;
     }
 
-
-    mapFromManualEntry(addressLine1, addressLine2, town, countyOrRegion, postcode, country, countryCode) {
+    mapFromManualEntry(addressLine1, addressLine2, addressLine3, town, countyOrRegion, postcode, country, countryCode) {
```
HarryAmmon: addressLine3 isn't part of manual entry, so this param isn't needed. Also doesn't need to return an object with addressLine3 as a property

HarryAmmon (changes requested):

HarryAmmon (approved):

PR #853, opened 2026-07-03
Add support for building a breadcrumb trail using async code (v13)

sussexrick:
This PR updates the `GovUkBreadcrumbComponent` view component for Umbraco v13 to accept either async or synchronous implementations of the service that provides the breadcrumb links. Only the synchronous service was supported before.

Adding the extra option enables async code to be called within the service without inefficient use of `.Result`. The async implementation is preferred but will fall back gracefully to the synchronous implementation if not provided.

This is a non-breaking change because the change to the constructor will be handled by DI.

A similar PR will follow for v17 so that this does not become a regression.

[AB#271232](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/271232)


AlexBeesley:
@copilot Any suggestions? None for me.

AlexBeesley (approved):
Looks good :)

PR #854, opened 2026-07-06
Add support for building a breadcrumb trail using async code (v17)

sussexrick:
This PR updates the `GovUkBreadcrumbs` partial view for Umbraco v17 to accept either async or synchronous implementations of the service that provides the breadcrumb links. Only the synchronous service was supported before.

Adding the extra option enables async code to be called within the service without inefficient use of `.Result`. The async implementation is preferred but will fall back gracefully to the synchronous implementation if not provided.

This is a non-breaking change because the async service is optional and the view will fall back to the previous behaviour.

[AB#271232](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/271232)

### DIFF
--- a/ThePensionsRegulator.GovUk.Frontend.Umbraco/Views/Shared/GOVUK/GovUkBreadcrumbs.cshtml
+++ b/ThePensionsRegulator.GovUk.Frontend.Umbraco/Views/Shared/GOVUK/GovUkBreadcrumbs.cshtml
@@ -1,12 +1,24 @@
-﻿@using ThePensionsRegulator.GovUk.Frontend.Umbraco.Services
-@using ThePensionsRegulator.Umbraco
+﻿@using Microsoft.Extensions.DependencyInjection
+@using ThePensionsRegulator.GovUk.Frontend.Umbraco.Models
+@using ThePensionsRegulator.GovUk.Frontend.Umbraco.Services
 @using ThePensionsRegulator.Umbraco.Core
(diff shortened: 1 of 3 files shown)


ThePensionsRegulator.GovUk.Frontend.Umbraco/Views/Shared/GOVUK/GovUkBreadcrumbs.cshtml
```diff
@@ -1,12 +1,24 @@
...
+@using ThePensionsRegulator.GovUk.Frontend.Umbraco.Services
 @using ThePensionsRegulator.Umbraco.Core
-@using Umbraco.Cms.Core.Routing
 @addTagHelper *, GovUk.Frontend.AspNetCore
-@inject IGovUkBreadcrumbLinksService breadcrumbLinksService
+@inject IServiceProvider serviceProvider
 @inject IUmbracoPublishedContentAccessor publishedContext
 @{
-    var model = breadcrumbLinksService.GetLinks(publishedContext.PublishedContent);
+    BreadcrumbViewModel? model = null;
```
AlexBeesley, line 9: *nitpick* - Maybe add a defensive null check here?

AlexBeesley:
Aside from my nitpick, looks fine to me unless @copilot has any suggestions,

AlexBeesley (approved):
All good 👍🏻

PR #861, opened 2026-07-16
Side nav collapse functionality using escape key - adjustments (v17)

### LINKED ISSUE #730: Sidenav collapse functionality using Escape key - outstanding adjustments
As discussed the sidenav collapse functionality is now implemented using the escape key to cater for the accessibility audit observations.

A couple of issues remain from the original PR - which are logged here for follow up on release 24

- Color contrast issue identified by @Max-Emery  - When the menu closes, focus doesn’t return anywhere meaningful. If the user had focus inside the side nav, they lose context completely. Could we restore focus to the element that opened the menu?
- Optimise side nav js as per @sussexrick feedback - Instead of repeating collapse logic - Try to select all the chevrons indicating an open menu and trigger a click() event.

Max-Emery:
Why:

During the PR review process, two outstanding issues were identified within the side navigation component.

In this PR:

- Refactored collapseAll() to remove duplicated collapse logic and reuse the existing functionality.
- Improved onEscapeKey behaviour:
  - Ensured focus remains visible when closing menus with Esc, providing a clearer and more accessible keyboard navigation experience.

Closes #730  

`Question:`  I've removed the event listener from document, leaving only the listener on the side navigation. My thinking is that this will prevent the menu from unexpectedly closing when users interact elsewhere on the page. It also means we no longer need to guard against scenarios where the side nav might be closed while focus is within editable elements.

However, if keeping the document-level listener is preferred, I can add the relevant guards back into onEscapeKey.

### DIFF
--- a/ThePensionsRegulator.Frontend.Umbraco/Backoffice/package-lock.json
+++ b/ThePensionsRegulator.Frontend.Umbraco/Backoffice/package-lock.json
@@ -2542,9 +2542,9 @@
       }
     },
     "node_modules/dompurify": {
-      "version": "3.4.11",
-      "resolved": "https://registry.npmjs.org/dompurify/-/dompurify-3.4.11.tgz",
-      "integrity": "sha512-zhlUV12GsaRzMsf9q5M254YhA4+VuF0fG+QFqu6aYpoGlKtz+w8//jBcGVYBgQkR5GHjUomejY84AV+/uPbWdw==",
+      "version": "3.4.12",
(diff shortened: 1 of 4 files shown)


villarin (dismissed):

villarin:
@Max-Emery new npm vulnerabilities identified. Please npm audit fix or upgrade dompurify dependency:

dompurify  <=3.4.11
DOMPurify: `CUSTOM_ELEMENT_HANDLING` bypasses `afterSanitizeElements` for allowed custom elements

villarin (approved):

PR #864, opened 2026-07-22
#232483 --govuk-frontend-version is set to "development" on consuming projects

Ethan-Lambourne:
Within the packaged CSS provided by this solution there is a variable loaded onto each page called `--govuk-frontend-version` which is set to "development". This is very misleading for any projects that consume this solution.

In this PR:
- Generated a `_tpr-version.scss` file with the pack version during dotnet pack so the CSS custom property `--govuk-frontend-version` is set to the real package version in published packages;
- Updated `Directory.Build.targets` to create the version partial (`_tpr-version.scss`) for packaging and then remove it after pack so that the value for `--govuk-frontend-version` remains as "development" within the actual dev environment;
- Added a unit test to verify the partial containing a version is generated, the `govuk-frontend.scss` is updated to import it, and then that the restore target cleans up the generated file afterwards (all of which is done in a temp folder to avoid altering files in the repo).

This is related to the work item: [AB#232483](https://dev.azure.com/thepensionsregulator/TPR/_workitems/edit/232483)


sussexrick (changes requested):
The idea is sound, but I think you can do this much more simply, without needing to copy the whole project folder or have a cleanup step.

You can set an MSBuild variable like this, which has a different value for local builds (typically Debug configuration) and pipeline builds (typically Release configuration).

```xml
<PropertyGroup>
   <MyVariable>default value</MyVariable>
</PropertyGroup>

<PropertyGroup Condition="'$(Configuration)' == 'Release'">
    <MyVariable>pipeline value</MyVariable>
</PropertyGroup>
```

If you look at `ThePensionsRegulator.Frontend\Directory.Build.targets` you can find a target `TprFrontend_GenerateSassVersion` that generates a SASS version file, so you can do similar to that. Use a filename that matches *.generated.* and make sure it is not checked into git as it's generated on every build. 

Because you generate a file in either scenario you can just include it in the standard *.scss file, without having to back it up and generate a new one. See `ThePensionsRegulator.Frontend\Styles\_open-sans.scss`.

Ethan-Lambourne:
@sussexrick I've actioned your comment in commit https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/pull/864/commits/8536b2601e3ca0aedc3579f344455c3b2455580f

PR #210, opened 2023-09-21
Set CacheLevel to Snapshot for Blocklist/RTE PropertyValueConverters

ancientyouthman:
Blocklist values were being cached until cleared / application restarted  in consuming apps (Pensions Dashboard), resulting in incorrect data being displayed in views.

Setting the PropertyCache level for these converters to Snapshot resolves this issue.

~(Also fixed a minor spelling mistake 🙂)~

PR #866, opened 2026-07-30
Property value converters were caching content across multiple requests

### LINKED ISSUE #865: Umbraco content is cached too aggressively in Umbraco 17
In v17 it looks like the page content is shared across requests rather than generated per request. This means that any modified content on a page is persisted across requests and causes an issue if we are looking for a placeholder to update. 

For example, if a first request for a page returns content that includes a token for replacement, a subsequent request might return the content with that token already replaced.

This issue was seen in v13, described in #210 and fixed in #211 by setting the `PropertyCacheLevel` to `Snapshot`.

In Umbraco v17 the concept of a snapshot is gone, and during the upgrade [this setting was replaced with `Element`](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/commit/0b17a8a6aa73c21e0d85fbef7cdd6a29b42ce626#diff-c89a7914b819940ae25873b61d4febadefcc6e276864b20e43b68e299c9171a0) but this was a best guess at the appropriate value, and it was [removed entirely from RichTextEditorPropertyValueConverter](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/commit/e5a727c02500464172d329c00e579175f9f36009#diff-dd77eb57b8f1e7653931ae1855da991c5c1568ee17290f62e9a66185e72bc53e) . Overriding of property values where the value differs per request (eg different users) is not working.

This can be reproduced by adding `{{token}}` to the value of a 'Page Heading' block (for example) and then putting this code in the controller:

```csharp
var headingBlock = model.Blocks.FindBlockByContentTypeAlias(ElementTypeAliases.PageHeading);
var heading = headingBlock.Content.Value<string>(PropertyAliases.PageHeading);
headingBlock.Content.OverrideValue(PropertyAliases.PageHeading, heading.ToString().Replace("{{token}}", DateTime.Now.ToLongTimeString()));
```

The time string will not update with each request because the token replaced with the initial value is preserved. It should update, because each request should see the token rather than the previously replaced value.

sussexrick:
In v17 it looks like the page content is shared across requests rather than generated per request. This means that any modified content on a page is persisted across requests and causes an issue if we are looking for a placeholder to update.

For example, if a first request for a page returns content that includes a token for replacement, a subsequent request might return the content with that token already replaced.

This issue was seen in v13, described in https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/pull/210 and fixed in https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/pull/211 by setting the PropertyCacheLevel to Snapshot.

In Umbraco v17 the concept of a snapshot is gone, and during the upgrade [this setting was replaced with Element](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/commit/0b17a8a6aa73c21e0d85fbef7cdd6a29b42ce626#diff-c89a7914b819940ae25873b61d4febadefcc6e276864b20e43b68e299c9171a0) but this was a best guess at the appropriate value, and it was [removed entirely from RichTextEditorPropertyValueConverter](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/commit/e5a727c02500464172d329c00e579175f9f36009#diff-dd77eb57b8f1e7653931ae1855da991c5c1568ee17290f62e9a66185e72bc53e) . Having reproduced the faulty behaviour, we now know the correct value is None.

This PR also adds some missing tests for property value converters.

Fixes #865.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <Version>10.1.2</Version>
+    <Version>10.1.3</Version>
     <InformationalVersion>$(Version)</InformationalVersion>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
--- a/ThePensionsRegulator.Umbraco.Core.Tests/Blocks/OverridableBlockGridPropertyValueConverterTests.cs
+++ b/ThePensionsRegulator.Umbraco.Core.Tests/Blocks/OverridableBlockGridPropertyValueConverterTests.cs
@@ -1,0 +1,74 @@
+﻿using Microsoft.Extensions.Options;
+using Moq;
+using ThePensionsRegulator.Umbraco.Core.Blocks;
+using ThePensionsRegulator.Umbraco.Core.PropertyEditors;
+using ThePensionsRegulator.Umbraco.Testing;
+using Umbraco.Cms.Core.Configuration.Models;
+using Umbraco.Cms.Core.Models.PublishedContent;
+using Umbraco.Cms.Core.PropertyEditors;
+using Umbraco.Cms.Core.PropertyEditors.ValueConverters;
+
+namespace ThePensionsRegulator.Umbraco.Core.Tests.Blocks
+{
+    public class OverridableBlockGridPropertyValueConverterTests
+    {
+        [Fact]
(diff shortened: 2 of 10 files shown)


sussexrick:
Doesn't fully fix the issue. Will start again with another PR.

PR #872, opened 2026-08-10
274812 change to return error message in front end when 29th February entered in day and month date controls

dcr670:
[AB#274812](https://dev.azure.com/thepensionsregulator/0158f35d-cea5-4d9d-adc8-9fc3d3d98793/_workitems/edit/274812) change to return error message in front end when 29th February entered in day and month date controls.

### DIFF
--- a/ThePensionsRegulator.GovUk.Frontend.Umbraco/ModelBinding/UmbracoDateInputModelBinder.cs
+++ b/ThePensionsRegulator.GovUk.Frontend.Umbraco/ModelBinding/UmbracoDateInputModelBinder.cs
@@ -271,6 +271,10 @@ namespace ThePensionsRegulator.GovUk.Frontend.Umbraco.ModelBinding
                 {
                     errors |= DateInputParseErrors.InvalidDay;
                 }
+                else if (parsedYear == 1900 && parsedMonth == 2 && parsedDay == 29)
+                {
+                    errors |= DateInputParseErrors.InvalidDay;
+                }
             }
 
             dateParts = errors == DateInputParseErrors.None ? new(
--- a/ThePensionsRegulator.GovUk.Frontend.Umbraco/ModelBinding/UmbracoDateInputModelBinder.cs
+++ b/ThePensionsRegulator.GovUk.Frontend.Umbraco/ModelBinding/UmbracoDateInputModelBinder.cs
@@ -271,7 +271,7 @@ namespace ThePensionsRegulator.GovUk.Frontend.Umbraco.ModelBinding
                 {
                     errors |= DateInputParseErrors.InvalidDay;
                 }
-                else if (parsedYear == 1900 && parsedMonth == 2 && parsedDay == 29)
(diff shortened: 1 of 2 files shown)


ThePensionsRegulator.GovUk.Frontend.Umbraco/ModelBinding/UmbracoDateInputModelBinder.cs
```diff
@@ -271,6 +271,10 @@ internal static DateInputParseErrors Parse(DateInputItemTypes itemTypes, string?
                 {
                     errors |= DateInputParseErrors.InvalidDay;
                 }
+                else if (parsedYear == 1900 && parsedMonth == 2 && parsedDay == 29)
```
sussexrick: While this addresses the specific test scenario logged, does it address the full problem? What if someone put 30 February, or 31 June? 

I think the test needs to be for a day number that is invalid for the month, rather than any specific example. You should be able to use the .NET DateTime code to check that.

dcr670: I think this is dealt by the code above.  parsedDay > maxDaysInMonth should trigger an error.

ThePensionsRegulator.GovUk.Frontend.Umbraco/ModelBinding/UmbracoDateInputModelBinder.cs
```diff
@@ -271,6 +271,10 @@ internal static DateInputParseErrors Parse(DateInputItemTypes itemTypes, string?
...
+                {
+                    errors |= DateInputParseErrors.InvalidDay;
+                }
```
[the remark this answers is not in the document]
dcr670: Changed to use expectYear.

ThePensionsRegulator.GovUk.Frontend.Umbraco/ModelBinding/UmbracoDateInputModelBinder.cs
sussexrick: Looking at the wider code, there is still a problem. We have this comment:

```
// If we know the year and month we can figure out the days in the month.
// If we only have the month and not a year, we should assume the year could be a leap year.
```

We use 2000 to calculate `maxDaysInMonth` _because_ it's a leap year. This was clearly deliberate because it allows 29 Feb. The day and month are then applied to the year 1900 which is _not_ a leap year, and that's why we have a bug.

While a scheme is unlikely to have 29 Feb as its year end date, this is a generic date entry field and it shouldn't disallow a valid date. When's your birthday? 29 Feb is a valid answer. If we want to disallow 29 Feb as a year end date I think that should be additional validation on that data item, not built into the generic date entry field.

Looking back at the bug logged, the problem isn't that 29 Feb is being allowed, it's that a 500 error is being thrown, and an invalid date is being saved. If we used a leap year for the default year in 'day and month' mode instead of 1900 it would fix the 500 error and save a valid date.

PR #883, opened 2026-08-19
Duplicate form id in global search

Max-Emery:
Why:

The header search component is generating invalid HTML because multiple form elements share the same id attribute value. IDs must be unique within a page.

In this PR:

- Add support to control whether an ID is assigned to form element, preventing duplicate IDs while minimising the impact on existing consuming projects.

### DIFF
--- a/ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
+++ b/ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
@@ -140,7 +140,7 @@ namespace ThePensionsRegulator.Frontend.HtmlGeneration
             {
                 form.Attributes.Add("action", tprHeaderBar.ActionPath);
             }
-            form.Attributes.Add("id", "tpr-header-search__form");
+
             form.Attributes.Add("method", "get");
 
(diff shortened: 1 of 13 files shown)


ThePensionsRegulator.Frontend/HtmlGeneration/ComponentGenerator.TprHeaderBar.cs
```diff
@@ -140,7 +140,12 @@ public virtual TagBuilder GenerateTprHeaderSearch(TprHeaderBar tprHeaderBar)
             {
                 form.Attributes.Add("action", tprHeaderBar.ActionPath);
             }
-            form.Attributes.Add("id", "tpr-header-search__form");
+
+            if (includeId)
```
villarin: Although this condition will sort out the duplicated element ids. If someone decides to use it they will end up with the same duplicate id.
Maybe we can expose the id as an attribute - so if the tag helper needs to be called twice we can specify the id instead?

Max-Emery: Actioned. Form ID can now be set programmatically on both the main search and second search within the mobile nav through new tag helper attributes.

villarin (approved):

PR #886, opened 2026-08-21
added no script address line 3 input

BrandonMallett-TPR:
Added two additional fields for address look up so that address line 3 could be edited when javascript is disabled in the browser.

### DIFF
--- a/GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
+++ b/GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
@@ -54,6 +54,20 @@
         </govuk-input>
     </govuk-client-side-validation>
 
+    <noscript>
+        <govuk-client-side-validation error-message-maxlength="Address line 3 must be 100 characters or less">
+            <govuk-input for="ShippingAddress.AddressLine3" autocomplete="address-line3"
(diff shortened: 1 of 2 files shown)


GovUk.Frontend.ExampleApp/Views/AddressLookup/_AddressLookupForm.cshtml
HarryAmmon, line 1: This change just updates the example application, which we do want to do. To make this change available in the Umbraco application we need to change the [partial view for the block rendered by Umbraco](https://github.com/thepensionsregulator/govuk-frontend-aspnetcore-extensions/blob/feature/address-lookup-v13-pt2/ThePensionsRegulator.Frontend.Umbraco/Views/Shared/TPR/TPRAddressLookup.cshtml).

Should these be wrapped by `noscript` tags? These inputs are shown when JS is disabled anyway, as when JS is enabled it deletes existing inputs.

If you also look in TPRAddressLookup.cshtml, you can see how we add Umbraco support for an input. All text should be customisable in Umbraco, so we should add the same support for that.

lib/govuk-frontend
HarryAmmon, line 1: I would not expect this to be changed as part of this PR

HarryAmmon (changes requested):

